PR #586, opened 2023-06-19
Simplify and modernize navigation

kugel3:
## Description
Navigation best practices have changed since we started using TCA's navigation beta, and there are also places where we use needlessly complicated constructions - for example, in CreateAccountCoordinator we have architectured the code so that it's possible to start on any step, but in reality we always start with NameAccount. Exploiting that fact lets us simplify both the view and the model.

This PR also gets rid of one usage of Relay, in the DappInteraction. I personally don't think Relay makes things better there, but it adds an extra layer of indirection, a new surface area for bugs.

This PR is meant partly as a starting point for discussions around what we may want to change w.r.t navigation.

### DIFF
--- a/Sources/Features/DappInteractionFeature/Coordinator/DappInteractionFlow.swift
+++ b/Sources/Features/DappInteractionFeature/Coordinator/DappInteractionFlow.swift
@@ -43,11 +43,13 @@ struct DappInteractionFlow: Sendable, FeatureReducer {
 		let interactionItems: NonEmpty<OrderedSet<AnyInteractionItem>>
 		var responseItems: OrderedDictionary<AnyInteractionItem, AnyInteractionResponseItem> = [:]
 
+		var interactionItem: AnyInteractionItem? = nil
+
 		@PresentationState
 		var personaNotFoundErrorAlert: AlertState<ViewAction.PersonaNotFoundErrorAlertAction>? = nil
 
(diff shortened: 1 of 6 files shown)


Sources/Features/DappInteractionFeature/Coordinator/DappInteractionFlow.swift
```diff
@@ -43,11 +43,13 @@ struct DappInteractionFlow: Sendable, FeatureReducer {
 		let interactionItems: NonEmpty<OrderedSet<AnyInteractionItem>>
 		var responseItems: OrderedDictionary<AnyInteractionItem, AnyInteractionResponseItem> = [:]
 
+		var interactionItem: AnyInteractionItem? = nil
```
davdroman, line 46: You might want to look at [this PR](https://github.com/davdroman/tca-relay-reducer/pull/1) which [computes the current request](https://github.com/davdroman/tca-relay-reducer/pull/1#discussion_r1391866388) (interaction item) based on the current path instead of holding onto it as state which is more prone to bugs by virtue of having to mutate it at different points during the execution flow.

kugel3: Thanks, I'll definitely have a look. Trying to get rid of the fork is something I want to do this week or next so this comes in handy!

PR #808, opened 2023-09-28
Remove BigDecimal from tests

kugel3:
## Description
Now BigDecimal can also be removed from the test where we made sure that RETDecimal encoded and decoded in the same way as BigDecimal.

### DIFF
--- a/Tests/EngineKitTests/DecimalTests.swift
+++ b/Tests/EngineKitTests/DecimalTests.swift
@@ -4,8 +4,6 @@ import TestingPrelude
 
 // MARK: - DecimalTests
 final class DecimalTests: TestCase {
-	typealias RETDecimal = EngineToolkit.Decimal
-
 //	func testBreakDecimal() throws {
 //		// 2^256 -1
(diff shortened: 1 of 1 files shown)


Tests/EngineKitTests/DecimalTests.swift
```diff
@@ -616,45 +590,6 @@ extension DecimalTests {
 			"1234123.4",
 			"123456.34",
 			"12345.234",
-			"1234.1234",
```
CyonAlexRDX, line 619: why remove so many strings literals to test?

kugel3: These are just used for testing round trip coding now, which uses the built-in machine readable string representation. So the test can be pretty limited, we don't need to look at rounding or anything like that. Really it's enough to check integers, leading zeros, trailing zeros, negatives and 0. I made the test very exhaustive earlier because we encoded as BigDecimal and the decoded as RETDecimal, that made sense to test more exhaustively.

PR #812, opened 2023-09-29
[ABW-2378] Set current gateway to mainnet after import

CyonAlexRDX:
Jira ticket: [ABW-2378](https://radixdlt.atlassian.net/browse/ABW-2378)

# Description
Change current network (gateway) to `.mainnet` after having finished importing a `Profile` (wallet backup data).

# Prevents

General soundness measure, safe to do since all Profiles should contain `.mainnet` accounts. 

# How to test
You can use [this Profile](https://drive.google.com/file/d/1-4cFc3xfZgu5hJ-Zshq3O90QFh5kh1bF/view?usp=drive_link) (requires RDX Works account)

Mnemonic is:
```
anger dizzy pond category scissors broccoli earth royal palace intact fox engage mix area engage right argue derive promote control hard order silver wait
```

[ABW-2378]: https://radixdlt.atlassian.net/browse/ABW-2378?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Sources/Profile/Snapshot/ProfileSnapshot.swift
+++ b/Sources/Profile/Snapshot/ProfileSnapshot.swift
@@ -25,7 +25,7 @@ public struct ProfileSnapshot:
 
 	/// Settings for this profile in the app, contains default security configs
 	/// as well as display settings.
-	public let appPreferences: AppPreferences
+	public private(set) var appPreferences: AppPreferences
 
 	/// Effectivly **per network**: a list of accounts, personas and connected dApps.
 	public let networks: Profile.Networks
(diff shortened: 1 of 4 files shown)


Sources/Profile/Snapshot/ProfileSnapshot.swift
```diff
@@ -25,7 +25,7 @@ public struct ProfileSnapshot:
 
 	/// Settings for this profile in the app, contains default security configs
 	/// as well as display settings.
-	public let appPreferences: AppPreferences
+	public private(set) var appPreferences: AppPreferences
```
CyonAlexRDX, line 28: Preferable=y we want the Snapshot to be immutable, but in this case it would just be a hassle to "roundtrip" through `Profile` in `ProfileStore`'s function `changeProfileSnapshot:to`, thus we allow this, but still make the setter private, having to go through `changeCurrentToMainnetIfNeeded`.

PR #814, opened 2023-09-29
Add explicit metadata when loading resource pages

GhenadieVP:
Fixes the issue when the stakes where not visible if those are fetched from the GW in the second page of resources.


CyonAlexRDX (approved):
LGTM

PR #816, opened 2023-10-02
Check non-zero total supply

kugel3:
## Description
This could potentially fix the crashes reported on [here](https://rdxworks.slack.com/archives/C03QFAWBRNX/p1695983208070469)

What this PR does is simply to not calculate the redeemable amount for pool units if the total supply is 0, thus avoiding a potential division-by-zero crash.


CyonAlexRDX (approved):
ACK `7a0956b3b07148dee10e3cf51b91efecc623aa71`

PR #817, opened 2023-10-02
[ABW-2387] Fix bug where a failing TX was never removed from request queue

CyonAlexRDX:
[Jira Ticket ABW-2387](https://radixdlt.atlassian.net/browse/ABW-2387)

# Description
We missed handling `addressReservationValue`


Sources/EngineKit/HashableExtensions.swift
```diff
@@ -46,118 +52,173 @@ extension Instructions: Hashable {
 // MARK: - Instruction + Hashable
 extension Instruction: Hashable {
 	public func hash(into hasher: inout Hasher) {
+		switch self {
```
kugel3, line 55: We already do this switch below, so if new cases are added it seems that the other switch will catch that, and the developer will be prompted to handle the new case in the place where that is actually needed. So what is the purpose of the extra switch?

Sources/EngineKit/HashableExtensions.swift
```diff
@@ -46,118 +52,173 @@ extension Instructions: Hashable {
...
-		switch (lhs, rhs) {
+// MARK: - Instruction + Equatable
+extension Instruction: Equatable {
+	public static func == (lhsOuter: Instruction, rhsOuter: Instruction) -> Bool {
+		switch lhsOuter {
+		case .takeAllFromWorktop, .takeFromWorktop, .takeNonFungiblesFromWorktop, .returnToWorktop, .assertWorktopContains, .assertWorktopContainsNonFungibles, .popFromAuthZone, .dropAuthZoneProofs, .dropAuthZoneRegularProofs, .dropAuthZoneSignatureProofs, .dropAllProofs, .dropNamedProofs, .pushToAuthZone, .createProofFromAuthZoneOfAmount, .createProofFromAuthZoneOfNonFungibles, .createProofFromAuthZoneOfAll, .burnResource, .cloneProof, .dropProof, .assertWorktopContainsAny, .createProofFromBucketOfAmount, .createProofFromBucketOfNonFungibles, .createProofFromBucketOfAll, .callFunction, .callMethod, .callRoyaltyMethod, .callMetadataMethod, .callRoleAssignmentMethod, .allocateGlobalAddress, .callDirectVaultMethod:
+			break
+			// If switch does not compile, you MUST handle it below, in EQ impl.
+		}
+		func `true`(mustNotHaveAssociatedValue _: Instruction) -> Bool {
```
kugel3, line 218: We should not have functions named "true"

Sources/EngineKit/HashableExtensions.swift
CyonAlexRDX: right, not needed for `hash:into`, only for equality, copy paste bug...

Sources/EngineKit/HashableExtensions.swift
CyonAlexRDX: will rename to `equal`

Sources/EngineKit/HashableExtensions.swift
```diff
@@ -268,12 +346,20 @@ extension ManifestValue: Hashable {
...
-		default:
+		case let (.addressReservationValue(lhs), .addressReservationValue(rhs)):
+			return lhs == rhs
+		default /* `switch lhsOuter` above ensures we have handled each `case` */:
 			return false
 		}
 	}
 
 	public func hash(into hasher: inout Hasher) {
+		switch self {
```
CyonAlexRDX, line 357: removing

kugel3:
What is the point of an empty function `mustNotHaveAssociatedValue`?

Sources/Features/DappInteractionFeature/Interactor/DappInteractor.swift
```diff
@@ -10,10 +10,18 @@ import ROLAClient
...
+	public var id: ID {
+		request.id
+	}
+}
+
 // MARK: - DappInteractor
 struct DappInteractor: Sendable, FeatureReducer {
 	struct State: Sendable, Hashable {
-		var requestQueue: OrderedSet<RequestEnvelope> = []
+		var requestQueue: IdentifiedArrayOf<RequestEnvelope> = []
```
CyonAlexRDX, line 24: This is the actual bug fix

Sources/Features/DappInteractionFeature/Interactor/DappInteractor.swift
```diff
@@ -333,13 +343,15 @@ struct DappInteractor: Sendable, FeatureReducer {
...
+				} else {
+					loggerGlobal.notice("Failed to send response back to dapp, error: \(error), not delegating `sentResponseToDapp`.")
 				}
 			}
 		}
 	}
 
 	func dismissCurrentModalAndRequest(_ request: RequestEnvelope, for state: inout State) {
-		state.requestQueue.remove(request)
+		state.requestQueue.remove(id: request.id)
```
CyonAlexRDX, line 354: The bug was here, we actually never removed the `request` since Equality treated Manifest `M != M`.

Sources/Features/GatewaySettingsFeature/Children/AddNewGateway/AddNewGateway+Reducer.swift
```diff
@@ -74,7 +74,12 @@ public struct AddNewGateway: Sendable, FeatureReducer {
 		case let .textFieldChanged(inputtedURL):
 			state.inputtedURL = inputtedURL
 			state.errorText = nil
-			let url = URL(string: inputtedURL)
+			let url: URL?
+			if #unavailable(iOS 17.0) {
+				url = URL(string: inputtedURL)
+			} else {
+				url = URL(string: inputtedURL, encodingInvalidCharacters: false)
```
CyonAlexRDX, line 81: iOS 17 auto encodes spaces...

Sources/EngineKit/HashableExtensions.swift
```diff
@@ -268,7 +339,9 @@ extension ManifestValue: Hashable {
 			return lhs == rhs
 		case let (.nonFungibleLocalIdValue(lhs), .nonFungibleLocalIdValue(rhs)):
 			return lhs == rhs
-		default:
+		case let (.addressReservationValue(lhs), .addressReservationValue(rhs)):
+			return lhs == rhs
+		default /* `switch lhsOuter` above ensures we have handled each `case` */:
```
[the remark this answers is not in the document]
kugel3: For example, you can put a combined case like `(.boolValue,, _), (.i8Value, _)`etc in the main switch

Sources/EngineKit/HashableExtensions.swift
kugel3: Why not just remove it and return `true`?

Sources/EngineKit/HashableExtensions.swift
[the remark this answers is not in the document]
CyonAlexRDX: that is an explosion of combinations.

Sources/EngineKit/HashableExtensions.swift
```diff
@@ -46,118 +52,166 @@ extension Instructions: Hashable {
 // MARK: - Instruction + Hashable
 extension Instruction: Hashable {
 	public func hash(into hasher: inout Hasher) {
+		func mustNotHaveAssociatedValue(_: Instruction) {}
```
kugel3, line 55: What is the point of this?

kugel3: Is the idea to cause a compilation error if someone later adds an associated value to one of those cases? If so, what is the benefit of having that check here, and not where the enum is defined?

Sources/EngineKit/HashableExtensions.swift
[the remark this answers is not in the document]
kugel3: No, it's the same number, one entry per case. You have already handled `(.boolValue,, .boolValue)` above, so if you put `(.boolValue,, _)` you know that `_` will be something else.

Sources/EngineKit/HashableExtensions.swift
CyonAlexRDX: @kugel3 where do you wanna have that check?

kugel3: I think it also deserves a comment, like "if an associated value is added to one of the cases called below, then it will no longer be possible to create the case and the corresponding line will fail to compile"

CyonAlexRDX: sure! I thought the name was clear enough....

kugel3: Except that the name is a lie, you can call that function with any instruction, also ones that do have an associated value. It's not the function that enforces it, it's the call site.

CyonAlexRDX: resolved on slack

Sources/EngineKit/HashableExtensions.swift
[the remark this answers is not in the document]
kugel3: That criticism would also apply to `default`, which would completely break the hasher, but I don't think it's something we need to worry about. It would be very visible if that huge case got moved up to the individual cases, more so than for a lonely default.

kugel3 (approved):
I still don't like those noop switches, I think it would be clearer and safer to eliminate default cases.

Sources/EngineKit/HashableExtensions.swift
```diff
@@ -46,229 +52,477 @@ extension Instructions: Hashable {
...
 	}
+}
 
-	public static func == (lhs: Instruction, rhs: Instruction) -> Bool {
-		switch (lhs, rhs) {
+// MARK: - Instruction + Equatable
+extension Instruction: Equatable {
+	public static func == (lhsOuter: Instruction, rhsOuter: Instruction) -> Bool {
+		// A function call like dummy(.someCase) will stop compiling if an
+		// associated value is later added to the pop case
```
kugel3, line 211: I guess this should say "added to `someCase`" now

kugel3: Remember that functions can have proper doc comments with /// and backticks for code

Sources/EngineKit/HashableExtensions.swift
```diff
@@ -522,14 +782,32 @@ extension TransactionType: Hashable {
...
+		case (.claimStakeTransaction, _):
+			return false
+
+		case let (.stakeTransaction(lhsStakes), .stakeTransaction(rhsStakes)):
+			return lhsStakes == rhsStakes
+		case (.stakeTransaction, _):
+			return false
+
+		case let (.unstakeTransaction(lhsUnstake), .unstakeTransaction(unstakeTransaction)):
+			return lhsUnstake == unstakeTransaction
```
kugel3, line 809: the rhs value is not named `rhsUnstake`, is that on purpose?

CyonAlexRDX: accidental, will fix tomorrow, in another PR, copy paste error.

PR #827, opened 2023-10-06
WIP fix

CyonAlexRDX:
WIP fix


Sources/Clients/BackupsClient/BackupsClient+Test.swift
```diff
@@ -16,7 +16,10 @@ extension BackupsClient: TestDependencyKey {
 		lookupProfileSnapshotByHeader: unimplemented("\(Self.self).lookupProfileSnapshotByHeader"),
 		importProfileSnapshot: unimplemented("\(Self.self).importProfileSnapshot"),
 		importCloudProfile: unimplemented("\(Self.self).importCloudProfile"),
-		loadDeviceID: unimplemented("\(Self.self).loadDeviceID")
+		loadDeviceID: unimplemented("\(Self.self).loadDeviceID"),
+		reclaimProfileOnThisDevice: unimplemented("\(Self.self).reclaimProfileOnThisDevice"),
```
CyonAlexRDX, line 20: We display an alert in `Main` informing user that their profile might be used on another device, in which the can select "reclaim" or "stop using" (not that copy, see copy further down).

Sources/Clients/ProfileStore/ProfileStore.swift
```diff
@@ -106,6 +108,13 @@ extension ProfileStore {
 		lens(\.profile)
 	}
 
+	/// A multicasting replaying async sequence of conflicting device the profile snapshot is used on.
+	public func usedOnOtherDevice() async -> AnyAsyncSequence<ProfileSnapshot.Header.UsedDeviceInfo> {
```
CyonAlexRDX, line 112: async sequence that `BackupClient` expose, used in `Main` reducer to listen to events where Profile might be used on another device. Replacing old `throw error` based solution

PR #829, opened 2023-10-09
[ABW-2271] Remove spurious portfolio reloading

kugel3:
## Description
The wallet reloads the account portfolios too often. This PR removes most of the unnecessary reloads, and even more cases where an already fetched portfolio is recreated without fetching.

### Notes
The next step is probably to move to an architecture where e.g. AccountPortfolioClient emits "events" like `portfolioChange(AccountAddress)` which we subscribe to, instead of a currentValueSubject of the portfolios. This will make it possible to distinguish between the first reading of an already fetched portfolio, and subsequent updates to the data. This is important because `.task` is called not just when a view is first created, but every time it comes back into view, causing us to set up subscriptions too often.

## How to test
Try everything related to portfolio reloading, making sure that views are automatically reloaded e.g. when
- Using the faucet
- Performing a transfer
- Accepting a remotely initiated transaction
- etc

## PR submission checklist


CyonAlexRDX (approved):
LGTM

GhenadieVP (approved):

PR #832, opened 2023-10-09
[ABW-2426] Keychain improvements

CyonAlexRDX:
This PR brings several improvements to keychain:
* Ability to read `service` and `accessGroup` (not used in app, but could e.g. be used for `DEBUG` builds and shown somewhere App Settings)
* Ability to check if keychain contains an Authentication protected item without triggering Auth Prompt
* Atomic load-else-save-new operation (`getDataWithoutAuth:forKey:ifNilSet` and `getDataWithAuth:forKey:ifNilSet`) which protects against data races if multiple Task try to to load else save, since this is an actor isolated method
* Add a Keychain demo for DEBUG builds available in App Settings (as seen in video)
* Keychain `label` now has suffix `" DEBUG"` for DEBUG builds, greatly helping us disambiguate between items when we look in Keychain Access on macOS.
* Unit testing of Keychain!
* The scheme `Unit Test` has been removed, instead you now run test using the scheme `Radix Wallet Dev (iOS)`

https://github.com/radixdlt/babylon-wallet-ios/assets/116169792/6a161cb9-91b6-42a2-b781-2343c38b9137

image

https://github.com/radixdlt/babylon-wallet-ios/assets/116169792/d1fd4fa6-f469-4d8c-875a-76845ac2afc1


App/RadixWalletTests/KeychainActorTests.swift
```diff
@@ -0,0 +1,152 @@
...
+	func testNoAuth() async throws {
+		try await onceNoAuthTest()
+	}
+
+	func testAuth() async throws {
+		try await onceAuthTest()
+	}
+
+	func onceNoAuthTest() async throws {
+		try await sut.removeAllItems()
```
GhenadieVP, line 22: I would call this in `setUp` and/or `tearDown` methods.

App/RadixWalletTests/KeychainActorTests.swift
```diff
@@ -0,0 +1,152 @@
...
+
+	func onceAuthTest() async throws {
+		try await sut.removeAllItems()
+		let startValue = try await sut.getDataWithAuth(
+			forKey: authRandomKey,
+			authenticationPrompt: "onceAuthTest"
+		)
+		XCTAssertNil(startValue)
+
+		let values = try await valuesFromManyTasks {
```
GhenadieVP, line 40: We should probably also test that the expected value is written, not only that there is only one value.

Sources/Clients/SecureStorageClient/SecureStorageClient+Interface.swift
```diff
@@ -37,6 +38,7 @@ extension SecureStorageClient {
 	public typealias SaveProfileHeaderList = @Sendable (ProfileSnapshot.HeaderList) async throws -> Void
 	public typealias DeleteProfileHeaderList = @Sendable () async throws -> Void
 
+	public typealias GetDeviceIdentifierSetIfNil = @Sendable (UUID) async throws -> UUID
```
GhenadieVP, line 41: I am not really sure this is the best idea, this is both a Query(read the identifier) and a Command(save the identifier).
Additionally, shouldn't it be a function that composes `LoadDeviceIdentifier` and `SaveDeviceIdentifier`?

Sources/Prelude/KeychainClient/KeychainActor.swift
```diff
@@ -0,0 +1,156 @@
+import Foundation
+import KeychainAccess
+
+// MARK: - KeychainActor
+@_spi(KeychainInternal)
```
GhenadieVP, line 5: is `@_spi` stable to be used? last time I did check it was an experimental feature.

Sources/Clients/SecureStorageClient/SecureStorageClient+Interface.swift
kugel3: I agree, I think we want to split up reading, generating and saving

Sources/Prelude/KeychainClient/KeychainActor.swift
kugel3: Why do we need it, for testing? Can't we solve it with testable?

Sources/Clients/SecureStorageClient/SecureStorageClient+Interface.swift
CyonAlexRDX: You are both missing one crucial detail. We CANNOT compose `Load` and `Save` because that happens outside of the actor, thus is not actor isolated, thus not data race free. Which I've asserted in the unit tests.....

So the actor must have this function and expose it.

This is used in `claimProfileSnapshot` which calls `saveDeviceIdentifierIfNeeded`, which before this PR was:
```swift
	public func saveDeviceIdentifierIfNeeded(_ deviceID: UUID) async throws {
		if let existing = try? await loadDeviceIdentifier() {
			if existing != deviceID {
				try await saveDeviceIdentifier(deviceID)
			}
		} else {
			try await saveDeviceIdentifier(deviceID)
		}
	}
```

which is NOT data race free, since it happens outside of the Keychain actor.

Sources/Prelude/KeychainClient/KeychainActor.swift
```diff
@@ -0,0 +1,156 @@
...
+public final actor KeychainActor: GlobalActor {
```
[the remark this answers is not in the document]
CyonAlexRDX: @GhenadieVP what do you mean - this has ALWAYS been an actor https://github.com/radixdlt/babylon-wallet-ios/blob/main/Sources/Prelude/KeychainClient/KeychainClient%2BLive.swift#L12

and I have a fix for sync read profile once in app launch in another branch.

Sources/Prelude/KeychainClient/KeychainActor.swift
CyonAlexRDX: we dont wanna expose this to much, in fact it used to be completely hidden - thus untested - https://github.com/radixdlt/babylon-wallet-ios/blob/main/Sources/Prelude/KeychainClient/KeychainClient%2BLive.swift#L12

After the `Mono` PR has been merged, I can removed `@_spi`, since then I can use `testable import`, but it cannot be done now, since I cannot testable import another target, I think.

Sources/Prelude/KeychainClient/KeychainClient+Interface.swift
```diff
@@ -6,35 +6,74 @@ import Tagged
 /// A CRUD client around Keychain, that provides async methods for operations that requires auth
 /// and sync methods for operations on data without authentication.
 public struct KeychainClient: Sendable {
-	/// `sync` adds or updates data for `key` protected with specified accessibility, if a `Label` is provided it
-	/// will be set. If a `Comment` is provided, it will be set.
-	public var setDataWithoutAuthForKey: SetDataWithoutAuthForKey
+	public var _getServiceAndAccessGroup: GetServiceAndAccessGroup
```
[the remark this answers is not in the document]
CyonAlexRDX: sure can be made private! but I can do it in a coming PR (trying to stream line git admin today...)

App/RadixWalletTests/KeychainActorTests.swift
CyonAlexRDX: how would that work? I used RANDOM data, and also I do not know the which of all tasks that will be used.

In a test of the Keychain CLIENT I can do that, by hook into the save method....

App/RadixWalletTests/KeychainActorTests.swift
CyonAlexRDX: sure, can change to that later - trying to avoid to much git admin today.

Sources/Prelude/KeychainClient/KeychainActor.swift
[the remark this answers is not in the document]
GhenadieVP: I know that it has been, but do we want to keep it that way?

Sources/Clients/SecureStorageClient/SecureStorageClient+Interface.swift
GhenadieVP: But why should this be on SecureStorageClient actor? I mean `ProfileStore` is also an actor.

CyonAlexRDX: @GhenadieVP SecureStorageClient is NOT an Actor. 

In code, ONLY SecureStorageClient dep should use KeychainClient, all other code always goes through SecureStorageClient.... it is an abstraction on top of Keychain.

Sources/Prelude/KeychainClient/KeychainActor.swift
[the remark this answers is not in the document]
CyonAlexRDX: @GhenadieVP yes ofc! that is what makes it safe! :D otherwise non of the unit tests in KeychainTests would pass...

App/RadixWalletTests/KeychainActorTests.swift
GhenadieVP: Shouldn't we test that the given task was executed in the proper serial order? meaning that the value written should be the value of the last task? Or maybe I am missing what you are trying to test here.

Sources/Clients/SecureStorageClient/SecureStorageClient+Interface.swift
GhenadieVP: Ok, my bad missed that SecureStorageClient is not actor. Would this `GetDeviceIdentifierSetIfNil` be used outside of ProfileStore? If not, how is `GetDeviceIdentifierSetIfNil` on SecureStorageClient safer than `saveDeviceIdentifierIfNeeded` on ProfileStore?

App/RadixWalletTests/KeychainActorTests.swift
GhenadieVP: yes, ofc, and are we ok with that? having unordered operations on the Keychain?

CyonAlexRDX: we have it already, on `main`.... this PR just makes them **free from data races**, on main they are not.

Sources/Clients/SecureStorageClient/SecureStorageClient+Interface.swift
CyonAlexRDX: there is no "saveDeviceIdentifierIfNeeded on ProfileStore", before this PR ProfileStore uses SecureStorageClient - as it should - and after this PR ProfileStore still uses SecureStorage. And SecureStorage, is an abstraction layer on top of Keychain, as it has always been.

The change in this PR is that the bug we had in `main` before, with multiple competing `loadElseGenerateAndSave` is not possible, the bug is eradicated. Apart from we are not doing it like that anymore... since https://github.com/radixdlt/babylon-wallet-ios/pull/828
This PR does makes `claimProfile` even more safe! But it is not called from multiple tasks... but if it would, it would still be safe.

App/RadixWalletTests/KeychainActorTests.swift
GhenadieVP: sure, I am using this PR as means to discuss about how we do use the Keychain, we could have it separately then, don't want to block this PR for merging.

Sources/Prelude/KeychainClient/KeychainActor.swift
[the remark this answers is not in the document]
CyonAlexRDX: @GhenadieVP right, so, we could need to SPLIT `KeychainActor` and thus `KeychainClient` into two different sets for that to happen. Since an actor *cannot let a stored property be marked `nonisolate`*, and in order to let `reading without auth` be non async, the methods would need to be `nonisolated`, which is not possible since they access actor-isolated `let keychain` stored property.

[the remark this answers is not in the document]
GhenadieVP: Are we only limited to using Actors? We cannot have the same functionality without needing to have a KeychainActor?

Sources/Clients/SecureStorageClient/SecureStorageClient+Interface.swift
GhenadieVP: Fair enough, I am still not sold on having this, unless you mean to have it as short term solution for the current state of the code. We do use that `saveDeviceIdentifierIfNeeded` as a workaround for the bug, and it really shouldn't be there.

GhenadieVP (approved):
LGTM!
I would want to discuss more about how we do operations on the Keychain, but could be done separately, not blocking this PR.

Sources/Prelude/KeychainClient/KeychainActor.swift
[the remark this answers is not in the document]
CyonAlexRDX: Actor is a great thing! it is what protects us from bugs in Keychain, at least after this PR, which does not use `Task {}` **unstructured** concurrency, like we had before, which was wrong.

Without actors, we can accidentally corrupt keychain data. We use `async` functions all over the code anyway, I don't really understand the reluctance? 

In a coming PR I'm gonna refactor ProfileStore, to be much simpler, using a single read of `profileSnapshot: ProfileSnapshot?` part of init of KeychainActor, which I hope will allow ProfileStore init to be sync.

Sources/Clients/SecureStorageClient/SecureStorageClient+Interface.swift
CyonAlexRDX: @GhenadieVP this PR really does not change the behaviour of ProfileStore at all, at least that is my claim! The only thing is that that function `saveDeviceIdentifierIfNeeded` has been made more safe. That is all.

Sources/Prelude/KeychainClient/KeychainActor.swift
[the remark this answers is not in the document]
CyonAlexRDX: I mean, if we would only have ONE SINGLE ACTOR in the whole code base, I would let it be the KeychainActor. It is the only crucial data we have in our entire code base.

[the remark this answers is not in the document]
kugel3: The problem with async is that it breaks TCA reducers. Instead of just reading from the keychain and going on with your day, you have to send an internal action, pick up that action and then continue doing what you did. So we should ideally limit async methods in clients as much as possible.

Would it not be possible to for example have an actor for the auth requiring methods, inside a type that is itself not an actor?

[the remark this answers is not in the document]
CyonAlexRDX: Lets take the actor discussion on Slack/on a call, since it does not related to this PR, since this PR does not change that... only moved the existing actor out into a seperate file.

PR #833, opened 2023-10-09
[ABW-2428] Resolve owner when extracting behaviors

kugel3:
Jira ticket: [ABW-2428](https://radixdlt.atlassian.net/browse/ABW-2428)

## Description
Before this PR, all uses of "Owner" were interpreted the same way as "Protected", which is incorrect. We should in fact resolve this by looking up the Owner value.

## How to test
Find an asset that uses Owner in a rule and make sure it is resolved properly (by looking at the behavior(s) we extract for the assets.

## PR submission checklist

[ABW-2428]: https://radixdlt.atlassian.net/browse/ABW-2428?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Sources/Clients/GatewayAPI/GatewayAPIClient+Extension.swift
+++ b/Sources/Clients/GatewayAPI/GatewayAPIClient+Extension.swift
@@ -207,37 +207,89 @@ extension GatewayAPI.EntityMetadataCollection {
 extension GatewayAPI.ComponentEntityRoleAssignments {
 	// FIXME: This logic should not be here, will probably move to OnLedgerEntitiesClient.
 	@Sendable public func extractBehaviors() -> [AssetBehavior] {
+		typealias AssignmentEntry = GatewayAPI.ComponentEntityRoleAssignmentEntry
 		typealias ParsedName = GatewayAPI.RoleKey.ParsedName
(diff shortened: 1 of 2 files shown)


Sources/Clients/GatewayAPI/GatewayAPIClient+Extension.swift
```diff
@@ -207,37 +207,58 @@ extension GatewayAPI.EntityMetadataCollection {
 extension GatewayAPI.ComponentEntityRoleAssignments {
 	// FIXME: This logic should not be here, will probably move to OnLedgerEntitiesClient.
 	@Sendable public func extractBehaviors() -> [AssetBehavior] {
```
GhenadieVP, line 209: this becoming more and more complex, could benefit from  adding some tests

GhenadieVP (approved):

Sources/Clients/GatewayAPI/GatewayAPIClient+Extension.swift
kugel3: Yes, there should definitely be tests, and I will add a longer comment explaining the logic

PR #834, opened 2023-10-11
[ABW-2438] Parsing Decimals

kugel3:
Jira ticket: [ABW-2438](https://radixdlt.atlassian.net/browse/ABW-2438)

## Description
- We want to accept numbers starting with a decimal separator (`.1`)
- For the guarantee, we want to save all the decimals the user enters
- Also, the guarantee stepper should work in 1% point increments
- When pressing plus/minus on a number with decimals, it will first go to the nearest higher/lower integer

Example for the last point:
If the number is `12.34` and you press +, the number will change to 13. If you press -, it would change to 12. 

## PR submission checklist

[ABW-2438]: https://radixdlt.atlassian.net/browse/ABW-2438?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Sources/EngineKit/RETDecimal+Extensions.swift
+++ b/Sources/EngineKit/RETDecimal+Extensions.swift
@@ -87,13 +87,18 @@ extension RETDecimal {
 // MARK: Truncation and rounding
 
 extension RETDecimal {
-	/// Truncates to `decimalPlaces` decimals
-	public func truncated(decimalPlaces: UInt) -> RETDecimal {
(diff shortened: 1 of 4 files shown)


GhenadieVP (approved):
LGTM!
Adding a simple test for parsing something like `.5` would be nice.

PR #835, opened 2023-10-11
Testing behaviors

kugel3:
## Description
- Adds test for `extractBehaviors()`
- Adds a thorough doc comment to that function
- Simplifies the way the function works

## How to test
This should not change how the app behaves, so the way to test it is to look at various resources and make sure the lists behaviors are still correct


CyonAlexRDX (approved):
Nice!

GhenadieVP (approved):

PR #848, opened 2023-10-18
[ABW-2465] Gracefully handle missing mnemonic

GhenadieVP:
Jira ticket: https://radixdlt.atlassian.net/browse/ABW-2465

## Context
Users might loose access to the on device stored mnemonic for various reasons - remove the passcode on the phone, migrate to a new phone, or any other operation on the Keychain that the Wallet cannot control. This PR aims to gracefully inform the users that the mnemonic is missing and the Seed Phrase needs to be imported.

## What was done:
- Made use in the Home screen of the current check done at startup that validates the existence of the babylon mnemonic, as well the its controlled entities public keys.
- On Home screen added the check for the existence of the mnemonics for the accounts that have their mnemonics stored on device.
- When users want to import a given account seed phrase, the Wallet show the appropriate import flow - for main seed phrase and for secondary ones.
- Add the alert in different contexts - Deriving public keys, signing a transaction - informing the user that the seed phrase is missing if that is the case.

## Demo of the error alert:

https://github.com/radixdlt/babylon-wallet-ios/assets/118184705/29e693c5-e3d5-4e4e-ac52-86a568f7fa96

## Demo of import seed phrases:

https://github.com/radixdlt/babylon-wallet-ios/assets/118184705/5eb13119-efeb-4808-babe-9168f09bda3f


RadixWallet/Clients/OverlayWindowClient/OverlayWindowClient+Interface.swift
```diff
@@ -102,6 +104,16 @@ extension OverlayWindowClient {
 	}
 }
 
+extension OverlayWindowClient.Item.AlertState {
+	public static var missingMnemonicAlert: Self {
+		// FIXME: Strings
+		.init(
+			title: { TextState("Could Not Complete") },
```
Sajjon, line 111: // FIXME: Strings

RadixWallet/Features/AccountDetailsFeature/Coordinator/AccountDetails+Reducer.swift
```diff
@@ -88,7 +94,7 @@ public struct AccountDetails: Sendable, FeatureReducer {
 		case loadMnemonicResult(TaskResult<MnemonicWithPassphraseAndFactorSourceInfo>)
 
 		case loadImport
-		case loadProfileSnapshotForRecoverMnemonicsFlow(TaskResult<ProfileSnapshot>)
+		case loadControlledEntitiesForRecoverMnemonicsFlow(TaskResult<EntitiesControlledByFactorSource>)
```
Sajjon, line 97: sure we can do this for now, important we do ABW-2206 soon, which moves import to Settings, never initated from any other screen than `Settings-> Account Settings & Security -> Seed Phrases`

RadixWallet/Features/AccountDetailsFeature/Coordinator/AccountDetails+Reducer.swift
```diff
@@ -128,8 +134,8 @@ public struct AccountDetails: Sendable, FeatureReducer {
 				// FIXME: Rename `ImportMnemonic` -> `ExportOrImportMnemonic` ?
 				ImportMnemonic()
 			}
-			Scope(state: /State.importMnemonics, action: /Action.importMnemonics) {
-				ImportMnemonicsFlowCoordinator()
+			Scope(state: /State.importMnemonic, action: /Action.importMnemonic) {
+				ImportMnemonicControllingAccounts()
```
Sajjon, line 138: hmm this kinds of break something Matt has said in the past, to always trigger import of ALL missing mnemonics when once trigger import...

RadixWallet/Features/DerivePublicKeys/DerivePublicKeys.swift
```diff
@@ -204,16 +205,23 @@ extension DerivePublicKeys {
...
+			return .delegate(.derivedPublicKeys(
+				hdKeys,
+				factorSourceID: deviceFactorSource.id.embed(),
+				networkID: networkID
+			))
+		} catch {
+			if error is FailedToFindFactorSource {
+				_ = await overlayWindowClient.scheduleAlert(.missingMnemonicAlert)
+			}
+			throw error
```
Sajjon, line 223: Should we rethrow even if we show alert? Or only rethrow inside an `else`?

RadixWallet/Features/HomeFeature/Coordinator/Home.swift
```diff
@@ -35,6 +35,7 @@ public struct Home: Sendable, FeatureReducer {
 
 	public enum InternalAction: Sendable, Equatable {
 		case accountsLoadedResult(TaskResult<Profile.Network.Accounts>)
+		case mnemonicAccessResult([FactorSourceID.FromHash: Bool])
```
Sajjon, line 38: Bool is a bit ambiguous... at least typealias? Does true mean has access? Or true is is missing?

RadixWallet/Profile/Entity/Account.swift
```diff
@@ -143,6 +143,13 @@ extension Profile.Network.Account {
 			control.transactionSigning.factorSourceID.kind == .ledgerHQHardwareWallet
 		}
 	}
+
+	public var deviceFactorSourceID: FactorSourceID.FromHash {
```
CyonAlexRDX, line 147: We cannot have this... `deviceFactorSourceID`... MUST be optional, and you MUST check that the factor source kind of `control.transasctionSigning` is indeed `.device`. It could be `.ledger`!

CyonAlexRDX (changes requested):
You must change the `public var deviceFactorSourceID: FactorSourceID.FromHash` into an optional, and check that it is not a `ledger` controlled account.

RadixWallet/Features/DerivePublicKeys/DerivePublicKeys.swift
CyonAlexRDX: Add a `FIXME`, relying on `error is ExpectedErrorType` is not the best code IMO. If we change the error thrown to something else, this breaks, without compiler informing us... It is OK for now, as a quickfix, but I want something more structured in place for the future (soon) which does not break when error type changes.

RadixWallet/Clients/OverlayWindowClient/OverlayWindowClient+Interface.swift
GhenadieVP: There is one above though

RadixWallet/Features/DerivePublicKeys/DerivePublicKeys.swift
GhenadieVP: The error still needs to be propagated further so the flow proceeds accordingly, the alert is "just" an intermediary info we show to the user.

As for a more robust solution, I think we talked in one of our standups that we should have a single/reusable slide up that reads from the keychain the mnemonic, that will also by implication show this alert. Currently, as you see, I had to add this check in several places.

RadixWallet/Features/AccountDetailsFeature/Coordinator/AccountDetails+Reducer.swift
GhenadieVP: @Sajjon Reverted back to use the whole import mnemonics flow.

RadixWallet/Features/ProfileBackupsFeature/RestoreProfileFromBackup/Children/ImportSeedPhrasesFlow/ImportMnemonicsFlowCoordinator.swift
```diff
@@ -79,9 +82,14 @@ public struct ImportMnemonicsFlowCoordinator: Sendable, FeatureReducer {
 				await send(.internal(.loadControlledEntities(TaskResult {
 					let ents = try await deviceFactorSourceClient.controlledEntities(snapshot)
 					try? await clock.sleep(for: .milliseconds(200))
-					return ents.filter { ent in
-						!userDefaultsClient.getFactorSourceIDOfBackedUpMnemonics().contains(ent.factorSourceID)
```
GhenadieVP, line 83: @Sajjon I don't fully understand why this is needed? I added additionally the check for the mnemonic access, let me know if I didn't break some scenario, when it is ok to not show the import mnemonic even though it is missing.

CyonAlexRDX: Yea best to remove this filtering now. Seems wrong to have it.

GhenadieVP: kept it for now, my check for mnemonic access will overrule it.

RadixWallet/Features/ProfileBackupsFeature/RestoreProfileFromBackup/Children/ImportSeedPhrasesFlow/ImportMnemonicControllingAccounts+View.swift
```diff
@@ -2,14 +2,15 @@ import ComposableArchitecture
...
-		.init(isSkippable: entitiesControlledByFactorSource.isSkippable)
+		.init(isSkippable: entitiesControlledByFactorSource.isSkippable, disableSkipImport: disableSkipImport)
 	}
 }
 
 // MARK: - ImportMnemonicControllingAccounts.View
 extension ImportMnemonicControllingAccounts {
 	public struct ViewState: Equatable {
 		let isSkippable: Bool
+		let disableSkipImport: Bool
```
CyonAlexRDX, line 13: why is this needed? why is not `isSkippable` enough?

RadixWallet/Features/ProfileBackupsFeature/RestoreProfileFromBackup/Children/ImportSeedPhrasesFlow/ImportMnemonicsFlowCoordinator.swift
CyonAlexRDX: hmm I think we really SHOULD delete any `getFactorSourceIDOfBackedUpMnemonics` logic here at all actually. Otherwise we are skipping import of a de facto missing mnemonic, just because app knows that user has sometime marked it as backed up. So lets remove:

`let mnemonicIsBackedUp = userDefaultsClient.getFactorSourceIDOfBackedUpMnemonics().contains(ent.factorSourceID)`

and change to:

```swift
if hasAccessToMneminic {
 return nil // exclude this mnemonic from mnemonics to import, already present,.
} else {
 return ent // user does not have access to this, needs importing.
}
```

RadixWallet/Features/ProfileBackupsFeature/RestoreProfileFromBackup/Children/ImportSeedPhrasesFlow/ImportMnemonicControllingAccounts+View.swift
GhenadieVP: leftover, will remove

CyonAlexRDX (approved):
LGTM

PR #858, opened 2023-10-20
Fix fastlane scan failing

CyonAlexRDX:
# Description
`bundle exec fastlane tests` failed 
  - locally due to RadixWalletTests requiring signing, and sometimes due to Mac being used
  - sometimes in CI due to Mac being used

This fixes that.

# Changes
* Remove swiftgen build phase, it seems it might have caused issues with code signing, since swiftgen mac target is used to format, also it can break CI... so best go back to have to manually call `swiftgen` locally after having updated resources
* Update code sign settings for RadixWalletTests, should hopefully be more robust now (no code sign)
* Remove mac / ipad / mac catalyst as supported platforms from RadixWalletTests
* Change from seemlingly broken `SCAN_DESTINATION` to `SCAN_DEVICE` which seems to... "just work".


RadixWallet.xcodeproj/project.pbxproj
```diff
@@ -8787,10 +8764,13 @@
 				CLANG_WARN_UNGUARDED_AVAILABILITY = YES_AGGRESSIVE;
 				CLANG_WARN_UNREACHABLE_CODE = YES;
 				CLANG_WARN__DUPLICATE_METHOD_MATCH = YES;
-				CODE_SIGN_STYLE = Automatic;
+				"CODE_SIGN_IDENTITY[sdk=iphoneos*]" = "iPhone Developer";
```
GhenadieVP, line 8767: We should probably continue using Config files, instead of modifying the pbxproj directly

CyonAlexRDX: yeah I have a branch for that, but it is a bigger task, so lets do it later.

GhenadieVP (approved):

PR #859, opened 2023-10-20
Properly check for the missing mnemonic

GhenadieVP:
Jira ticket: https://radixdlt.atlassian.net/browse/ABW-2511

Moved the check in a single place, where we do know for sure that the mnemonic is missing. Was initially reticent to put the alert in secureStorageClient, to not be too upfront with displaying the alert, but it works well in current setup. We should move this potentially closer to UI, when we will have a dedicated sheet for reading the mnemonic from the Keychain.


CyonAlexRDX (approved):
LGTM

PR #860, opened 2023-10-20
[ABW-2427] ProfileStore sync init

CyonAlexRDX:
[Jira Ticket ABW-2427](https://radixdlt.atlassian.net/browse/ABW-2427)

# Changes
* UserDefaults => sync ([see discussion](https://github.com/pointfreeco/isowords/discussions/191))
* Keychain => sync
* SecureStorageClient => sync
* ProfileStore methods => sync (internally, but still an `actor` so consumers still need `await profileStore.someFunc()`)
* Reduction of complexity in `ProfileStore` (no more `ephemeralProfile`)
* ProfileStore now eagerly generates and **saves** a Profile (and BDFS (as in `main`)) and onboards users if no mainnet accounts.
* Code coverage increase of ProfileStore from near near zero to `95%`

## 95% Code Coverage 🎉

Screenshot 2023-10-26 at 11 04 32

# Existing wallet works
https://github.com/radixdlt/babylon-wallet-ios/assets/116169792/09f476df-c221-444e-8f5e-0f1e667c2f27

# Keychain test works
https://github.com/radixdlt/babylon-wallet-ios/assets/116169792/ac295289-c5dc-431b-9ec2-722c65635edb

# Delete wallet create new works
https://github.com/radixdlt/babylon-wallet-ios/assets/116169792/306b06e9-d0fd-434b-8e46-13905e346dd3

# Warning when used on two devices

IMG_0579

[See google drive (has change a bit since the recoding...)](https://drive.google.com/file/d/1rbtDI8OwcC6Fkq6SV5A9kkhto1zfxju_/view?usp=share_link)


RadixWallet.xcodeproj/xcuserdata/sajjon.xcuserdatad/xcschemes/xcschememanagement.plist
Sajjon, line 1: oops, should delete this

RadixWallet/Clients/ProfileStore/ProfileStore.swift
Sajjon, line 1: The most important file of the PR obviously...

RadixWallet/Clients/ProfileStore/ProfileStore.swift
```diff
@@ -1,182 +1,75 @@
...
 
-	/// Current Profile
-	let profileStateSubject: AsyncCurrentValueSubject<ProfileState>
+	/// Only mutable since we need to update the description with async, since reading
+	/// device model and name is async.
+	private var deviceInfo: DeviceInfo
 
-	init() async {
-		self.profileStateSubject = await .init(Self.restoreFromSecureStorageIfAble())
+	init() {
```
Sajjon, line 16: init is now sync!

RadixWallet/Clients/ProfileStore/ProfileStore.swift
```diff
@@ -1,182 +1,75 @@
...
+		let metaDeviceInfo = Self._deviceInfo()
+		let stuff = Self._loadSavedElseNewProfile(metaDeviceInfo: metaDeviceInfo)
+		self.deviceInfo = stuff.deviceInfo
+		self.profileSubject = AsyncCurrentValueSubject(stuff.profile)
+
+		if let conflictingOwners = stuff.conflictingOwners {
+			Task {
+				loggerGlobal.notice("ProfileStore:init => emitOwnershipConflict")
+				try await self.emitOwnershipConflict(
+					ownerOfCurrentProfile: conflictingOwners.ownerOfCurrentProfile
```
Sajjon, line 26: this emits the alert on `overlayWindowClient`, the result of this is that it is seen simultaneously with Keychain Bio Auth prompt... not ideal. Can perhaps "buffer" overlayClient until a new method `ready` has been called, done in Splash after initial unlock bio is done... or something similar.

RadixWallet/Clients/ProfileStore/ProfileStore.swift
```diff
@@ -1,182 +1,75 @@
...
-		/// be discarded.
-		case ephemeral(EphemeralProfile)
-
-		/// When the async Task that loads profile - if any - from secure
-		/// storage completes and indeed a profile was found, the state
-		/// is changed in to this.
-		case persisted(Profile)
-	}
+// MARK: - ConflictingOwners
+public struct ConflictingOwners: Sendable, Hashable {
```
Sajjon, line 34: move out of this file.

RadixWallet/Clients/ProfileStore/ProfileStore.swift
```diff
@@ -288,364 +223,305 @@ extension ProfileStore {
...
 				)
+			} else {
+				return try (
+					deviceInfo: metaDeviceInfo.deviceInfo,
+					profile: _tryGenerateAndSaveNewProfile(deviceInfo: deviceInfo),
+					conflictingOwners: nil
+				)
 			}
+		} catch {
+			fatalError("Unable to use app. error: \(error)")
```
Sajjon, line 331: should probably allow for an "escape hatch" here, I mentioned it on Slack, where user could see a list of keychain items with the label we use on mnemonics and be able to see their mnemonics.

RadixWalletTests/Clients/ProfileStoreTests/ProfileStoreTests.swift
CyonAlexRDX, line 1: Will continue to add
Even more tests here.

RadixWallet/Clients/ProfileStore/ProfileStore.swift
```diff
@@ -1,182 +1,75 @@
...
-	let profileStateSubject: AsyncCurrentValueSubject<ProfileState>
+	/// Only mutable since we need to update the description with async, since reading
+	/// device model and name is async.
+	private var deviceInfo: DeviceInfo
 
-	init() async {
-		self.profileStateSubject = await .init(Self.restoreFromSecureStorageIfAble())
+	init() {
+		let metaDeviceInfo = Self._deviceInfo()
+		let stuff = Self._loadSavedElseNewProfile(metaDeviceInfo: metaDeviceInfo)
```
GhenadieVP, line 18: stuff :)?

RadixWallet/Clients/ProfileStore/ProfileStore.swift
```diff
@@ -185,100 +78,142 @@ extension ProfileStore {
...
+	func _update(profile: Profile) throws {
+		try saveProfileAfterUpdateItsHeader(profile)
+	}
+}
+
+// MARK: Private
+extension ProfileStore {
+	/// Asserts identity and ownership of a profile, then updates its header, saves it and emits an update.
+	/// - Parameter updated: Profile to save (after updating its header).
+	private func saveProfileAfterUpdateItsHeader(
```
GhenadieVP, line 174: The name of the function is not consistent with its content, the header is updated first after that the profile is saved. Could this function name be "just" "saveProfile"?

RadixWallet/Clients/ProfileStore/ProfileStore.swift
```diff
@@ -288,364 +223,306 @@ extension ProfileStore {
...
-				loggerGlobal.trace("Verified that device owns profile snapshot ✅")
-			}
-		} else {
-			loggerGlobal.notice("Found no DeviceIdentifier in keychain rectifiying by saving the one found in profile header: \(lastUsedOnDevice.id) into keychain.")
-			do {
-				try await secureStorageClient.saveDeviceIdentifier(lastUsedOnDevice.id)
-			} catch {
-				loggerGlobal.error("Failed to rectify deviceID discrepancy for users, error: \(error) (failed to load deviceID from keychain)")
+	private func _assertOwnership(of profile: Profile) throws {
+		try Self._assertOwnership(of: profile, against: deviceInfo) {
```
GhenadieVP, line 256: is `Self` actually needed?

RadixWallet/Clients/ProfileStore/ProfileStore.swift
```diff
@@ -288,364 +223,306 @@ extension ProfileStore {
...
+			Task {
```
GhenadieVP, line 257: This task could be inside `emitOwnershipConflict`, which then will become a sync nonThrowing function.

RadixWallet/Clients/ProfileStore/ProfileStore.swift
```diff
@@ -288,364 +223,306 @@ extension ProfileStore {
...
+private struct MetaDeviceInfo: Sendable, Hashable {
+	let deviceInfo: DeviceInfo
+	let fromDeprecatedDeviceID: Bool
 }
 
-extension UserDefaultsClient {
-	public func getActiveProfileID() -> ProfileSnapshot.Header.ID? {
-		stringForKey(.activeProfileID).flatMap(UUID.init(uuidString:))
+extension OverlayWindowClient.Item.AlertState {
+	public static func profileUsedOnAnotherDeviceAlert(
```
GhenadieVP, line 487: Not quite sure about this alert, should probably be a proper sheet that can display a better info. But this is a product decision.

RadixWallet/Clients/BackupsClient/BackupsClient+Live.swift
```diff
@@ -21,35 +21,26 @@ extension BackupsClient: DependencyKey {
...
-				case .existingProfile:
-					return await profileStore.profile.snapshot()
-				default:
-					loggerGlobal.error("Expected to find persisted profile in ProfileStore, but was \(profileOutcome)")
-					struct NoPersistedProfile: Error {}
-					throw NoPersistedProfile()
-				}
+				await profileStore.profile.snapshot()
 			},
 			loadProfileBackups: { () -> ProfileSnapshot.HeaderList? in
```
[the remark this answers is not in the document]
Sajjon: Out of scope for this PR is to change clients methods that CAN be sync now, probably best done in a follow up PR. 

But ProfileStore still remains an actor, and I think that is good. At least writing to Profile... But we can follow up later!

RadixWallet/Clients/ProfileStore/ProfileStore.swift
```diff
@@ -1,182 +1,75 @@
...
-///     func getLoadProfileOutcome() async -> LoadProfileOutcome
-///     var profile: Profile (async)
-///     func values() -> AnyAsyncSequence<Profile>
-///     func commitEphemeral() async throws
-///     func deleteProfile() async throws
-///     func update(profile: Profile) async throws
-///     func importProfileSnapshot(:ProfileSnapshot) async throws
-///
-///
 public final actor ProfileStore {
```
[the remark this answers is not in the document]
Sajjon: Indeed, and then we can use `@ProfileStore` if we want to... but not sure it is something we want. But can mark it!

RadixWallet/Clients/ProfileStore/ProfileStore.swift
```diff
@@ -288,364 +223,306 @@ extension ProfileStore {
...
-		// Always update the header along with the snapshot itelf,
-		// so we are sure that the Header in Snapshot is synced with the Header in the HeadersList
-		try await updateProfileHeadersList(snapshot)
-		try await secureStorageClient.saveProfileSnapshot(snapshot)
+	/// Updates the in-memory copy of profile in ProfileStores and saves it, by
+	/// updates `headerList` (Keychain),  `activeProfileID` (UserDefaults)
+	/// and saves a snapshot of the profile into Keychain.
+	/// - Parameter profile: Profile to save
+	private func _saveProfileAndEmitUpdate(_ profile: Profile) throws {
+		profileSubject.send(profile)
```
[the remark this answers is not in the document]
Sajjon: good catch

RadixWallet/Clients/ProfileStore/ProfileStore.swift
Sajjon: haha thx :)

RadixWallet/Clients/ProfileStore/ProfileStore.swift
```diff
@@ -185,100 +78,142 @@ extension ProfileStore {
...
-			loggerGlobal.critical("Failed to save newly generated device identifier, error: \(error)")
+	public enum Error: String, Swift.Error, Equatable {
+		case profileIDMismatch
+		case profileUsedOnAnotherDevice
+	}
+}
+
+// MARK: "Private"
+extension ProfileStore {
+	func _lens<Property>(
```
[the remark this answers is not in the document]
Sajjon: Drawback of Xcode only project (mono target) is that `internal` suddenly is visible to the whole app... which we dont want. And since Swift Core Team forwhatever reason still do not wanna give use `typeprivate` access modifier, we have to retort to `internal` access modifiers when we split methods between files.... :/ :/ :/ I use `_` prefix to mark that it is not something other types should call

RadixWallet/Clients/ProfileStore/ProfileStore.swift
Sajjon: so name and content is consistent...? FIRST update header, then profile. Name says "saveProfile AFTER Update header" right?

I think `saveProfiel` is hiding to much, more happens... `updateWallet:withProfile` ? hmm

RadixWallet/Clients/ProfileStore/ProfileStore.swift
Sajjon: Screenshot 2023-10-24 at 10 33 38

yes

RadixWallet/Clients/ProfileStore/ProfileStore.swift
Sajjon: Mmm lets leave it to Matt

RadixWallet/Clients/ProfileStore/ProfileStore.swift
Sajjon: sure! does not help us in `init` though :/

RadixWallet/Clients/ProfileStore/ProfileStore.swift
GhenadieVP: to be consistent with the content, shouldn't it be named then `saveProfileAfterUpdatingItsHeader`? emphasis on `Update` vs `Updating`. In current form I read it as "save profile, after that update its header"

RadixWallet/Clients/ProfileStore/ProfileStore.swift
GhenadieVP: ohh, ok, so it is static. missed that.

RadixWallet/Clients/ProfileStore/ProfileStore.swift
Sajjon: sure!

RadixWallet/Clients/OnboardingClient/OnboardingClient+Interface.swift
```diff
@@ -1,22 +1,28 @@
...
-		self.commitEphemeral = commitEphemeral
+		self.finishedOnboarding = finishedOnboarding
 	}
 }
 
 extension OnboardingClient {
-	public typealias LoadProfile = @Sendable () async -> LoadProfileOutcome
+	public typealias LoadProfile = @Sendable () async -> Profile
+
+	public typealias FinishedOnboarding = @Sendable () async -> EqVoid
```
GhenadieVP, line 23: Minor: should be `FinishOnboarding`? I mean it should describe an action to perform. Same for `UnlockApp`. or maybe other naming, like "handleFinishedOnboarding"

RadixWallet/Clients/ProfileStore/ProfileStore.swift
```diff
@@ -1,284 +1,287 @@
...
+
+		let profile = try! Self._tryGenerateAndSaveNewProfile(deviceInfo: deviceInfo)
+		self.profileSubject.send(profile)
 	}
 
-	public func importCloudProfileSnapshot(_ header: ProfileSnapshot.Header) async throws {
-		try assertProfileStateIsEphemeral()
+	public func finishedOnboarding() async {
+		@Dependency(\.device) var device
+		if !profile.hasMainnetAccounts() {
```
GhenadieVP, line 152: Note: This will go away with my upcoming PR, where it is not possible to not have an account on a network.

RadixWallet/Clients/ProfileStore/ProfileStore.swift
```diff
@@ -1,284 +1,287 @@
...
+		case profileUsedOnAnotherDevice
+	}
+}
+
+// MARK: "Private"
+extension ProfileStore {
+	func _lens<Property>(
+		_ transform: @escaping @Sendable (Profile) -> Property?
+	) -> AnyAsyncSequence<Property> where Property: Sendable & Equatable {
+		profileSubject.compactMap(transform)
```
GhenadieVP, line 212: I know this probably old code, but this will update all previous profiles as well, right? If so, what for?

RadixWallet/Clients/SecureStorageClient/SecureStorageClient+Live.swift
```diff
@@ -265,77 +314,63 @@ extension SecureStorageClient: DependencyKey {
...
-				if res.wasNil {
-					loggerGlobal.info("deviceIdentifier was nil, saved new value.")
-					return identifierToSetIfNil
-				} else {
-					let loaded = try jsonDecoder().decode(UUID.self, from: res.value)
-					loggerGlobal.trace("Loaded deviceIdentifier: \(loaded)")
-					return loaded
-				}
+			getDeviceInfoSetIfNil: { _ in
+				fatalError()
```
GhenadieVP, line 350: I guess this should not be here :)

RadixWalletTests/Clients/ProfileStoreTests/ProfileStoreTests.swift
```diff
@@ -1,277 +1,1122 @@
...
+			let newProfile = await withTestClients {
+				// GIVEN no profile
+				$0.noProfile()
+			} operation: {
+				// WHEN ProfileStore.init()
+				await ProfileStore.init().profile
+			}
+
+			// THEN new profile without network is used
+			XCTAssertNoDifference(newProfile.networks.count, 0)
```
GhenadieVP, line 177: Minor, could assert if empty

RadixWallet/Clients/OnboardingClient/OnboardingClient+Interface.swift
CyonAlexRDX: yeah probably, will change

RadixWallet/Clients/ProfileStore/ProfileStore.swift
CyonAlexRDX: old profiles? no? there is only one, `CurrentValueSubject` right?

RadixWallet/Clients/SecureStorageClient/SecureStorageClient+Live.swift
CyonAlexRDX: will remove

RadixWalletTests/Clients/ProfileStoreTests/ProfileStoreTests.swift
```diff
@@ -1,277 +1,1122 @@
...
-				$0
-			}
-			$0.date = .constant(Date(timeIntervalSince1970: 0))
-			$0.userDefaultsClient.stringForKey = { _ in
-				"BABE1442-3C98-41FF-AFB0-D0F5829B020D"
+// swiftformat:disable redundantInit
+
+extension DependencyValues {
+	private mutating func _profile(_ profile: Profile?) {
+		secureStorageClient.loadProfile = { _ in
```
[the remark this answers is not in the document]
CyonAlexRDX: that would be testing two layers simultaneously, and is a testing anti-pattern. We should only test one layer at a time.

RadixWalletTests/Clients/ProfileStoreTests/ProfileStoreTests.swift
```diff
@@ -1,277 +1,1122 @@
...
-			$0.secureStorageClient.loadDeviceIdentifier = {
-				.init(uuidString: "BABE1442-3C98-41FF-AFB0-D0F5829B020D")!
+			// WHEN ProfileStore.init()
+			ProfileStore.init()
+		}
+
+		func then(_ d: inout DependencyValues) {
+			d.secureStorageClient.deleteDeprecatedDeviceID = {
+				// THEN deleteDeprecatedDeviceID is called
+				deleteDeprecatedDeviceID_is_called.fulfill()
```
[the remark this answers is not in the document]
CyonAlexRDX: it is a Void function. We should write tests for `secureStorageClient`, which asserts keychainClient remove value for key is used...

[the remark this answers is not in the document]
GhenadieVP: ok, so it does not receive the id to delete.

RadixWallet/Clients/OnboardingClient/OnboardingClient+Interface.swift
CyonAlexRDX: fixed

RadixWallet/Clients/SecureStorageClient/SecureStorageClient+Live.swift
CyonAlexRDX: fixed

RadixWalletTests/Clients/ProfileStoreTests/ProfileStoreTests.swift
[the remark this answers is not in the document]
GhenadieVP: I would worry more about the profileStore not working properly in integration with the Keychain, rather than worry about going against an anti-pattern. 

IMO, testing the two separately, ProfileStore and Keychain, does not guarantee that they would work properly when put together. Do we plan to write have some integration tests on another occasion, or we would rely on testing with the real App?

GhenadieVP (commented):
Looking very good, massive amount of work 🏅 💯!
Would have preferred some integration tests, but can be done on another occasion. 

Will perform some local testing, before given final Approval.

RadixWalletTests/Clients/ProfileStoreTests/ProfileStoreTests.swift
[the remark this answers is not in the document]
CyonAlexRDX: think perhaps we could use Reducer tests for that purpose (what you call "integration test") and use **live** AccountsClient/FactorSourcesClient etc which will be using `ProfileStore` and then as you say, use a test Keychain client.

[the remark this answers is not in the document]
GhenadieVP: Sure, but if we cannot test ProfileStore with a test keychain easily, the same will happen for those clients.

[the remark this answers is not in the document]
CyonAlexRDX: we can, it is just a dependency... not sure what your point is?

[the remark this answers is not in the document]
GhenadieVP: I mean testing with a real Keychain, not with mock of the KeychainClient.

Currently, we only test with a real Keychain when running the App. Given how crucial is this part of the app, it would definitely benefit to have tests against a real Keychain, if this is possible.

When ProfileStore mutates a profile, it really would be best to test that it is actually saved in the Keychain, by loading it back and asserting against it, instead of mocking the secureStorageClient save and load profile functions.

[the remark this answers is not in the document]
CyonAlexRDX: iOS Simulator Keychain is not a real keychain, so cannot be done.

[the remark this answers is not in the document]
CyonAlexRDX: But in the cake with 6 layers:
CreateAccount+Reducer
AccountsClient
ProfileStore
SecureStorageClient
KeychainClient
Keychain

We can test all of the top 5 layers! That is very good! and should be enough. If our KeychainClient+Live impl is thin and stupid enough (it is!) then we are good, since we use KeychainAccess

GhenadieVP (approved):
LGTM in local testing.

RadixWallet/Clients/AccountsClient/AccountsClient+Live.swift
```diff
@@ -54,19 +54,19 @@ extension AccountsClient: DependencyKey {
...
 				do {
-					let network = try await getProfileStore().profile.network(id: networkID)
+					let network = try await profileStore.profile.network(id: networkID)
 					// N.B. `accounts` is NonEmpty so `isEmpty` should always evaluate to `false`.
 					return !network.accounts.isEmpty
 				} catch {
 					return false
 				}
 			},
 			updateAccount: { updatedAccount in
```
kugel3, line 68: This one might also be a bit risky, if someone calls it with a large number of accounts it will probably break TCA sending. Perhaps we could have `updateAccounts` instead? Or at least add a yield.

CyonAlexRDX: Yes let's use a bulk add.

CyonAlexRDX: it is not used anywhere currently with a collection of changes. So it is more a tip for @GhenadieVP in his Hide Accounts PR to use a bulk and yes he is already doing that:

https://github.com/radixdlt/babylon-wallet-ios/pull/871/files#diff-e0bc333f55b4429f352323b6796ac0bab3ed78975edea4e7ca17523620f26da9R22

it is a single call for `N` many accounts, not `N` many calls.

kugel3: Yeah, I just want to discourage people getting into that situation in the future.

RadixWallet/Clients/DeviceFactorSourceClient/DeviceFactorSourceClient+Live.swift
```diff
@@ -20,7 +21,7 @@ extension DeviceFactorSourceClient: DependencyKey {
 				// FIXME: Uh this aint pretty... but we are short on time.
 				if let overridingSnapshot = maybeSnapshot {
```
kugel3, line 22: I'm not a fan of naming optionals `maybe-`, to me it is clear from context that's it's optional, but i guess it's a matter of taste

RadixWallet/Clients/DeviceFactorSourceClient/DeviceFactorSourceClient+Live.swift
```diff
@@ -20,7 +21,7 @@ extension DeviceFactorSourceClient: DependencyKey {
...
 					let networkID = Radix.Gateway.default.network.id
-					let profile = try Profile(snapshot: overridingSnapshot)
+					let profile = Profile(snapshot: overridingSnapshot)
 					let network = try profile.network(id: networkID)
 					accounts = network.accounts.elements
 					personas = network.personas.elements
```
kugel3, line 27: Could this be a function that returns accounts and personas, or nil (alternatively, throws)?

RadixWallet/Clients/DeviceFactorSourceClient/DeviceFactorSourceClient+Live.swift
```diff
@@ -39,9 +40,14 @@ extension DeviceFactorSourceClient: DependencyKey {
 					unsecuredEntityControl.transactionSigning.factorSourceID == factorSource.id
 				}
 			}
+
+			let isMnemonicMarkedAsBackedUp = userDefaultsClient.getFactorSourceIDOfBackedUpMnemonics().contains(factorSource.id)
```
kugel3, line 44: Back**ed**Up I guess?

RadixWallet/Clients/FactorSourcesClient/FactorSourcesClient+Live.swift
```diff
@@ -3,16 +3,16 @@ extension FactorSourcesClient: DependencyKey {
 	public typealias Value = FactorSourcesClient
 
 	public static func live(
-		profileStore getProfileStore: @escaping @Sendable () async -> ProfileStore = { await .shared }
+		profileStore: ProfileStore = .shared
```
kugel3, line 6: So much nicer

RadixWallet/Clients/KeychainClient/KeychainActor.swift
```diff
@@ -1,134 +1,123 @@
...
-@_spi(KeychainInternal)
-public final actor KeychainActor: GlobalActor {
-	@_spi(KeychainInternal)
-	public static let shared = KeychainActor()
+private let keychainService = {
+	var defaultService = "Radix Wallet" // DO NOT CHANGE THIS EVER
+	#if DEBUG
+	defaultService += " DEBUG"
+	#endif
+	return defaultService
```
kugel3, line 6: Or just
```
#if DEBUG
private let keychainService = "Radix Wallet DEBUG"
#else
private let keychainService = "Radix Wallet"
#endif
```

RadixWallet/Clients/LocalAuthenticationClient/LocalAuthenticationClient+Live.swift
```diff
@@ -26,44 +26,33 @@ extension LAContext {
...
-		try await withCheckedThrowingContinuation { [weak self] (continuation: CheckedContinuation<Bool, Swift.Error>) in
-			guard let self else {
-				continuation.resume(throwing: Error.contextDeinitialized)
-				return
-			}
-			var error: NSError?
-			let canEvaluate = self.canEvaluatePolicy(policy, error: &error)
+	) throws -> Bool {
+		var error: NSError?
+		let canEvaluate = self.canEvaluatePolicy(policy, error: &error)
```
kugel3, line 31: Why explicit self?

RadixWallet/Clients/LocalAuthenticationClient/LocalAuthenticationClient+Live.swift
```diff
@@ -26,44 +26,33 @@ extension LAContext {
...
-			case let .failure(error):
-				continuation.resume(throwing: error)
-				return
-			}
+		switch result {
+		case let .success(canEvaluate_):
+			return canEvaluate_
+		case let .failure(error):
+			throw error
 		}
```
kugel3, line 51: Isn't this the same as `return try result.get()`?

RadixWallet/Clients/ProfileStore/ProfileStore+AsyncSequence+Updates.swift
```diff
@@ -0,0 +1,66 @@
+
+
+// MARK: Public
+extension ProfileStore {
+	/// The current network if any
+	public func network() async throws -> Profile.Network {
+		try profile.network(id: profile.networkID)
```
kugel3, line 7: This looks like it would be more natural on profile itself?

kugel3: Why is it async by the way, old habits?

RadixWallet/Clients/ProfileStore/ProfileStore+AsyncSequence+Updates.swift
```diff
@@ -0,0 +1,66 @@
...
+	}
+
+	public var network: Profile.Network? {
+		profile.network
+	}
+
+	/// A multicasting replaying async sequence of distinct Profile.
+	public func values() async -> AnyAsyncSequence<Profile> {
```
kugel3, line 15: Same here, why async?

RadixWallet/Clients/ProfileStore/ProfileStore+AsyncSequence+Updates.swift
```diff
@@ -0,0 +1,66 @@
...
+		_lens { $0 }
+	}
+
+	/// A multicasting replaying async sequence of distinct Accounts for the currently selected network.
+	public func accountValues() async -> AnyAsyncSequence<Profile.Network.Accounts> {
```
kugel3, line 20: And here

RadixWallet/Clients/ProfileStore/ProfileStore+AsyncSequence+Updates.swift
```diff
@@ -0,0 +1,66 @@
...
+
+	/// A multicasting replaying async sequence of distinct FactorSources
+	public func factorSourcesValues() async -> AnyAsyncSequence<FactorSources> {
+		_lens {
+			$0.factorSources
+		}
+	}
+
+	@_disfavoredOverload
+	private func lens<Property>(
```
kugel3, line 55: These lenses are not used right now, right?

kugel3 (approved):
Awesome!

RadixWallet/Clients/DeviceFactorSourceClient/DeviceFactorSourceClient+Live.swift
CyonAlexRDX: Ghenadie has replaced this cod in his branch :)

RadixWallet/Clients/LocalAuthenticationClient/LocalAuthenticationClient+Live.swift
CyonAlexRDX: fixed

RadixWallet/Clients/LocalAuthenticationClient/LocalAuthenticationClient+Live.swift
CyonAlexRDX: yes, I think we had some logging there earlier... fixing

RadixWallet/Clients/ProfileStore/ProfileStore+AsyncSequence+Updates.swift
CyonAlexRDX: yes they are used by all our clients still.

kugel3: How though, they are both private!

RadixWallet/Clients/ProfileStore/ProfileStore+AsyncSequence+Updates.swift
CyonAlexRDX: Yes these should be removed. 

But the `async` here makes no difference from outside of actor isolation, i.e. `AccountsClient` still MUST write `let asyncSeq = await profileStore.values()` to get the async sequence, since `values` is an actor isolated method, and must remain so since it access (an actor isolated) stored property - `profileSubject` (all stored properties of an actor must be actor isolated).

RadixWallet/Clients/ProfileStore/ProfileStore+AsyncSequence+Updates.swift
CyonAlexRDX: yes? but these 5 use them...

```swift
	/// A multicasting replaying AsyncSequence of distinct Profile.
	public func values() -> AnyAsyncSequence<Profile> {
		_lens { $0 }
	}

	/// A multicasting replaying AsyncSequence of distinct Accounts for the currently selected network.
	public func accountValues() -> AnyAsyncSequence<Profile.Network.Accounts> {
		_lens {
			$0.network?.accounts
		}
	}

	/// A multicasting replaying AsyncSequence of distinct Personas for the currently selected network.
	public func personaValues() -> AnyAsyncSequence<Profile.Network.Personas> {
		_lens {
			$0.network?.personas
		}
	}

	/// A multicasting replaying AsyncSequence of distinct Gateways
	public func currentGatewayValues() -> AnyAsyncSequence<Radix.Gateway> {
		_lens {
			$0.appPreferences.gateways.current
		}
	}

	/// A multicasting replaying AsyncSequence of distinct Gateways
	public func gatewaysValues() -> AnyAsyncSequence<Gateways> {
		_lens {
			$0.appPreferences.gateways
		}
	}

	/// A multicasting replaying AsyncSequence of distinct FactorSources
	public func factorSourcesValues() -> AnyAsyncSequence<FactorSources> {
		_lens {
			$0.factorSources
		}
	}

```

RadixWallet/Clients/ProfileStore/ProfileStore+AsyncSequence+Updates.swift
CyonAlexRDX: fixed

RadixWallet/Clients/DeviceFactorSourceClient/DeviceFactorSourceClient+Live.swift
CyonAlexRDX: not important enough to mess with my `number of commits: 123, number of changed files: 123` miracle!

RadixWallet/Clients/ProfileStore/ProfileStore+AsyncSequence+Updates.swift
kugel3: But that's the underscored method, which takes a function. Looks like they could be using the keypath version though

Issue #820, opened 2023-10-02
Three-letter BIP39 word always returns ambiguous lookup result?


dapperdachshunds:
A user on Telegram is facing an issue where the iOS wallet keeps telling him some words are invalid. These words happen to be three-letter words. He claims to have confirmed that the words are correct and on the BIP39 word list.

I did some digging, and while I'm not very familiar with Swift and can't actually test this myself, I may have some pointers to where this issue (if it exists) might come from. It looks to me like any three-letter word will always return an ambiguous lookup result, which in turn results in a failed validation state.

- When the word field loses focus (e.g. you go to the next one), it triggers a lookup of your input here: https://github.com/radixdlt/babylon-wallet-ios/blob/7b925413a24c850e0a029d74ab5ebd1c0bffb731/Sources/Features/ImportMnemonic/ImportMnemonic.swift#L291-L292
- If the result of that lookup is that the word is known, but ambiguous (as in, there are more results that start with those letters), which is the case for at least some three-letter words, it sets hasFailedValidation to true: https://github.com/radixdlt/babylon-wallet-ios/blob/7b925413a24c850e0a029d74ab5ebd1c0bffb731/Sources/Features/ImportMnemonic/ImportMnemonic.swift#L293-L295
- My guess is that the 'ambiguous' result is returned here, because there are simply more words starting with those three letters: https://github.com/radixdlt/babylon-wallet-ios/blob/7b925413a24c850e0a029d74ab5ebd1c0bffb731/Sources/Cryptography/Mnemonic/BIP39/BIP39%2BWordList.swift#L147-L149
... as a result of this simple 'starts with' filter here: https://github.com/radixdlt/babylon-wallet-ios/blob/7b925413a24c850e0a029d74ab5ebd1c0bffb731/Sources/Cryptography/Mnemonic/BIP39/BIP39%2BWordList.swift#L131

For the English BIP39 list, lookups start becoming unambiguous from 4 characters or more if I'm not mistaken, so from that point it shouldn't be an issue.

Possible solution: check for an exact match first, which should cover exactly the case for a three-letter word (since >=4 letters is unambiguous and <3 letters doesn't match any words so that's always invalid)

stultulon:
I managed to reproduce this and now I feel mentally ill. Here’s what I found:

[...]

PR #870, opened 2023-10-26
[ABW-2532] Fix issue with 3 letter bip39 word when losing focus

CyonAlexRDX:
Fixes https://github.com/radixdlt/babylon-wallet-ios/issues/820

# Description
This PR fixes a bug where some 3 letter words are not counted as valid when TextField loses focus, e.g. "add", which is a valid BIP39 word. "add" has some _word friends_ like "addict", also starting with "add", the bug was that since "add" was not unambigously identified (due to friend "addict") we incorrectly did not set the `ImportMnemonicWord.value` to `.complete` upon losing focus.

Apart from the fix, some unit tests capturing this has been added.

https://github.com/radixdlt/babylon-wallet-ios/assets/116169792/92bf3d1e-d55b-41e5-aac5-f3fded67837f


RadixWallet/Features/SettingsFeature/ImportFromOlympiaLegacyWallet/Coordinator/ImportOlympiaWalletCoordinator+View.swift
```diff
@@ -51,9 +51,7 @@ extension ImportOlympiaWalletCoordinator.Path {
 						/State.importMnemonic,
 						action: Action.importMnemonic,
 						then: { childStore in
-							NavigationView {
-								ImportMnemonic.View(store: childStore)
-							}
+							ImportMnemonic.View(store: childStore)
```
CyonAlexRDX, line 54: @kugel3 reverting the change you made recently, which broke toolbar in Import from Olympia, which combined with the 3 letter bug, prevents users from inputting some mnemonics (containing 3 letter words which are ambigious with 3 letters)

Do you remember why you made the change? relating to navigation title perhaps?

kugel3: Yes, there should be a navigation view at the top level, it looks very strange without it. But can't we have it _only_ there, and ditch the dummy one?

Sajjon: ImportMnemonic is used from many places in the app:
* Olympia migration
* Restore wallet - part of ImportManyMnemonicsFlow
* Import from Account (should be merge with bottom)
* Import from Display Seed Phrases

So having it in ImportMnemonic itself ensure toolbar is always visible disregarding of starting point. But I guess we have no option but to manually ensure it is visible on all places...? Bah.... ought to be an easier way of having a toolbar :/ 

Anyway I will take a look

CyonAlexRDX: fixing toolbar was tricker than I had hoped, so will be done in a separate PR,

RadixWallet/Features/ImportMnemonic/ImportMnemonic.swift
```diff
@@ -310,7 +310,15 @@ public struct ImportMnemonic: Sendable, FeatureReducer {
 
 		case let .word(id, child: .delegate(.lostFocus(displayText))):
 			switch lookup(input: displayText, state) {
-			case .known(.ambigous), .unknown(.notInList):
+			case let .known(.ambigous(candidates, input)):
+				if let exactMatch = candidates.first(where: { $0.word == input }) {
+					state.words[id: id]?.value = .complete(text: displayText, word: exactMatch, completion: ImportMnemonicWord.State.WordValue.Completion.auto(match: .exact))
```
GhenadieVP, line 315: multiple lines please

GhenadieVP (approved):

PR #871, opened 2023-10-26
Hide/unhide accounts/personas

GhenadieVP:
Accounts and Personas hide/unhide functionality.

## What was done:
- Protect the access to Profile.Network.accounts and Profile.Network.personas, by exposing their read and write through dedicated functions, ensuring that the proper data is used in the app. Mostly needed, because the accounts and personas when hidden, those are not deleted, still need to be kept in the Profile.
- Implemented the hide logic for accounts:
    - When account is hidden we do add the proper flag, and then it is hidden from the app. It will not be used/displayed in any context -> Home Screen, Transfer select accounts, Seed Phrases screen and so on.
    - When an account is hidden, it is proactively deleted from the authorised dapps shared accounts.
    - Hidding accounts does not impact the derivation of new accounts addresses, hidden accounts are still account for when determining the next account index.
- Implemented the personas hide logic:
    -  When persona is hidden we do add the proper flag, and then it is hidden from the app. It will not be used/displayed in any context -> Personas, Authorised Dapps, Login Requests and so on.
    - When a persona is hidden, it is proactively deleted from the authorized dapps references.
    - By implication, if an authorized dapp no longer has any personas references  -> it is removed.
    - Hiding the personas does not impact the logic of showing the First Time Persona screen.
    - Hiding personas does not impact the derivation of new accounts addresses, hidden accounts are still account for when determining the next persona index.

### Hide/unhide accounts

https://github.com/radixdlt/babylon-wallet-ios/assets/118184705/a96d0729-06f3-4344-84ae-29d5e5ecc427

### Hide/unhide accounts restore

https://github.com/radixdlt/babylon-wallet-ios/assets/118184705/9f400721-4ff4-47c9-9463-7e2a6ff59927


RadixWallet/Clients/AccountsClient/AccountsClient+Live.swift
```diff
@@ -14,18 +14,20 @@ extension AccountsClient: DependencyKey {
...
 		}
 
-		let getAccountsOnCurrentNetwork: GetAccountsOnCurrentNetwork = {
-			try await profileStore.network().accounts
-		}
-
 		let getCurrentNetworkID: GetCurrentNetworkID = { await profileStore.profile.networkID }
 
-		let getAccountsOnNetwork: GetAccountsOnNetwork = { try await profileStore.profile.network(id: $0).accounts }
+		let getAccountsOnNetwork: GetAccountsOnNetwork = {
```
CyonAlexRDX, line 19: I think we ought to rename the "get accounts" methods "getUnhiddenAccounts" to make it clear in code also a reminder that accounts can be hidden.

RadixWallet/Clients/DeviceFactorSourceClient/DeviceFactorSourceClient+Live.swift
```diff
@@ -20,10 +20,10 @@ extension DeviceFactorSourceClient: DependencyKey {
 				// FIXME: Uh this aint pretty... but we are short on time.
 				if let overridingSnapshot = maybeSnapshot {
 					let networkID = Radix.Gateway.default.network.id
-					let profile = try Profile(snapshot: overridingSnapshot)
+					let profile = Profile(snapshot: overridingSnapshot)
 					let network = try profile.network(id: networkID)
-					accounts = network.accounts.elements
-					personas = network.personas.elements
+					accounts = network.getAccounts().elements
```
CyonAlexRDX, line 25: `getUnhiddenAccounts()` and `getUnhiddenPersonas()`

CyonAlexRDX: or `NonHidden`?

RadixWalletTests/ProfileTests/TestCases/EntitiesHidingTests/EntitiesHidingTests.swift
```diff
@@ -0,0 +1,9 @@
+import EngineToolkit
+@testable import Radix_Wallet_Dev
+import XCTest
+
+final class EntitiesHidingTests: TestCase {
+	func test_hideAccount() {
+		let profile =
```
CyonAlexRDX, line 7: poor compiler...

RadixWalletTests/Features/AccountPreferencesTests/AccountPreferencesTests.swift
```diff
@@ -0,0 +1,50 @@
+@testable import Radix_Wallet_Dev
+import XCTest
+
+@MainActor
+final class AccountPreferencesTests: TestCase {
+	func test_hideAccount_flow() async {
```
CyonAlexRDX, line 6: split this test into multiple tests. Dont care about code duplication to much. And please use GIVEN WHEN THEN format, with comment in code as well. 

We should almost always only have ONE single assert! and with name of test reflecting that assert. That makes tests easier to read, maintain and fix once broken. Also makes it clear in CI log which part of an otherwise big test like this which actually failed.

CyonAlexRDX: remember for `TestStore` you can use `exhaustivity = .off`

RadixWallet/Clients/AccountsClient/AccountsClient+Live.swift
GhenadieVP: I deliberately wanted to have just Accounts, meaning the accounts that can be used, don't want to put there "unhidden", or any other qualifier, maybe in the future we will have some other conditions for not showing some account.

RadixWalletTests/ProfileTests/TestCases/EntitiesHidingTests/EntitiesHidingTests.swift
GhenadieVP: :D need to add this

RadixWallet/Clients/AccountsClient/AccountsClient+Live.swift
CyonAlexRDX: tribal knowledge, so add `/// comment` ?

RadixWallet/Profile/Entity/Account.swift
```diff
@@ -65,6 +65,10 @@ extension Profile.Network {
 		/// A required non empty display name, used by presentation layer and sent to Dapps when requested.
 		public var displayName: NonEmpty<String>
 
+		/// Flags that are currently set on entity.
+		@DefaultCodable.EmptyCollection
+		public var flags: Set<EntityFlag>
```
Sajjon, line 70: Set will be painful once we add more Flags, since JSOn testing will be broken since order is unstable. Can we use OrderedSet? Or do you wanna defer the pain? (Non-deterministic random JSON encoding unit test failures)

GhenadieVP: I did update to IdentifiedArray in the mean time

RadixWallet/Clients/AccountsClient/AccountsClient+Live.swift
kugel3: Since we are using "hide" as an alternative to deletion, I think it is quite clear that any unadorned method will just deal with the non-hidden accounts.

RadixWallet/Clients/DeviceFactorSourceClient/DeviceFactorSourceClient+Live.swift
kugel3: The default should be to get the visible accounts, a future method to get all also the hidden accounts should have the qualifier

RadixWallet/Clients/AccountsClient/AccountsClient+Live.swift
```diff
@@ -60,7 +62,7 @@ extension AccountsClient: DependencyKey {
 				do {
 					let network = try await profileStore.profile.network(id: networkID)
 					// N.B. `accounts` is NonEmpty so `isEmpty` should always evaluate to `false`.
-					return !network.accounts.isEmpty
+					return network.hasAnyAccount()
```
kugel3, line 65: `hasSomeAccount` seems more idiomatic

RadixWallet/Clients/EntitiesVisibilityClient/EntitiesVisibilityClient+Interface.swift
```diff
@@ -0,0 +1,20 @@
...
+extension EntitiesVisibilityClient {
+	public struct HiddenEntitiesStats: Hashable, Sendable {
+		public let hiddenAccountsCount: Int
+		public let hiddenPersonasCount: Int
+	}
+
+	public typealias HideAccount = @Sendable (Profile.Network.Account) async throws -> Void
+	public typealias HidePersona = @Sendable (Profile.Network.Persona) async throws -> Void
+	public typealias UnhideAllEntities = @Sendable () async throws -> Void
+	public typealias GetHiddenEntitiesStats = @Sendable () async throws -> HiddenEntitiesStats
```
kugel3, line 19: entityCount seems more natural, two counts doesn't seem to qualify as "statistics"...

RadixWallet/Clients/EntitiesVisibilityClient/EntitiesVisibilityClient+Test.swift
```diff
@@ -0,0 +1,23 @@
+extension DependencyValues {
+	public var entitiesVisibilityClient: EntitiesVisibilityClient {
```
kugel3, line 2: Do we really need another client for this?

RadixWallet/Clients/OverlayWindowClient/OverlayWindowClient+Interface.swift
```diff
@@ -89,8 +89,15 @@ extension OverlayWindowClient {
...
 
+			// FIXME: Strings
 			public static let seedPhraseImported = Self(text: "Seed Phrase Imported")
 
+			// FIXME: Strings
+			public static let accountHidden = Self(text: "Account hidden")
+
+			// FIXME: Strings
+			public static let personaHidden = Self(text: "Persona hidden")
+
```
kugel3, line 100: I moved `updated` to the place where it was used, it seems that all of these are single use things, so it seemed more natural to put them at the call site. That way they can also go into the corresponding L10n sub struct. Or are there any particular reasons to put them here?

RadixWallet/Clients/PersonasClient/PersonasClient+Interface.swift
```diff
@@ -9,6 +9,7 @@ public struct PersonasClient: Sendable {
 
 	public var saveVirtualPersona: SaveVirtualPersona
 	public var hasAnyPersonaOnAnyNetwork: HasAnyPersonaOnAnyNetworks
+	public var hasAnyPersonaOnCurrentNetwork: HasAnyPersonaOnCurrentNetwork
```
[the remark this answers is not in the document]
kugel3: or actually just "hasPersona..." is enough

RadixWallet/Clients/PersonasClient/PersonasClient+Interface.swift
```diff
@@ -52,4 +56,16 @@ extension PersonasClient {
 	public struct PersonaNotFoundError: Error {
 		let id: Profile.Network.Persona.ID
 	}
+
+	public func determinePersonaPrimacy() async -> PersonaPrimacy {
+		let hasAnyPersonaOnAnyNetwork = await hasAnyPersonaOnAnyNetwork()
+		let hasAnyPersonaOnCurrentNetwork = await hasAnyPersonaOnCurrentNetwork()
+		let isFirstPersonaOnAnyNetwork = !hasAnyPersonaOnAnyNetwork
```
kugel3, line 63: some

RadixWallet/Core/Resources/Generated/AssetResource.generated.swift
```diff
@@ -92,6 +92,7 @@ public enum AssetResource {
   public static let unknownComponent = ImageAsset(name: "unknown-component")
   public static let xrd = ImageAsset(name: "xrd")
   public static let accountSecurity = ImageAsset(name: "accountSecurity")
+  public static let accountsPersonaHidding = ImageAsset(name: "accounts-persona-hidding")
```
kugel3, line 95: hiding?

RadixWallet/Features/AccountAndPersonaHidingFeature/AccountAndPersonaHiding+Reducer.swift
```diff
@@ -0,0 +1,81 @@
+// MARK: - AccountAndPersonaHiding
+public struct AccountAndPersonaHiding: FeatureReducer {
+	public struct State: Hashable, Sendable {
+		public var hiddenEntitiesStats: EntitiesVisibilityClient.HiddenEntitiesStats?
```
kugel3, line 4: Here also I don't think "stats" is appropriate...

RadixWallet/Features/AccountAndPersonaHidingFeature/AccountAndPersonaHiding+Reducer.swift
```diff
@@ -0,0 +1,81 @@
...
+		switch viewAction {
+		case .task:
+			return .run { send in
+				let hiddenEntitiesStats = try await entitiesVisibilityClient.getHiddenEntitiesStats()
+				await send(.internal(.hiddenEntitesStatsLoaded(hiddenEntitiesStats)))
+			}
+		case .unhideAllTapped:
+			state.confirmUnhideAllAlert = .init(
+				title: .init(L10n.AppSettings.EntityHiding.unhideAllSection),
+				message: .init(L10n.AppSettings.EntityHiding.unhideAllConfirmation),
```
kugel3, line 41: I suppose the image should also be called "entity-hiding"?

RadixWallet/Features/AccountAndPersonaHidingFeature/AccountAndPersonaHiding+View.swift
```diff
@@ -0,0 +1,93 @@
...
+}
+
+extension AccountAndPersonaHiding {
+	public struct ViewState: Equatable {
+		public let hiddenAccountsCount: Int
+		public let hiddenPersonasCount: Int
+
+		public var hiddenAccountsText: String {
+			if hiddenAccountsCount == 1 {
+				L10n.AppSettings.EntityHiding.hiddenAccount(1)
```
kugel3, line 17: Why does this string take an argument?

RadixWallet/Features/AccountListFeature/Coordinator/AccountList+Reducer.swift
```diff
@@ -6,20 +6,12 @@ public struct AccountList: Sendable, FeatureReducer {
 	public struct State: Sendable, Hashable {
 		public var accounts: IdentifiedArrayOf<AccountList.Row.State>
 
-		public init(accounts: IdentifiedArrayOf<AccountList.Row.State>) {
-			self.accounts = accounts
-		}
-
 		public init() {
-			self.accounts = []
+			self.init(accounts: [])
```
kugel3, line 10: Why is `accounts` an identifiable array in the first place? Unless we want to manipulate it directly in this reducer, that's not needed.

RadixWallet/Features/AccountPreferencesFeature/AccountPreferences+View.swift
```diff
@@ -58,13 +60,24 @@ extension AccountPreferences {
 	}
 }
 
+extension AccountPreferences.View {
+	@MainActor
+	func hideAccountButton(_ viewStore: ViewStoreOf<AccountPreferences>) -> some View {
```
kugel3, line 65: ViewStore is not needed for sending anymore, only for reading

RadixWallet/Features/DappInteractionFeature/Children/Login/Coordinator/Login.swift
```diff
@@ -7,7 +7,7 @@ struct Login: Sendable, FeatureReducer {
 		let dappMetadata: DappMetadata
 		let loginRequest: P2P.Dapp.Request.AuthLoginRequestItem
 
-		var isFirstPersonaOnAnyNetwork: Bool? = nil
+		var personaPrimacy: PersonaPrimacy? = nil
```
kugel3, line 10: As nice as this sounds, it's not entirely clear what persona primacy means!

kugel3: Aha, so it's no longer a bool, but then at least put a doc comment on the type explaining what it is

RadixWallet/Features/DappInteractionFeature/Children/OneTimePersonaData/OneTimePersonaData.swift
```diff
@@ -97,22 +97,21 @@ struct OneTimePersonaData: Sendable, FeatureReducer {
 			}
 
 		case .appeared:
-			return checkIfFirstPersonaByUserEver()
+			return .run { send in
+				await send(.internal(.personaPrimacyDetermined(personasClient.determinePersonaPrimacy())))
```
kugel3, line 101: I'd say this would be slightly clearer on two lines

RadixWallet/Features/SettingsFeature/AppSettings/AppSettings+Reducer.swift
```diff
@@ -102,6 +108,10 @@ public struct AppSettings: Sendable, FeatureReducer {
 			state.destination = .profileBackupSettings(.init())
 			return .none
 
+		case .accountAndPersonaHidingButtonTapped:
+			state.destination = .accountAndPersonasHidding(.init())
```
kugel3, line 112: hiding

RadixWallet/Prelude/DefaultValueDecodable.swift
```diff
@@ -0,0 +1,82 @@
...
+/// Provides the Type of the decoded value, also its default value
+public protocol DefaultValueProvider {
+	associatedtype Value: Codable
+	static var defaultValue: Value { get }
+}
+
+// MARK: - DefaultCodable
+public enum DefaultCodable {
+	/// The property wrapper itself.
+	/// Need to make use of the tagging with the protocol, since the value provided in `init` will not be
```
kugel3, line 16: I'm pretty sure I've written DefaultCodable in the simple straightforward way, are you sure it's not possible to just provide the value as an argument?
```
@DefaultCodable([])
var flags: Flags
```

RadixWallet/Prelude/Extensions/Collection+Extra.swift
```diff
@@ -32,3 +32,19 @@ extension OrderedSet where Element: Hashable {
...
+	/// Mutates in place the elements of the collection
+	public mutating func mutateAll(_ mutate: (inout Self.Element) -> Void) {
+		for index in indices {
+			mutate(&self[index])
+		}
+	}
+}
+
+extension RangeReplaceableCollection {
+	/// Filters in place the elements of the collection
```
kugel3, line 46: This still creates a new array though, right? I suppose you could loop over the array and remove the elements you don't want?

kugel3 (approved):
Excellent!

I think we should unify the naming

personaAccountHiding
personasAndAccountsHiding
entityHiding

I prefer the last one

RadixWallet/Prelude/DefaultValueDecodable.swift
kugel3: Or maybe not actually, seems like I never manage to solve that at the property level, instead I made a type level protocol `DefaultingType` that in this case `Flags` would conform to. Not necessarily better than your solution, but I'll paste it here for comparison:

```

import Foundation

// MARK: - DefaultingType

/**
 `DefaultingType` is a protocol that makes sure that a type, when used in a property, is never `nil`, even if the value is missing when decoding.
 Its requirement is that the type has a static var `defaultValue` of type `Self`. For enums, this can also be fulfilled by having an enum case called `defaultValue`.

     struct Location: DefaultingType {
         let lat: Int
         let lon: Int

         static let defaultValue: Self = .init(lat: 0, lon: 0)
     }

     struct User: Decodable {
         // Will never be decoded to nil
         let location: Location

         // Any corrupted values will be decoded to defaultValue
         let locations: [Location]
     }
 */
public protocol DefaultingType: Decodable {
  static var defaultValue: Self { get }
}

/// A helper type to decode `DefaultingType` values
private struct DefaultingWrapper<T: DefaultingType>: Decodable {
  let value: T

  public init(from decoder: Decoder) throws {
    self.value = (try? .init(from: decoder)) ?? .defaultValue
  }
}

extension KeyedDecodingContainer {
  public func decode<T: DefaultingType>(_ type: T.Type, forKey key: Self.Key) throws -> T {
    (try? decode(DefaultingWrapper.self, forKey: key).value) ?? .defaultValue
  }

  public func decodeIfPresent<T: DefaultingType>(_ type: T.Type, forKey key: Self.Key) throws -> T? {
    try decodeIfPresent(DefaultingWrapper.self, forKey: key)?.value
  }

  public func decode<T: DefaultingType>(_ type: [T].Type, forKey key: Self.Key) throws -> [T] {
    guard let result = try? decode([DefaultingWrapper<T>].self, forKey: key) else { return [] }
    return result.map(\.value)
  }

  public func decodeIfPresent<T: DefaultingType>(_ type: [T].Type, forKey key: Self.Key) throws -> [T]? {
    guard let result = try? decodeIfPresent([DefaultingWrapper<T>].self, forKey: key) else { return nil }
    return result.map(\.value)
  }
}
```

CyonAlexRDX: I changed \`\`\` into \`\`\`swift" for Swift colour highlighting

RadixWallet/Clients/EntitiesVisibilityClient/EntitiesVisibilityClient+Interface.swift
CyonAlexRDX: I agree with Gustaf

RadixWallet/Clients/EntitiesVisibilityClient/EntitiesVisibilityClient+Test.swift
CyonAlexRDX: We discussed merging PersonasClient and AccountsClients before, at least that is what I suggest, which under the hood (in the Live Impl) can share the same impl of a lot of the functions, or most APIs work with `some EntityProtocol` instead of concrete types. 

And then these new functions would be simple `hideEntity` instead of `hideAccount` and `hidePersona`. 

But I think it can be done as a tech debt item after this PR. i.e. pragmatic fastest way forward is to merge this as is, and then you merge them all into `EntitiesClient`?

RadixWallet/Features/AccountAndPersonaHidingFeature/AccountAndPersonaHiding+View.swift
CyonAlexRDX: rather this while function could be: 

```swift
public var hiddenAccountsText: String {
    L10n.AppSettings.EntityHiding.hiddenAccount(hiddenAccountsCount)
}
```

right?

RadixWallet/Prelude/Extensions/Collection+Extra.swift
CyonAlexRDX, line 46: Better English: `"/// Filters the elements in place"` ("in place" last).

RadixWallet/Profile/Entity/EntityProtocol.swift
```diff
@@ -2,6 +2,8 @@ import EngineToolkit
 
 // MARK: - EntityBaseProtocol
 public protocol EntityBaseProtocol {
+	typealias Flags = IdentifiedArrayOf<EntityFlag>
```
CyonAlexRDX, line 5: I think `OrderedSet` is preferable over `IdentifiedArray` is it not? We use Identified when it is a struct with an ID right? When the ID *is* the `value` then it is the same thing as `OrderedSet`, just more confusing to user IdentifiedArrayOf? Right?

CyonAlexRDX (approved):
Nice! I think Flags should be OrderedSet.

RadixWallet/Profile/Entity/EntityProtocol.swift
kugel3: I maintain that a flags variable should be a Set, or better yet, an OptionSet, but that might be overkill.

- When you only care about inclusion vs exclusion, and it should be unique:
Set or OptionSet (the latter gives some syntactic sugar)

- When you only care about order:
Array

- When you only care about order but the elements have to be identified because of ForEach
still just Array

- When you care about order, but also want uniqueness of elements (and O(1) lookup):
OrderedSet

- When you want explicit keys and ignore order (and want O(1) ops)
Dictionary

- When you want explicit keys and order (and want O(1) ops)
OrderedDictionary

- When you want implicit keys and order (and want O(1) ops)
IdentifiedArray

GhenadieVP: We cannot use Sets as we do have to be compatible with Android encoding/decoding, where:
`flags: Set<EntityFlag>` is encoded as `"flags":["deletedByUser"]`.

Swift encoder is different, where `Set<EntityFlag>` or `OrderedSet<EntityFlag>` is encoded as:
```
"flags": {
    {
       "deletedByUser": {}
    }
}
``

File: RadixWallet/Prelude/DefaultValueDecodable.swift (hunk 21 above)
Comment (GhenadieVP) [reply RadixWallet/Prelude/DefaultValueDecodable.swift:16]: @kugel3 Problem with being able to do just:
```
@DefaultCodable([])
var flags: Flags
```

Is that the given default value will not be used when initialised with Decoder, and it makes sense, since it is a different initialiser. So we have to use this whole dance with protocols.

File: RadixWallet/Clients/EntitiesVisibilityClient/EntitiesVisibilityClient+Test.swift (hunk 8 above)
Comment (GhenadieVP) [reply RadixWallet/Clients/EntitiesVisibilityClient/EntitiesVisibilityClient+Test.swift:2]: Yes, this Client is mostly needed since we kinda need a merge of PersonasClient and AccountsClients, and as @CyonAlexRDX suggested we could probably have EntitiesClient, but I'll leave it for another PR.

File: RadixWallet/Clients/PersonasClient/PersonasClient+Interface.swift (hunk 10 above)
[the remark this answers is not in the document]
Comment (GhenadieVP) [reply RadixWallet/Clients/PersonasClient/PersonasClient+Interface.swift:12]: Used `some`, as `hasPersona` kinda tells there can be only one Persona.

File: RadixWallet/Features/AccountAndPersonaHidingFeature/AccountAndPersonaHiding+Reducer.swift (hunk 14 above)
Comment (GhenadieVP) [reply RadixWallet/Features/AccountAndPersonaHidingFeature/AccountAndPersonaHiding+Reducer.swift:41]: yeah, I think it just uses the name of the resource I did download, will update.

File: RadixWallet/Features/AccountAndPersonaHidingFeature/AccountAndPersonaHiding+View.swift (hunk 15 above)
Comment (GhenadieVP) [reply RadixWallet/Features/AccountAndPersonaHidingFeature/AccountAndPersonaHiding+View.swift:17]: This is how currently the string is added in Crowding, was done by Android, will update once the string is also updated.

@CyonAlexRDX about just having:
```
public var hiddenAccountsText: String {
    L10n.AppSettings.EntityHiding.hiddenAccount(hiddenAccountsCount)
}
```
I am not sure if Crowding + L10n can handle the multiplicity with just one function, but probably it can, don't know.

File: RadixWallet/Features/AccountListFeature/Coordinator/AccountList+Reducer.swift (hunk 16 above)
Comment (GhenadieVP) [reply RadixWallet/Features/AccountListFeature/Coordinator/AccountList+Reducer.swift:10]: I think it is for `ForEach`, kept the same definitions.

File: RadixWallet/Prelude/DefaultValueDecodable.swift (hunk 21 above)
Comment (kugel3) [reply RadixWallet/Prelude/DefaultValueDecodable.swift:16]: Yeah I remembered that when I looked at my old code, that's why I also had to resort to type level solutions. But there are discussions on Swift evolution that might fix this

File: RadixWallet/Profile/Entity/EntityProtocol.swift (hunk 23 above)
Comment (kugel3) [reply RadixWallet/Profile/Entity/EntityProtocol.swift:5]: We could always do it manually, but it doesn't seem worth it. So sets are encoded as dictionaries with the elements as keys, and void as values?

Comment (GhenadieVP) [reply RadixWallet/Profile/Entity/EntityProtocol.swift:5]: > So sets are encoded as dictionaries with the elements as keys, and void as values?
Yes, so it seems.

Comment (kugel3) [reply RadixWallet/Profile/Entity/EntityProtocol.swift:5]: It seems Set is encoded and decoded just like Array in Swift, try this:
```Swift
	struct TestStruct: Codable {
		let set: Set<String>
		let array: [String]
	}

	func test_encode() throws {
		let value = TestStruct(set: ["one", "two"], array: ["three", "four"])
		let encoded = try JSONEncoder().encode(value)
		print("ENCODED:\n\(String(data: encoded, encoding: .utf8))")
	}

	func test_decode() throws {
		let string = "{\"array\":[\"three\",\"four\"],\"set\":[\"one\",\"two\"]}"
		guard let data = string.data(using: .utf8) else { return }
		let decoded = try JSONDecoder().decode(TestStruct.self, from: data)
		print("DECODED:\n\(decoded)")
	}
```

kugel3: Works even for OrderedSet

kugel3: Ah, that's surprising, but I guess enums are hard to encode for swift

PR #877, opened 2023-10-30
Deposit instruction based on deposit rules

GhenadieVP:
Jira issue: https://radixdlt.atlassian.net/browse/ABW-2361

When transferrin between user accounts, determine the deposit instruction(and by implication the requirement of additional signature) based on the account deposit rule.
If an user owned account, they are about to deposit into, would allow the deposit to happen without requiring a signature the Wallet will not add it.

# Demos

## AcceptAll deposit rule:
- Resource has no exception -> `try_deposit_or_abort` instruction is used.
- Resource has deny exception -> `deposit` instruction is used.

https://github.com/radixdlt/babylon-wallet-ios/assets/118184705/89aee54d-9671-4cf5-9b7a-1d0964130249

## AcceptKnown deposit rule
- Resource is known to the account and it has no exception -> `try_deposit_or_abort` instruction is used.
- Resource is known to the account and it has deny exception -> `deposit` instruction is used.
- Resource is not known to the account and it has no exception -> `deposit` instruction is used.
- Resource is not known to the account and it has allow exception -> `try_deposit_or_abort` instruction is used.

https://github.com/radixdlt/babylon-wallet-ios/assets/118184705/f41f8e5e-3caf-4e22-9151-acab7ba15db9

## DenyAll deposit rule
- Resource has no exception -> `deposit` instruction is used.
- Resource has allow exception -> `try_deposit_or_abort` instruction is used.

https://github.com/radixdlt/babylon-wallet-ios/assets/118184705/293377da-c36f-42f9-a755-d0336e55172c


RadixWallet/Clients/OnLedgerEntitiesClient/Helpers/OnLedgerEntitiesClient+CreateEntity.swift
```diff
@@ -502,6 +494,34 @@ extension OnLedgerEntity.Account.PoolUnitResources {
...
+	public var filterEmptyVaults: OnLedgerEntity.OwnedFungibleResources {
+		.init(
+			xrdResource: xrdResource.flatMap { $0.amount > 0 ? $0 : nil },
+			nonXrdResources: nonXrdResources.filter { $0.amount > 0 }
+		)
+	}
+}
+
+extension OnLedgerEntity.Account {
+	public var filterEmptyVaults: OnLedgerEntity.Account {
```
GhenadieVP, line 513: Now we filter the empty vault resource at reducer level, as with this PR we still need to keep track of resource with empty vaults to know if a given resource was ever owned by the account.

RadixWallet/Clients/OnLedgerEntitiesClient/Helpers/OnLedgerEntitiesClient+CreateEntity.swift
```diff
@@ -502,6 +494,34 @@ extension OnLedgerEntity.Account.PoolUnitResources {
 	}
 }
 
+extension [OnLedgerEntity.OwnedNonFungibleResource] {
+	public var filterEmptyVaults: [OnLedgerEntity.OwnedNonFungibleResource] {
```
kugel3, line 498: I think something with nonEmptyVaults is more correct, since "filter" is ambiguous (because of how `filter` works. If we definitely want to use the word filter, the correct form would be "filtering", since this is not an in place filter method.

RadixWallet/Clients/OnLedgerEntitiesClient/OnLedgerEntitiesClient+Live.swift
```diff
@@ -374,15 +374,15 @@ extension Address {
 extension OnLedgerEntity.Account.PoolUnitResources {
 	// The fungible resources used to build up the pool units.
 	// Will be used to filter out those from the general fungible resources list.
-	var fungibleResourceAddresses: [String] {
-		radixNetworkStakes.compactMap(\.stakeUnitResource?.resourceAddress.address) +
-			poolUnits.map(\.resource.resourceAddress.address)
+	var fungibleResourceAddresses: [ResourceAddress] {
```
kugel3, line 377: Better than strings!

RadixWallet/Features/AssetTransferFeature/AssetTransfer+Reducer.swift
```diff
@@ -232,30 +234,79 @@ extension AssetTransfer {
...
-	let userHasAccessToMnemonic = if let deviceFactorSourceID = receivingAccount.left?.deviceFactorSourceID {
-		await secureStorageClient.containsMnemonicIdentifiedByFactorSourceID(deviceFactorSourceID)
-	} else { false }
 
-	guard isUserAccount, isSoftwareAccount, userHasAccessToMnemonic else {
+	if case let .left(userAccount) = receivingAccount {
+		@Dependency(\.secureStorageClient) var secureStorageClient
+		// TODO: Temporary revert of checking if the receiving account is a ledger account
+		let isSoftwareAccount = true // !receivingAccount.isLedgerAccount
+		let userHasAccessToMnemonic = if let deviceFactorSourceID = userAccount.deviceFactorSourceID {
```
kugel3, line 246: I understand that we want to be modern and use Swift 5.9 but `= if let` hurts my eyes. Couldn't this just use `??`?

kugel3 (approved):
Github seems to think there is an error in a test

RadixWalletTests/Features/AssetTransferTests/AssetTransferDepositRuleTests.swift
```diff
@@ -0,0 +1,143 @@
...
+				amount: .one(),
+				metadata: .init(nil)
+			)
+		),
+		nonFungibleResources: [],
+		poolUnitResources: .init(radixNetworkStakes: [], poolUnits: [])
+	)
+
+	// MARK: - AcceptAll deposit rule
+	func test_GIVEN_depositRuleIsAcceptAll_WHEN_resourceHasNoException_THEN_noSignatureIsRequired() async {
```
CyonAlexRDX, line 27: minor, but I highlight the GIVEN WHEN THEN in test names by used of `__` instead of `_` thus making tests easier to read.

Also annotating `// GIVEN` `// WHEN` and `// THEN` in the body of the test, *matching* the method name, greatly increases readability!

RadixWalletTests/Features/AssetTransferTests/AssetTransferDepositRuleTests.swift
CyonAlexRDX, line 1: All tests here are async, so all tests should use the new `try await withTimeLimit {` that I added part of the ProfileStoreTests, have a look.

The plan here is that for ALL tests which are marked `async` should use `withTimeLimit`. So to ensure we never accidentally write tests which do not complete.

CyonAlexRDX (changes requested):
nice! Need to change tests to use `withTimeLimit`, also what is the plan for:

```
	// TODO: Temporary revert of checking if the receiving account is a ledger account
		let isSoftwareAccount = true // !receivingAccount.isLedgerAccount
```
not to be addressed by this PR? do we have another ticket for that? or should we not at all have logic be dependent on if ledger or not?

RadixWallet/Features/AssetTransferFeature/AssetTransfer+Reducer.swift
```diff
@@ -232,30 +234,79 @@ extension AssetTransfer {
...
 	let recipientAddress = receivingAccount.address
-	let userHasAccessToMnemonic = if let deviceFactorSourceID = receivingAccount.left?.deviceFactorSourceID {
-		await secureStorageClient.containsMnemonicIdentifiedByFactorSourceID(deviceFactorSourceID)
-	} else { false }
 
-	guard isUserAccount, isSoftwareAccount, userHasAccessToMnemonic else {
+	if case let .left(userAccount) = receivingAccount {
+		@Dependency(\.secureStorageClient) var secureStorageClient
+		// TODO: Temporary revert of checking if the receiving account is a ledger account
+		let isSoftwareAccount = true // !receivingAccount.isLedgerAccount
```
CyonAlexRDX, line 245: should we remove "is ledger" logic completely? or is there another ticket for it?

GhenadieVP: well, I didn't want to modify the current setup. And I don't fully recall the details around ledger accounts.

RadixWalletTests/Features/AssetTransferTests/AssetTransferDepositRuleTests.swift
GhenadieVP: It would be nice to use such tools as https://github.com/Quick/Quick to do this kind of testing.

RadixWalletTests/Features/AssetTransferTests/AssetTransferDepositRuleTests.swift
```diff
@@ -24,120 +24,167 @@ final class AssetTransferDepositRuleTests: TestCase {
...
+	) async throws {
+		try await assertSignatureIsRequired(for: account, isRequired: false, onLedgerAccounts: onLedgerAccounts)
 	}
 
-	private func assertSignatureIsRequired(for account: Profile.Network.Account, isRequired: Bool, onLedgerAccounts: [OnLedgerEntity.Account]) async {
-		await withDependencies { d in
-			d.onLedgerEntitiesClient.getEntities = { _, _, _, _ in
-				onLedgerAccounts.map { .account($0) }
+	private func assertSignatureIsRequired(for account: Profile.Network.Account, isRequired: Bool, onLedgerAccounts: [OnLedgerEntity.Account]) async throws {
+		try await withTimeLimit(.fast) {
```
CyonAlexRDX, line 179: @GhenadieVP hmm I think best to have it in each test... like I did, else one might do async stuff later in test, and that would defeat the purpose of this.

CyonAlexRDX (approved):
Approved, good job!

PR #878, opened 2023-10-31
[ABW-2206] Merge AccountList into `Home`, clean up Export/Import mnemonic prompts and display

CyonAlexRDX:
[ABW-2206](https://radixdlt.atlassian.net/browse/ABW-2206)

## Description
The intended behaviour was not to import/export mnemonics from AccountDetails, but to navigate to Seed Phrases screen in Settings.

# Changes
* Clicking "Import" button in Seed Phrases starts import Mnemonics (plural) Flow as intended
* Added a "Imported Seed Phrase" HUD (`overlayWindowClient`) when we import mnemonics
* Merge `AccountList` feature into `Home` (calling the row reducers: `Home.AccountRow`)
* Introduce a protocol impl by `AccountDetails.State` and `Home.AccountRow` for shared stated using `AccountWithInfo` struct, sharing logic like `exportMnemonicNeeded` and `importMnemonicNeeded` etc
* EXTRA: Bring back `changeToMainnetIfNeeded` functionality part of import profile, which was accidentally removed in previous #860 PR

# Demo

## Import Mnemonic
https://github.com/radixdlt/babylon-wallet-ios/assets/116169792/07ccbdd4-d60c-45d8-88a3-f699dbc52452

## Export Mnemonc
https://github.com/radixdlt/babylon-wallet-ios/assets/116169792/8bd68fa9-cda5-472f-b070-fdb8abac56cd

## Assert we show export mnemonic when XRD -> `>0`
https://github.com/radixdlt/babylon-wallet-ios/assets/116169792/37526d43-09fc-4f95-8164-559bbfca34e5

## Seed Phrases screen import works as expected

https://github.com/radixdlt/babylon-wallet-ios/assets/116169792/bd0e0ca1-b141-46c4-845e-db0a875f3928

## Seed Phrases screen export works as expected

https://github.com/radixdlt/babylon-wallet-ios/assets/116169792/175d8860-d259-4100-a9da-4a8f04558d10

## Import mnemonic by doing Legacy wallet import 
https://github.com/radixdlt/babylon-wallet-ios/assets/116169792/80f0fabd-99dd-42ee-b9d8-e08aa286c5c2

## Export prompt after faucet usage on Account Details
https://github.com/radixdlt/babylon-wallet-ios/assets/116169792/e4ede58b-81b5-4934-b292-c6c652105825

[...]

PR #879, opened 2023-10-31
Dont tell user mnemonic M is missing when they are about to import it...

CyonAlexRDX:
# Description
We have a bug where the "Could Not Complete" _missing mnemonic_  alert is shown when user is importing an Olympia Wallet with mnemonic.

# Changes
* This PR changes to `notifyIfMissing: false` when we are checking if Olympia Mnemonic already exists.
* Adds a loading indicator in the `Import` button in ImportMnemonic feature which disabled the button and shows loader while the mnemonic is being imported
* Small infra improvement, prevent showing confusing error due to account recovery check being done on empty profile

## BEFORE (bug)
https://github.com/radixdlt/babylon-wallet-ios/assets/116169792/84336205-b374-4da7-85c9-ace84626ec62

## AFTER (also added loader)

https://github.com/radixdlt/babylon-wallet-ios/assets/116169792/74f7d17d-c20e-4f64-bbc0-999b58e6d36f

### DIFF
--- a/RadixWallet/Features/ImportMnemonic/ImportMnemonic.swift
+++ b/RadixWallet/Features/ImportMnemonic/ImportMnemonic.swift
@@ -380,7 +380,6 @@ public struct ImportMnemonic: Sendable, FeatureReducer {
 			return .send(.delegate(.doneViewing(markedMnemonicAsBackedUp: false)))
 
 		case .destination(.presented(.onContinueWarning(.buttonTapped))):
-			state.destination = nil
 			guard let mnemonic = state.mnemonic else {
 				loggerGlobal.error("Can't read mnemonic")
 				struct FailedToReadMnemonic: Error {}
--- a/RadixWallet/Features/ImportMnemonic/ImportMnemonic.swift
+++ b/RadixWallet/Features/ImportMnemonic/ImportMnemonic.swift
@@ -60,6 +60,9 @@ public struct ImportMnemonic: Sendable, FeatureReducer {
 			words.compactMap(\.completeWord)
 		}
 
+		/// irrelevant for readonly
+		public var isProgressing = false
+
 		public let persistStrategy: PersistStrategy?
 
 		public let readonlyMode: ReadonlyMode?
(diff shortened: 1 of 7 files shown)


RadixWallet/Features/ImportMnemonic/ImportMnemonic.swift
```diff
@@ -60,6 +60,9 @@ public struct ImportMnemonic: Sendable, FeatureReducer {
 			words.compactMap(\.completeWord)
 		}
 
+		/// irrelevant for readonly
+		public var isProgressing = false
```
CyonAlexRDX, line 64: hard to name this one, cannot be named `isSavingMnemonic` since if `persistStrategy` is `nil`, we are in fact not saving. Hmm maybe could be named `isImportOfMnemonicInProgress` ? but I think this is OK a name.

RadixWallet/Features/ImportMnemonic/ImportMnemonic.swift
```diff
@@ -445,6 +448,7 @@ public struct ImportMnemonic: Sendable, FeatureReducer {
 			state.debugMnemonicPhraseSingleField = mnemonic
 			if let mnemonic = try? Mnemonic(phrase: mnemonic, language: state.language) {
 				state.words = State.words(from: mnemonic, isReadonlyMode: state.readonlyMode != nil)
+				state.isProgressing = true
```
CyonAlexRDX, line 451: not important, since DEBUG only, but why not.

RadixWallet/Features/SplashFeature/Splash.swift
```diff
@@ -103,11 +103,15 @@ public struct Splash: Sendable, FeatureReducer {
...
 			return .run { send in
-				await send(.internal(.loadProfile(onboardingClient.loadProfile())))
+				await send(.internal(.loadedProfile(onboardingClient.loadProfile())))
 			}
 
-		case .loadProfile:
-			return checkAccountRecoveryNeeded()
+		case let .loadedProfile(profile):
+			if profile.networks.isEmpty {
+				return delegateCompleted(accountRecoveryNeeded: false)
```
CyonAlexRDX, line 111: no need to checkAccountRecovery for empty "ephemeral" profile.

RadixWallet/Features/ImportMnemonic/ImportMnemonic.swift
kugel3: `var isSavingMnemonic: Bool? = nil`

GhenadieVP (approved):
LGTM!

RadixWallet/Features/ImportMnemonic/ImportMnemonic.swift
CyonAlexRDX: sure, changing in my other mnemonics PR https://github.com/radixdlt/babylon-wallet-ios/pull/878

PR #880, opened 2023-10-31
Fix/abw 2365 i os messages for transactions submission end state

GhenadieVP:
Jira ticket: paste link here

Complements / depends on / relates to #link to other PR

## Description
This PR solves / fixes / adds / includes functionality 1 and functionality 2.

### Notes
Additional notes.

1. Step 1
2. Step 2
3. Verify

1. Step 1
2. Step 2
3. Verify

## Screenshot

| - | - |

## Video
paste link here

## PR submission checklist


RadixWallet/Clients/SubmitTransactionClient/SubmitTransactionClient+Interface.swift
```diff
@@ -50,17 +27,15 @@ public struct SubmitTXRequest: Sendable, Hashable {
...
-	public let maxPollTries: Int
 	public let sleepDuration: TimeInterval
-	public init(maxPollTries: Int, sleepDuration: TimeInterval) {
-		self.maxPollTries = maxPollTries
+	public init(sleepDuration: TimeInterval) {
 		self.sleepDuration = sleepDuration
 	}
 
-	public static let `default` = Self(maxPollTries: 20, sleepDuration: 2)
+	public static let `default` = Self(sleepDuration: 2)
```
CyonAlexRDX, line 40: have we checked with GW about this change? We should really impl incremental backing off strategy, perhaps as easy as adding 0.5 or 1 seconds to sleep duration after each poll attempt.

RadixWallet/Clients/SubmitTransactionClient/SubmitTransactionClient+Live.swift
```diff
@@ -5,79 +5,42 @@ extension SubmitTransactionClient: DependencyKey {
...
-							await statusSubject.send(.init(
-								txID: txID,
-								result: .failure(.failedToGetTransactionStatus(txID: txID, error: .init(pollAttempts: pollCountHolder.value)))
-							))
-						default:
-							/// Continue polling, the poll failed due to internal URLSession error.
-							break
-						}
-					}
+			while true {
```
CyonAlexRDX, line 19: can be made into a function we recursively call and add 0.5 or 1 sec after each attempt, as to not load GW to much.

RadixWallet/Clients/SubmitTransactionClient/SubmitTransactionClient+Interface.swift
GhenadieVP: yes, we do perform infinite polling, and stop only when we have certain transaction status returned.

RadixWallet/Clients/SubmitTransactionClient/SubmitTransactionClient+Live.swift
GhenadieVP: we could do that, but this pretty much keeps the same behaviour as we currently have in production. Will ask the network team about the load on the transaction status endpoint, I assume it is quite lightweight.

RadixWallet/Features/TransactionReviewFeature/SubmitTransaction/SubmitTransaction.swift
```diff
@@ -117,93 +122,104 @@ public struct SubmitTransaction: Sendable, FeatureReducer {
...
-
-	var isSubmitted: Bool {
-		switch self {
-		case .failedToGetStatus, .rejected, .committedFailure, .submittedUnknown, .submittedPending, .committedSuccessfully: true
-		case .submitting, .notYetSubmitted: false
-		}
-	}
+	private func transactionCommittedSuccesfully(_ state: State) -> Effect<Action> {
+		// TODO: Could probably be moved in other place. TransactionClient? AccountPortfolio?
+		Task.detached { [intent = state.notarizedTX.intent] in
```
CyonAlexRDX, line 159: hmm yes this is wrong, cannot be here in Reducer, even though `detached` I THINK this will be cancelled when the reducer feature is dismissed? did you check? can probably be put into a client (again)?

CyonAlexRDX (approved):
Nice! Approved, but before merge, lets resolve:
1. Task.detached in reducer
2. incremental backing off?

RadixWallet/Features/TransactionReviewFeature/SubmitTransaction/SubmitTransaction.swift
GhenadieVP: I did check, `detached` task is not cancelled, since it is a top-level task. Moved to AccountPortfoliosClient.

RadixWallet/Clients/SubmitTransactionClient/SubmitTransactionClient+Live.swift
GhenadieVP: Added 1 second increment.
Did keep the `while` loop instead of recursion, as to not grow the call stack.

PR #884, opened 2023-11-02
Fix NFT pagination

GhenadieVP:
iOS 17.1 introduced a hard crash when dismissing multiple screens and the intermediary screens have the toolbar involved.

ref -> https://developer.apple.com/forums/thread/738362

Added the safe version of the toolbar.


kugel3 (approved):
Reluctantly

CyonAlexRDX (approved):
LGTM, but did you check if Mnemonics word suggestions works with this? Maybe it even fixes some scenarios were word suggestions was NOT shown, that would be awesome 🤩

PR #903, opened 2023-11-13
Improved verification that user has written down seed phrase

GhenadieVP:
Ask user to verify their seed phrase after backing up.

# What was done
- A new VerifyMnemonic view is added.
- Randomly select 3 words + the checksum word to be verified by the user.
- Add proper focus for the words that need to be added.
- When screen is opened the first word is focused.
- Tapping `return` on the keyboard, jump to the next empty word to facilitate user entry.

# Demo

https://github.com/radixdlt/babylon-wallet-ios/assets/118184705/e9358eb9-abf3-48dd-9374-f5f597e6cd8c


RadixWallet/Features/HomeFeature/Coordinator/Home+View.swift
```diff
@@ -113,7 +113,7 @@ private extension View {
 			state: /Home.Destination.State.exportMnemonic,
 			action: Home.Destination.Action.exportMnemonic,
 			content: { childStore in
-				NavigationView {
+				NavigationStack {
```
Sajjon, line 116: Does this have any effect on toolbar visibility? :)

RadixWallet/Features/HomeFeature/Children/Mnemonics+Utils.swift
```diff
@@ -1,7 +1,8 @@
 extension ExportMnemonic.State {
 	static func export(
 		_ input: SimplePrivateFactorSource,
-		title: String
+		title: String,
+		context: ReadonlyMode.Context
```
GhenadieVP, line 5: fix to properly send the context the mnemonic is shown from

RadixWallet/Features/ImportMnemonic/ImportMnemonic.swift
```diff
@@ -299,12 +316,15 @@ public struct ImportMnemonic: Sendable, FeatureReducer {
 		}
 
 		public var body: some Reducer<State, Action> {
-			Scope(state: /State.markMnemonicAsBackedUp, action: /Action.markMnemonicAsBackedUp) {
+			Scope(state: /State.backupConfimartion, action: /Action.backupConfimartion) {
 				EmptyReducer()
```
Sajjon, line 320: The scoping with EmptyReducer can be removed completely. See recent Slack thread.

RadixWallet/Features/ImportMnemonic/VerifyMnemonic/VerifyMnemonic+View.swift
```diff
@@ -0,0 +1,154 @@
...
+				}
+				.scrollIndicators(.hidden)
+			}
+			// FIXME: Strings
+			.navigationTitle(L10n.ConfirmMnemonicBackedUp.title)
+		}
+
+		@ViewBuilder
+		private func wordsGrid(viewStore: ViewStoreOf<VerifyMnemonic>) -> some SwiftUI.View {
+			SwiftUI.Grid(horizontalSpacing: .small2, verticalSpacing: .medium1) {
```
GhenadieVP, line 90: Use of Grid instead of LazyVGrid so that the content properly scrolls to last word.
Otherwise, if the random 3 words are from first 10 words, the last word would have been loaded due to Lazy mechanism, thus the focus on the last word will not happen until the user explicitly scrolls so the view is loaded.

RadixWallet/Features/ImportMnemonic/VerifyMnemonic/VerifyMnemonic+View.swift
```diff
@@ -0,0 +1,154 @@
+import SwiftUI
+
+extension VerifyMnemonic.State {
+	public var viewState: VerifyMnemonic.ViewState {
+		let enumeratedWords = mnemonic.words.identifiablyEnumerated()
+		let wordViewStates = enumeratedWords.map { word in
+			let isConfirmationWord = wordsToConfirm.contains(word)
+			return VerifyMnemonic.ViewState.WordViewState(
+				word: word,
+				placeholder: isConfirmationWord ? "" : "•••••",
```
Sajjon, line 10: See Slack, we agreed to use 4, not 5, dots, matching the BIP39 specification that all words (in English word list) are unambiguously identified by the first 4 chars.

RadixWallet/Prelude/Extensions/IdentifiedArrayOf+Extensions.swift
```diff
@@ -5,3 +5,44 @@ extension IdentifiedArrayOf {
...
+extension Collection where Element: Identifiable {
+	public func identified() throws -> IdentifiedArrayOf<Element> {
+		guard Set(map(\.id)).count == count else {
+			throw IdentifiedArrayError.clashingIDs
+		}
+		return .init(uniqueElements: self)
+	}
+
+	public func uniqueIdentified() -> IdentifiedArrayOf<Element> {
+		.init(uncheckedUniqueElements: self)
```
Sajjon, line 24: this crashes for duplicates... we ought to never crash in PROD. Best to never use `unchecked`, it has bit us in the a** in the past. hmm is it used? can we filter out copies in PROD instead (but crash in DEBUG?)?

RadixWallet/Features/ImportMnemonic/VerifyMnemonic/VerifyMnemonic.swift
```diff
@@ -0,0 +1,93 @@
...
+
+			let identifiedWords = mnemonic.words.identifiablyEnumerated()
+			let checksumWord = identifiedWords.last!
+			let randomWords = identifiedWords
+				.dropLast() // without checksum word
+				.shuffled() // randomize
+				.prefix(Self.numberOfWordsToConfirm) // take the required number of words
+				.sorted { $0.offset < $1.offset } // sort after shuffling
+
+			self.wordsToConfirm = .init(randomWords + [checksumWord])!
```
Sajjon, line 26: lets never use `+`, it is crazy slow for compiler.

RadixWallet/Features/ImportMnemonic/VerifyMnemonic/VerifyMnemonic.swift
```diff
@@ -0,0 +1,93 @@
+import ComposableArchitecture
+import SwiftUI
+
+// MARK: - DisplayMnemonic
+public struct VerifyMnemonic: Sendable, FeatureReducer {
+	public struct State: Sendable, Hashable {
+		public static let numberOfWordsToConfirm = 3
```
Sajjon, line 7: minor, but this ought to be `numberOfRandomWordsToConfirm` since it is in fact `3 + 1`. Maybe a comment?

RadixWallet/Features/ImportMnemonic/VerifyMnemonic/VerifyMnemonic.swift
```diff
@@ -0,0 +1,93 @@
...
+	public func reduce(into state: inout State, viewAction: ViewAction) -> Effect<Action> {
+		switch viewAction {
+		case let .wordChanged(identifiedWord):
+			// reset the invalidMnemonic state
+			state.invalidMnemonic = false
+			state.enteredWords.updateOrAppend(identifiedWord)
+			return .none
+		case let .textFieldFocused(focus):
+			// Don't set focus to nil if it will be changed to another field
+			if focus == nil {
```
Sajjon, line 56: This is easier to read IMO:

```swift
let done = state.enteredWords.reduce(true) { $0 && !$1.element.isEmpty }

if let focus {
	state.focusedField = focus
} else if done {
	state.focusedField = nil
}
return .none
```

Sajjon: `done` is less misleading than `didEnterAllWords`, because `didEnterAllWords` make it sound like all inputs are known to be valid BIP39 words, but no such validation has (yet) happened right.

Sajjon: `element.isEmpty` appears 3 times in code, might be worth adding an extension:

```swift
extension OffsetIdentified where Element: Collection {
    var isEmpty: Bool { element.isEmpty }
}
```

CyonAlexRDX:
LGTM! Some stuff that ought to be changed IMO, but not necessarily blocking

RadixWallet/Features/HomeFeature/Coordinator/Home+View.swift
kugel3: Speaking of toolbar, is the idea to show the completion toolbar here, or is that too much work?

GhenadieVP: The toolbar is not needed though in this case, I don't show the word completion when verifying the mnemonic.

RadixWallet/Features/ImportMnemonic/ImportMnemonic.swift
```diff
@@ -20,6 +20,21 @@ public struct ImportMnemonic: Sendable, FeatureReducer {
 			return wordCount
 		}
 
+		public var factorSourceID: FactorSourceID.FromHash? {
```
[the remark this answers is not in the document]
GhenadieVP: need to look into it, I kept the previous behaviour here, just introduced this reusable `factorSourceID` property.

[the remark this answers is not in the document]
GhenadieVP: changed.

CyonAlexRDX (approved):
LGTM !

PR #907, opened 2023-11-14
Max XRD transfer

GhenadieVP:
JIRA - https://radixdlt.atlassian.net/browse/ABW-2497

Display contextual alerts when user tries to transfer the max amount of XRD. 

By setting the `defaultFee=1 XRD` the following logic is applied:
- When tapping Max button the remaining XRD amount is used as Max.
- If the remaining `XRD >= defaultFee`, we do display the alert allowing users to choose the amount.
- If the remaining `XRD < defaultFee`, we do display the alert informing the user that they will have to pay the fee from other account.

#Demo

https://github.com/radixdlt/babylon-wallet-ios/assets/118184705/ea855b02-29c7-4321-934d-961b9a21fb46

### DIFF
--- a/RadixWallet/Features/AssetTransferFeature/Components/TransferAccountList/ReceivingAccount/Asset/FungibleResourceAsset+Reducer.swift
+++ b/RadixWallet/Features/AssetTransferFeature/Components/TransferAccountList/ReceivingAccount/Asset/FungibleResourceAsset+Reducer.swift
@@ -24,6 +24,9 @@ public struct FungibleResourceAsset: Sendable, FeatureReducer {
 
 		// MARK: - Mutable state
 
+		@PresentationState
+		public var alert: AlertState<ViewAction.AlertAction>?
+
 		public var transferAmountStr: String = ""
 		public var transferAmount: RETDecimal? = nil
 
@@ -44,6 +47,18 @@ public struct FungibleResourceAsset: Sendable, FeatureReducer {
 		case maxAmountTapped
 		case focusChanged(Bool)
 		case removeTapped
+
+		case alertAction(PresentationAction<AlertAction>)
+
+		public enum AlertAction: Hashable, Sendable {
+			case chooseXRDAmountAlert(ChooseXRDAmountAlert)
+			case needsToPayFeeFromOtherAccount(EqVoid)
+
+			public enum ChooseXRDAmountAlert: Hashable, Sendable {
+				case deductFee(RETDecimal)
+				case sendAll(RETDecimal)
+			}
+		}
 	}
 
 	public enum DelegateAction: Equatable, Sendable {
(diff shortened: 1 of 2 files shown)


RadixWallet/Features/AssetTransferFeature/Components/TransferAccountList/ReceivingAccount/Asset/FungibleResourceAsset+Reducer.swift
```diff
@@ -44,6 +47,23 @@ public struct FungibleResourceAsset: Sendable, FeatureReducer {
 		case maxAmountTapped
 		case focusChanged(Bool)
 		case removeTapped
+
+		case alertAction(PresentationAction<AlertAction>)
```
kugel3, line 51: I think the state and action parts should have the same name.

RadixWallet/Features/AssetTransferFeature/Components/TransferAccountList/ReceivingAccount/Asset/FungibleResourceAsset+View.swift
```diff
@@ -78,6 +78,7 @@ extension FungibleResourceAsset.View {
 				}
 			}
 			.padding(.medium3)
+			.alert(store: store.scope(state: \.$alert, action: { .view(.alertAction($0)) }))
```
kugel3, line 81: Can you make a private extension on the store with an `alert` var

RadixWallet/Features/AssetTransferFeature/Components/TransferAccountList/ReceivingAccount/Asset/FungibleResourceAsset+Reducer.swift
```diff
@@ -77,6 +110,61 @@ public struct FungibleResourceAsset: Sendable, FeatureReducer {
...
+				return .none
+
+			case .dismiss:
+				return .none
+			}
 		}
 	}
 }
+
+extension AlertState where Action == FungibleResourceAsset.ViewAction.AlertAction {
```
kugel3, line 134: Just to be safe, can you use explicit types instead of .inits, those can sometimes slow down typechecking

kugel3 (approved):
I am not convinced this should even be a Feature, but if it is, I suppose we might as well put the alert there.

RadixWallet/Features/AssetTransferFeature/Components/TransferAccountList/ReceivingAccount/Asset/FungibleResourceAsset+Reducer.swift
```diff
@@ -77,6 +110,61 @@ public struct FungibleResourceAsset: Sendable, FeatureReducer {
...
+	fileprivate static func chooseXRDAmount(feeDeductedAmount: RETDecimal, maxAmount: RETDecimal) -> Self {
+		.init(
+			title: .init(L10n.AssetTransfer.MaxAmountDialog.title),
+			message: .init(L10n.AssetTransfer.MaxAmountDialog.body),
+			buttons:
+			[
+				.default(
+					.init(L10n.AssetTransfer.MaxAmountDialog.saveXrdForFeeButton(feeDeductedAmount.formatted())),
```
CyonAlexRDX, line 142: how does this look when you have 100 million XRD? (🐋)? The size of the button title will be tiny...?

CyonAlexRDX (commented):
Maybe Alert is unfit use? since button titles will be tiny for large XRD amounts..?

RadixWallet/Features/AssetTransferFeature/Components/TransferAccountList/ReceivingAccount/Asset/FungibleResourceAsset+Reducer.swift
GhenadieVP: 100 million XRD will be formatted as 100M though, thanks to @kugel3 formatter

GhenadieVP:
> Maybe Alert is unfit use? since button titles will be tiny for large XRD amounts..?

It actually looks ok, I was also reticent initially, but given that we do have proper large number formatting + max 8 decimals places - it looks ok.

PR #911, opened 2023-11-16
iOS prompt users to write down the seed phrase after creating a persona

GhenadieVP:
Prompt users to write down the seed phrase once they have created a persona.

# What has been done

1. Added the proper notifications to guide the users to personas screen where they can  backup the seed phrase.
2. Added an async stream to be used to notify screens that a seed phrase for personas needs to be written down.

# Debug only
Skip verification (saves us some time) 
https://github.com/radixdlt/babylon-wallet-ios/assets/116169792/077ed6ce-6d04-462b-9514-7541be5478af


RadixWallet/Features/DappsAndPersonas/Personas/Coordinator/PersonasCoordinator+Reducer.swift
```diff
@@ -34,6 +34,7 @@ public struct PersonasCoordinator: Sendable, FeatureReducer {
 	public enum InternalAction: Sendable & Equatable {
 		case personaPrimacyDetermined(PersonaPrimacy)
 		case loadedPersonaDetails(PersonaDetails.State)
+		case personasSeedPhraseBackedup(personaIDs: Set<Profile.Network.Persona.ID>)
```
Sajjon, line 37: *WrittenDown or *Exported or at least not typo (BackedUp).

RadixWallet/Profile/Entity/EntityProtocol.swift
```diff
@@ -144,3 +144,17 @@ extension EntityBaseProtocol {
 		flags.contains(.deletedByUser)
 	}
 }
+
+extension EntityProtocol {
+	public var deviceFactorSourceID: FactorSourceID.FromHash? {
```
Sajjon, line 149: I think we have this in multiple places...

RadixWallet/Profile/Entity/Persona.swift
```diff
@@ -166,3 +166,15 @@ extension Profile.Network.Personas {
 		filter(\.isHidden).asIdentifiable()
 	}
 }
+
+extension Profile.Network.Persona {
+	public var shouldWriteDownSeedPhrase: Bool {
```
Sajjon, line 171: This cannot be merged as is. SHOULD also check: `secureStorageClient.containsMnemonic(deviceFactorSource.id)`, since it is possible for that to be `false` (missing in keychain!) and for `backedUpIds` to not contains `deviceFactorSource.id`, in which case this current impl would return `true` for `shouldWriteDownSeedPhrase`, but the mnemonic is missing!

CyonAlexRDX (changes requested):
I will update the shouldWriteDownSeedPhrase computed property before this can be merged

CyonAlexRDX (approved):
I've fixed what needed fixing

PR #905, opened 2023-11-14
Make Xcode usable again

kugel3:
## Description
- Breaks out destination actions into its own case in FeatureReducer
- Adds a local function for scoping state on Stores, to radically improve type-checking
- Adds a local keypath for scoping state on Reducers, to radically improve type-checking
 
### Notes
Affects most reducer and view files

## How to test
Should not change any behaviour, but since it affects almost all screens, it would be good to test all navigation

## PR submission checklist
- [ ] I have tested account to account transfer flow and have confirmed that it works

PR #898, opened 2023-11-08
Improve type checking everywhere

kugel3:
## Description

Tries to improve type checking, by providing more context to the type checker. Also unifies how we handle destinations, so it looks the same everywhere.

**ADDED:** I used a built in profiling command in the terminal to find all the difficult functions, I managed to rewrite all of them except one - `body` in `DappInteractor` - because of Relay. Will do that later.

Makes other simplifications.

The next step is to move destination actions out of `ChildAction` and give them their own case in `FeatureAction`.

Applies the following pattern to all destinations (and turns non-destinations into destinations):
```swift
private extension StoreOf<ResourcesList> {
	var destination: PresentationStoreOf<ResourcesList.Destination> {
		scope(state: \.$destination) { .child(.destination($0)) }
	}
}

@MainActor
private extension View {
	func destinations(with store: StoreOf<ResourcesList>) -> some View {
		let destinationStore = store.destination
		return addAsset(with: destinationStore)
			.confirmDeletionAlert(with: destinationStore)
	}

	private func addAsset(with destinationStore: PresentationStoreOf<ResourcesList.Destination>) -> some View {
		sheet(
			store: destinationStore,
			state: /ResourcesList.Destination.State.addAsset,
			action: ResourcesList.Destination.Action.addAsset,
			content: { AddAsset.View(store: $0) }
		)
	}

	private func confirmDeletionAlert(with destinationStore: PresentationStoreOf<ResourcesList.Destination>) -> some View {
		alert(
			store: destinationStore,
			state: /ResourcesList.Destination.State.confirmAssetDeletion,
			action: ResourcesList.Destination.Action.confirmAssetDeletion
		)
	}
}
```

## How to test
This should not affect any behaviour anywhere, but it touches every single "destination", which is when we present alerts, sheets etc, or push another view onto the stack.

These things should probably be tested with extra care:

- removing gateway alert
- overlay alert
- splash alert
- authorised dapps - presentDappDetails + dismissing on forgetDapp

- ImportMnemonicsFlowCoordinator
- ImportMnemonicControllingAccounts
- AssetTransferMessage
- TransferAccountList
- introduction to personas slideup

## PR submission checklist
- [ ] I have tested account to account transfer flow and have confirmed that it works

PR #915, opened 2023-11-17
All type checking

kugel3:
## Description
All type checking fixes in one place

Merges *Improve type checking everywhere [#898](https://github.com/radixdlt/babylon-wallet-ios/pull/898)* and *Make Xcode usable again [#905](https://github.com/radixdlt/babylon-wallet-ios/pull/905)*

## How to test
This should not affect any behaviour anywhere, but it touches every single "destination", which is when we present alerts, sheets etc, or push another view onto the stack.

These things should probably be tested with extra care:

removing gateway alert

overlay alert

splash alert

authorised dapps - presentDappDetails + dismissing on forgetDapp

ImportMnemonicsFlowCoordinator

ImportMnemonicControllingAccounts

AssetTransferMessage

TransferAccountList

introduction to personas slideup

## PR submission checklist


CyonAlexRDX (approved):
Did not review this since I reviewed the two it originates from.

PR #916, opened 2023-11-17
Small fixes

GhenadieVP:
Two quick small fixes:
- Hide Dev Preferences section for mainnet.
- Add cancel button in Max XRD alert.
- Don't show max XRD alert is remaining amount is zero


CyonAlexRDX (approved):
LGTM

PR #922, opened 2023-11-21
Prevent PoolUnits crash

kugel3:
## Description
[Slack thread](https://rdxworks.slack.com/archives/C05V4202J90/p1700474625827809)

The user with account [account_rdx128ldshjmrge7j98ydag7qgsdk2t79qqlxpw5j2warzqnju7mg6x605](https://dashboard.radixdlt.com/account/account_rdx128ldshjmrge7j98ydag7qgsdk2t79qqlxpw5j2warzqnju7mg6x605/pool-units) experienced a crash when opening the pool unit view. This turned out to be because the total supply of Ociswap LP XRD/BCN was `0`, which caused a division-by-zero crash. This PR displays the text "Total supply is 0 - could not calculate redemption value", similar to how we already handle the case when the total supply property is missing.

## Screenshot
image

## PR submission checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -136,7 +136,7 @@ jobs:
       checks: write
     name: "Unit test"
     runs-on: macos-13-xlarge
-    timeout-minutes: 15
+    timeout-minutes: 18
 
     needs:
       - linting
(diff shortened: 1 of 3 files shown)


CyonAlexRDX (approved):
LGTM

PR #924, opened 2023-11-22
Upgrade to TCA 1.4.2

kugel3:
## Description
Upgrade to TCA 1.4

### Notes
This also disables macro validation on CI. Please check that this is done in a suitable point in the CI script

## How to test
Everything should keep working as before. Navigation is most likely to break, it shouldn't affect smaller details.

## PR submission checklist


.github/workflows/ci.yml
```diff
@@ -189,7 +188,9 @@ jobs:
           brew install xcbeautify
 
       - name: "Unit tests"
-        run: bundle exec fastlane tests
+        run: |
+          defaults write com.apple.dt.Xcode IDESkipMacroFingerprintValidation -bool YES
```
Sajjon, line 192: nice that it was this simple!

RadixWallet.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
```diff
@@ -306,6 +306,15 @@
         "version" : "0.5.0"
       }
     },
+    {
+      "identity" : "swift-syntax",
```
Sajjon, line 310: does this increase build time? if it does, we ought to bump timeout from 18 to.. 20 min?

RadixWallet/Features/DappInteractionFeature/Coordinator/DappInteractionFlow.swift
```diff
@@ -146,6 +146,7 @@ struct DappInteractionFlow: Sendable, FeatureReducer {
 					.ifCaseLet(/MainState.chooseAccounts, action: /MainAction.chooseAccounts) {
 						AccountPermissionChooseAccounts()
 					}
+				EmptyReducer()
```
Sajjon, line 149: intentional? is there some kind of contraint on only being able to do 3 `ifCaseLet`? Sounds arbitrary?

Hmm... should we not use Scope here btw??

kugel3: The problem here is that type checking takes too long, with the new overloads.

CyonAlexRDX (approved):
LGTM

RadixWallet/Features/DappInteractionFeature/Coordinator/DappInteractionFlow.swift
kugel3: But yes, Scope also works, so I switched to the new CaseKeyPath version of Scope

PR #927, opened 2023-11-23
[ABW-2610] Pool Unit Issues

kugel3:
Jira ticket: ABW-2610

## Description
Show all pool units

### Notes
An unpushed commit also filters out zero-amount pool units, but it's not entirely clear how to count pool units.

## PR submission checklist


CyonAlexRDX (approved):
Looks reasonable!

PR #938, opened 2023-12-05
Add DEBUG button in ImportMnemonic for Olympia legacy accounts which …

CyonAlexRDX:
https://github.com/radixdlt/babylon-wallet-ios/assets/116169792/4a952214-d772-4e65-b872-e0d788c9f171

### DIFF
--- a/RadixWallet/Features/ImportMnemonic/ImportMnemonic+View.swift
+++ b/RadixWallet/Features/ImportMnemonic/ImportMnemonic+View.swift
@@ -148,8 +148,23 @@ extension ImportMnemonic {
 							.padding(.horizontal, .medium2)
 							.padding(.bottom, .medium3)
 						} else {
-							Button("DEBUG Acc.Rec.Scan.mne (24)") {
-								viewStore.send(.debugUseTestingMnemonicWithActiveAccounts)
+							if !(viewStore.isWordCountFixed && viewStore.wordCount == .twentyFour) {
+								Button("DEBUG AccRecScan Olympia 15") {
+									viewStore.send(.debugUseOlympiaTestingMnemonicWithActiveAccounts)
+								}
+								.buttonStyle(
+									.secondaryRectangular(
+										shouldExpand: true,
+										isDestructive: true,
+										isInToolbar: true
+									)
+								)
+								.padding(.horizontal, .medium2)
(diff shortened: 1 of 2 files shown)


RadixWallet/Features/ImportMnemonic/ImportMnemonic+View.swift
```diff
@@ -148,8 +148,23 @@ extension ImportMnemonic {
 							.padding(.horizontal, .medium2)
 							.padding(.bottom, .medium3)
 						} else {
-							Button("DEBUG Acc.Rec.Scan.mne (24)") {
-								viewStore.send(.debugUseTestingMnemonicWithActiveAccounts)
+							if !(viewStore.isWordCountFixed && viewStore.wordCount == .twentyFour) {
```
kugel3, line 151: Why do we only want to show it when wordcount is 24?

RadixWallet/Features/ImportMnemonic/ImportMnemonic.swift
```diff
@@ -451,7 +452,10 @@ public struct ImportMnemonic: Sendable, FeatureReducer {
 			let toPaste = pasteboardClient.getString() ?? ""
 			return .send(.view(.debugMnemonicChanged(toPaste)))
 
-		case .debugUseTestingMnemonicWithActiveAccounts:
+		case .debugUseOlympiaTestingMnemonicWithActiveAccounts:
+			return .send(.view(.debugMnemonicChanged("section canoe half crystal crew balcony duty scout half robot avocado gas all effort piece")))
```
kugel3, line 456: I would make these local constants, but it's not important

kugel3 (approved):

RadixWallet/Features/ImportMnemonic/ImportMnemonic+View.swift
CyonAlexRDX: No this code says "don't allow using a 15 word mnemonic as Babylon Device Factor Source" - since BDFS requires 24words

RadixWallet/Features/ImportMnemonic/ImportMnemonic.swift
CyonAlexRDX: Might remove later...:

PR #951, opened 2023-12-08
Update engine toolkit

GhenadieVP:
Update RET to the latest version.


RadixWallet/Features/AssetTransferFeature/AssetTransfer+Reducer.swift
```diff
@@ -258,8 +258,8 @@ func instructionForDepositing(
 
 	return try ManifestBuilder.accountTryDepositOrAbort(
 		recipientAddress.intoEngine(),
-		nil,
-		bucket
+		bucket,
+		nil
```
CyonAlexRDX, line 262: @GhenadieVP is the flip of arguments here intended? A change in RET?

GhenadieVP: yes, the order flipped.

PR #954, opened 2023-12-11
[ABW-2606] Show Radix banner

kugel3:
Jira ticket: ABW-2606

## Description
Show a Radix banner at the bottom of the home screen when starting as new user, until it's been dismissed once.

### Notes
The button currently opens https://rdx.works

## Screenshot

image

## PR submission checklist


RadixWallet/Clients/UserDefaults+Dependency+Extension/UserDefaults+Dependency+Extension.swift
```diff
@@ -4,6 +4,7 @@ import DependenciesAdditions
 // MARK: - UserDefaultsKey
 public enum UserDefaultsKey: String, Sendable, Hashable, CaseIterable {
 	case hideMigrateOlympiaButton
+	case hideRadixBanner
```
GhenadieVP, line 7: hmm, I guess it is a matter of preference, but I would've preferred this to be `showRadixBanner` instead.

But, I presume that you did chose to have `hide...` so we can easily show the banner also for existing users, but I don't think it is needed, I recall that we should show the banner only for `new` users, we did change that?

GhenadieVP: Additionally, maybe it would be more useful to have a flag like `fresh`, `new` user? from which could derive in the future any new behaviour we want.

kugel3: I preferred hide- because I saw showing as the default, but it could be either way around.

The way I set it up now it would be shown once to everyone, and then going forward only to new users. But you might be right that we should actually not show it at all to people who are already onboarded.

GhenadieVP (approved):

PR #956, opened 2023-12-12
Show Stake Claim NFT details

GhenadieVP:
Display the stake claim nft details

# Demo

https://github.com/radixdlt/babylon-wallet-ios/assets/118184705/5895b7e6-cecb-43f0-9d04-5ac435bddf4c

### DIFF
--- a/RadixWallet/Core/FeaturePrelude/AddressView/AddressView.swift
+++ b/RadixWallet/Core/FeaturePrelude/AddressView/AddressView.swift
@@ -99,7 +99,6 @@ extension AddressView {
 			}
 		}
 		.multilineTextAlignment(.leading)
-		// .minimumScaleFactor(0.5)
 	}
 
 	private var image: Image {
--- a/RadixWallet/Core/FeaturePrelude/AddressView/AddressView.swift
+++ b/RadixWallet/Core/FeaturePrelude/AddressView/AddressView.swift
@@ -85,20 +85,20 @@ extension AddressView {
 		}
 	}
 
+	@ViewBuilder
 	private var addressView: some View {
-		Group {
-			if format == .full {
-				Text("\(identifiable.address.formatted(format))\(image)")
-					.lineLimit(nil)
-			} else {
-				HStack(spacing: .small3) {
-					Text(identifiable.address.formatted(format))
-						.lineLimit(1)
-					image
-				}
+		if format == .full {
+			Text("\(identifiable.address.formatted(format))\(image)")
+				.lineLimit(nil)
+				.multilineTextAlignment(.leading)
+				.minimumScaleFactor(0.5)
+		} else {
+			HStack(spacing: .small3) {
+				Text(identifiable.address.formatted(format))
+					.lineLimit(1)
+				image
 			}
 		}
-		.multilineTextAlignment(.leading)
 	}
 
 	private var image: Image {
--- a/RadixWallet/Core/FeaturePrelude/AddressView/AddressView.swift
+++ b/RadixWallet/Core/FeaturePrelude/AddressView/AddressView.swift
@@ -99,7 +99,7 @@ extension AddressView {
 			}
 		}
 		.multilineTextAlignment(.leading)
-		.minimumScaleFactor(0.5)
+		// .minimumScaleFactor(0.5)
 	}
 
 	private var image: Image {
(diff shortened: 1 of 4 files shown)


RadixWallet/Core/FeaturePrelude/AddressView/AddressView.swift
```diff
@@ -99,7 +99,6 @@ extension AddressView {
 			}
 		}
 		.multilineTextAlignment(.leading)
-		.minimumScaleFactor(0.5)
```
GhenadieVP, line 102: Removed, the stake claim NFT id was made very small. Prefer just truncating the text instead.

RadixWallet/Features/AssetsFeature/Components/NonFungibleAssetList/Components/Details/NonFungibleTokenDetails+View.swift
```diff
@@ -189,12 +189,13 @@ extension NonFungibleTokenDetails.ViewState.TokenDetails {
 }
 
 extension OnLedgerEntity.NonFungibleToken.NFTData {
+	private static let standardFields: [OnLedgerEntity.NonFungibleToken.NFTData.StandardField] = [.name, .description, .keyImageURL]
```
GhenadieVP, line 192: This is used for now to show the `claim_epoch` and `claim_amount` as arbitrary fields. It needs to be clarified if we should display those two fields in a different way in NFT details.

RadixWallet/Features/AssetsFeature/Components/PoolUnitsList/Components/LSUResource/LSUStake.swift
```diff
@@ -97,6 +107,12 @@ public struct LSUStake: FeatureReducer {
 			}
 
 			// TODO: Show details
```
CyonAlexRDX, line 109: is this todo comment still relevant?

RadixWallet/Core/FeaturePrelude/AddressView/AddressView.swift
kugel3: We should probably keep some kind of scale factor, otherwise adressess will never scale. How about 0.8?

RadixWallet/Features/AssetsFeature/Components/PoolUnitsList/Components/LSUResource/LSUStake.swift
GhenadieVP: definitely not.

RadixWallet/Features/AssetsFeature/Components/NonFungibleAssetList/Components/Details/NonFungibleTokenDetails+View.swift
kugel3: This is a good idea, but for most, if not all, standard fields we will want to map the key to a localised string key, right?

RadixWallet/Features/AssetsFeature/Components/PoolUnitsList/Components/LSUResource/LSUStake.swift
kugel3: That was probably my comment when I realised we weren't showing the details yet

kugel3 (approved):

RadixWallet/Features/AssetsFeature/Components/NonFungibleAssetList/Components/Details/NonFungibleTokenDetails+View.swift
GhenadieVP: Standard fields do have a predefined way of being displayed - like the `name` is shown as the sheet title, the `keyImageURL` is shown as an image. We don't actually display the name of the field, we do show its content in some predefined way.
But true, if there is ever the case of showing the name of the standard field, it would be localised.

RadixWallet/Core/FeaturePrelude/AddressView/AddressView.swift
kugel3: We do have an option for showing a full address. But in general, the multilineTextAlignment, should be moved inside group, since it only applies for the first case. And you might keep the scale factor there too.

Then you can ditch the Group and make it a @ViewBuilder instead

RadixWallet/Features/AssetsFeature/Components/NonFungibleAssetList/Components/Details/NonFungibleTokenDetails+View.swift
kugel3: Sure, for those ones, but this PR makes me think that there might be fields that should be displayed as labeled values. I suppose actually it would make more sense to separate that concept from the standard field concept, and just have something like `fieldNameMapping: [OnLedgerEntity.NonFungibleToken.NFTData.AnyKey: LocalizedStringKey]`

GhenadieVP: ArbitraryFields are the ones that are named, but not localised. It was introduced this with my previous PR.

PR #957, opened 2023-12-12
Show Factor Source in Seed Phrases even if it controls 0 accounts

CyonAlexRDX:
(Android already does this) List a Factor Source in `App Settings -> Account Sec & Settings -> Seed Phrases` even if it does not control any account, i.e. for DFS call it `X` we with `CP_BAO` we show it as an ODFS and a BDFS, even it `X` controls 0 Babylon accounts and/or 0 Olympia accounts.

# Demo 1 - See Slack
(Video too large for GH)
[Slack](https://rdxworks.slack.com/archives/C03QFAWBRNX/p1702378108609239)

# Demo 2 - Import Mnemonics Flow
However, when prompted to Import Mnemonics we show ONLY Babylon accounts if any, if no Babylon accounts and if Olympia accounts, we show them.

Where we left off from Demo 1, the state of the Profile, is that contains 3 factor sources, all with both set of crypto parameters (support both Olympia and Babylon) see screenshot, and then watch demo video.

simulator_screenshot_C39BD87B-5F43-449A-92DA-EE8C376AEE38

We should only be asked for three mnemonics, and for the first DFS, the main one, we only see the 4 babylon accounts, not the 2 olympia accounts. But for the two remaining mnemonics we see two olympia accounts each of them controls.

https://github.com/radixdlt/babylon-wallet-ios/assets/116169792/b8b7bef3-3e74-411e-a7ad-b28e31422b5e

### DIFF
--- a/RadixWallet/Core/SharedModels/Entities/EntitiesControlledByFactorSource.swift
+++ b/RadixWallet/Core/SharedModels/Entities/EntitiesControlledByFactorSource.swift
@@ -33,12 +33,12 @@ extension EntitiesControlledByFactorSource {
 		}
 
 		public let id: ID
-		public let accounts: NonEmptyAccounts
-		public let hiddenAccounts: NonEmptyAccounts?
+		public let accounts: IdentifiedArrayOf<Profile.Network.Account>
+		public let hiddenAccounts: IdentifiedArrayOf<Profile.Network.Account>
 	}
 
 	public var olympia: AccountsControlledByKeysOnSameCurve? {
-		guard let olympiaAccounts else { return nil }
+		guard deviceFactorSource.supportsOlympia else { return nil }
 		return AccountsControlledByKeysOnSameCurve(
 			id: .init(factorSourceID: deviceFactorSource.id, isOlympia: true),
 			accounts: olympiaAccounts,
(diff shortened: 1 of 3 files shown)


RadixWallet/Core/SharedModels/Entities/EntitiesControlledByFactorSource.swift
```diff
@@ -56,23 +56,23 @@ extension EntitiesControlledByFactorSource {
...
-	public var babylonAccountsHidden: NonEmptyAccounts? {
-		NonEmpty(rawValue: hiddenAccounts.filter(not(\.isOlympiaAccount)).asIdentifiable())
+	public var babylonAccountsHidden: IdentifiedArrayOf<Profile.Network.Account> {
+		hiddenAccounts.filter(not(\.isOlympiaAccount)).asIdentifiable()
 	}
 
 	/// Non hidden
-	public var olympiaAccounts: NonEmptyAccounts? {
-		NonEmpty(rawValue: accounts.filter(\.isOlympiaAccount).asIdentifiable())
+	public var olympiaAccounts: IdentifiedArrayOf<Profile.Network.Account> {
```
kugel3, line 69: Do we ever query this by id? Even if we do, O(n) vs O(1) shouldn't matter very much since you are unlikely to have tens of thousands of accounts. Maybe there are other reasons, like enforcing uniqueness?

kugel3 (approved):
Good, this removes some potential confusion

PR #960, opened 2023-12-13
Documentation only

CyonAlexRDX:
Documentation of OARS

### DIFF
--- a/RadixWallet/Clients/ProfileStore/ProfileStore.swift
+++ b/RadixWallet/Clients/ProfileStore/ProfileStore.swift
@@ -19,6 +19,7 @@
 ///		func values() -> AnyAsyncSequence<Profile>
 /// 	func unlockedApp() async -> Profile
 ///	 	func finishedOnboarding() async
+///     func finishOnboarding(with _: AccountsRecoveredFromScanningUsingMnemonic) async throws
 ///		func importCloudProfileSnapshot(_ h: ProfileSnapshot.Header) throws
 ///	 	func importProfileSnapshot(_ s: ProfileSnapshot) throws
 ///	 	func deleteProfile(keepInICloudIfPresent: Bool) throws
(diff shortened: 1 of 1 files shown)


RadixWallet/Clients/ProfileStore/ProfileStore.swift
```diff
@@ -19,6 +19,7 @@
 ///		func values() -> AnyAsyncSequence<Profile>
 /// 	func unlockedApp() async -> Profile
 ///	 	func finishedOnboarding() async
+///     func finishOnboarding(with _: AccountsRecoveredFromScanningUsingMnemonic) async throws {
```
GhenadieVP, line 22: trailing `{` should not be needed, no?

RadixWallet/Clients/ProfileStore/ProfileStore.swift
```diff
@@ -32,6 +33,11 @@
 /// async read `device.name` and `device.model` and update the Profile's header's `creatingDevice`
 /// and `lastUsedOnDevice` to use these values.
 ///
+/// If a user creates a Profile using Account Recovery Scan, app is suppose to call `finishOnboarding:with`,
```
GhenadieVP, line 36: I can't believe I am the one to spot spelling mistakes :), but should be `supposeD` instead of `suppose`? And probably should add an THE in `next time user starts THE app`?

RadixWallet/Clients/ProfileStore/ProfileStore.swift
```diff
@@ -205,14 +211,27 @@ extension ProfileStore {
 		var bdfs = accountsRecoveredFromScanningUsingMnemonic.deviceFactorSource
 		bdfs.hint.name = name
 		bdfs.hint.model = .init(model)
+
 		let accounts = accountsRecoveredFromScanningUsingMnemonic.accounts
-		let network = Profile.Network(networkID: .mainnet, accounts: accounts, personas: [], authorizedDapps: [])
+
+		// It is important that we always create the mainnet `Profile.Network` and
+		// add it, even if `accounts` is empty, since during App launch we check
```
GhenadieVP, line 218: feels like something is missing after `since during App launch we check`, or the combination with the next line does not sound right.

RadixWallet/Clients/ProfileStore/ProfileStore.swift
```diff
@@ -205,14 +211,27 @@ extension ProfileStore {
...
+		// `profile.networks.isEmpty` is used to determine if we should onboard
+		// user or not, thus, this ensure that we do not onboard a user who has
```
GhenadieVP, line 220: `ensureS`?

GhenadieVP (approved):

PR #961, opened 2023-12-13
[ABW-2651] Enable MARS for non-Mainnet (current network)

CyonAlexRDX:
[Jira Ticket](https://radixdlt.atlassian.net/browse/ABW-2651)

We accidentally did not read the current network... now MARS works on non mainnet (current network). 

Another important fix: if GW responds with error, we dismiss the flow (as intended)

https://github.com/radixdlt/babylon-wallet-ios/assets/116169792/a1668248-324b-4748-b893-d34f6a8eaee3

### DIFF
--- a/RadixWallet/Features/AccountRecoveryScan/Children/AccountRecoveryScanInProgress/AccountRecoveryScanInProgress.swift
+++ b/RadixWallet/Features/AccountRecoveryScan/Children/AccountRecoveryScanInProgress/AccountRecoveryScanInProgress.swift
@@ -69,7 +69,7 @@ public struct AccountRecoveryScanInProgress: Sendable, FeatureReducer {
 			active: IdentifiedArrayOf<Profile.Network.Account>,
 			inactive: IdentifiedArrayOf<Profile.Network.Account>
 		)
-		case failedToDerivePublicKey
+		case failed
 	}
 
 	// MARK: - Destination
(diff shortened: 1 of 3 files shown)


RadixWallet/Features/AccountRecoveryScan/Children/AccountRecoveryScanInProgress/AccountRecoveryScanInProgress.swift
```diff
@@ -148,14 +154,14 @@ public struct AccountRecoveryScanInProgress: Sendable, FeatureReducer {
...
 				state.mode = .factorSourceWithID(id: id, .loading)
-				return .run { [networkID = state.networkID, forOlympiaAccounts = state.forOlympiaAccounts] send in
+				return .run { [forOlympiaAccounts = state.forOlympiaAccounts] send in
 					let result = await TaskResult<IndicesUsedByFactorSource> {
 						try await factorSourcesClient.indicesOfEntitiesControlledByFactorSource(
 							.init(
 								entityKind: .account,
 								factorSourceID: id.embed(),
 								derivationPathScheme: forOlympiaAccounts ? .bip44Olympia : .cap26,
-								networkID: networkID
```
CyonAlexRDX, line 158: this was the fix, we were suppose to use `nil`... forgot that :D

RadixWallet/Features/AccountRecoveryScan/Children/AccountRecoveryScanInProgress/AccountRecoveryScanInProgress.swift
```diff
@@ -320,6 +327,9 @@ extension AccountRecoveryScanInProgress {
 					)
 				)
 			)
+		} catch: { error, send in
+			loggerGlobal.error("Failed to scan network, error: \(error)")
+			await send(.delegate(.failed))
```
CyonAlexRDX, line 332: another important fix: this ensure that we are not stuck in a loading state, but rather we dismiss the reducer...

RadixWallet/Features/AccountRecoveryScan/Children/AccountRecoveryScanInProgress/AccountRecoveryScanInProgress.swift
```diff
@@ -69,7 +69,7 @@ public struct AccountRecoveryScanInProgress: Sendable, FeatureReducer {
 			active: IdentifiedArrayOf<Profile.Network.Account>,
 			inactive: IdentifiedArrayOf<Profile.Network.Account>
 		)
-		case failedToDerivePublicKey
+		case failed
```
kugel3, line 72: Why this change?

kugel3 (approved):

RadixWallet/Features/AccountRecoveryScan/Children/AccountRecoveryScanInProgress/AccountRecoveryScanInProgress.swift
Sajjon: Because it is not the only way to fail, we have **four** different failures:
* failed to calculate derivation paths
* failed to derive public keys
* failed to create account
* failed to scan network (either bad response or failed to parse JSON) using GW endpoint

But currently we do not act on them... so we can use one shared error for all cases, and in the future if we need to disambiguate we will do so. 

Agreed?

kugel3: Got it

PR #962, opened 2023-12-13
Show name of the network instead of its id

GhenadieVP:
Simply show the network name, if available, instead of its id.


CyonAlexRDX (approved):
LGTM

PR #963, opened 2023-12-13
[ABW-1840] Detect Jailbreak

kugel3:
Jira ticket: [ABW-1840](https://radixdlt.atlassian.net/browse/ABW-1841)

## Description
Check if device is jailbroken and show dialog. 

### Notes
The dialog is shown every time the app launches after being purged from memory, which might be quite often.

This uses IOSSecuritySuite, which is a pretty powerful library, with 2000 stars on github. There are lightweight alternatives that have essentially only a single file, but they only have a handful of stars, and are not updated very frequently.

## Screenshot
image


RadixWallet/Clients/IOSSecurityClient/IOSSecurityClient+Live.swift
```diff
@@ -0,0 +1,9 @@
+import IOSSecuritySuite
+
+extension IOSSecurityClient: DependencyKey {
+	public static let liveValue = Self(
+		isJailbroken: {
+			IOSSecuritySuite.amIJailbroken() || true
```
Sajjon, line 6: do we really wanna default to `true` ? feels a bit to aggresive?

RadixWallet/Features/HomeFeature/Coordinator/Home.swift
```diff
@@ -92,6 +98,18 @@ public struct Home: Sendable, FeatureReducer {
 
 	public func reduce(into state: inout State, viewAction: ViewAction) -> Effect<Action> {
 		switch viewAction {
+		case .onFirstAppear:
+			if iOSSecurityClient.isJailbroken() {
+				state.destination = .acknowledgeJailbreakAlert(.init(
+					title: .init("Possible jailbreak detected"), // FIXME: Strings - get real string
+					message: .init("It appears that your device might be jailbroken. To ensure the security of your Accounts and assets, using the Radix Wallet on jailbroken devices is not recommended. Please confirm if you wish to continue anyway at your own risk."), // FIXME: Strings - splash_rootDetection_title - should be messageIOS
+					buttons: [
+						.cancel(.init("I Understand the Risk")),
```
Sajjon, line 107: FIXME: Strings on the button title as well?

kugel3: yes

RadixWallet/Clients/IOSSecurityClient/IOSSecurityClient+Live.swift
kugel3: This is even more aggressive, it's an OR! Forgot to remove it, thanks for spotting it. Btw I first put ?? there, confusing them the other way around.

CyonAlexRDX: Yeah haha I missed that it was OR 😅😅😅

CyonAlexRDX (approved):
LGTM

PR #967, opened 2023-12-15
[ABS-2662] Enter Seed Phrase Problems

kugel3:
Jira ticket: [ABS-2662](https://radixdlt.atlassian.net/browse/ABW-2662)

## Description
This fixes some issues with seed phrase entry. It also upgrades the debug buttons with a new power. If you press the small "M", it will only paste in the mnemonic, not press continue. That makes it possible to test "almost correct" mnemonics etc.

The second issue mentioned in the ticket is partly due to using the automatic debug entry, which is not a realistic scenario, it will not typically be possible to press Continue with an invalid mnemonic.

## Video

https://github.com/radixdlt/babylon-wallet-ios/assets/123396602/7f896e5c-1d50-40ea-9162-d6f7cd391cea

## PR submission checklist


CyonAlexRDX (approved):
LGTM

PR #978, opened 2023-12-21
1.7.0 RET upgrade

GhenadieVP:
Upgrade RET to latest version to bring in new transaction types.


RadixWallet/EngineKit/ReviewedTransaction.swift
```diff
@@ -18,173 +18,75 @@ public enum TransactionKind: Hashable, Sendable {
...
+					accountProofs: presentedProofs,
 					accountWithdraws: accountWithdraws,
 					accountDeposits: accountDeposits,
-					addressesInManifest: addressesInManifest,
-					metadataOfNewlyCreatedEntities: metadataOfNewlyCreatedEntities,
-					dataOfNewlyMintedNonFungibles: dataOfNewlyMintedNonFungibles,
-					addressesOfNewlyCreatedEntities: addressesOfNewlyCreatedEntities
+					addressesInManifest: encounteredEntities,
+					metadataOfNewlyCreatedEntities: newEntities.metadata,
+					dataOfNewlyMintedNonFungibles: [:],
```
GhenadieVP, line 72: this is no more provided by RET, most probably it was overlooked.

GhenadieVP:
Most of the files are format related, with the latest swiftformat version

RadixWallet/EngineKit/ReviewedTransaction.swift
kugel3: Is it `dataOfNewlyMintedNonFungibles` that is no longer provided? Or row 172?

GhenadieVP: `dataOfNewlyMintedNonFungibles` is no longer provided, and there is work in progress on RET side to add it back.

RadixWallet/Features/TransactionReviewFeature/TransactionReview.swift
```diff
@@ -868,31 +869,6 @@ extension TransactionReview {
 
 		typealias NonFungibleToken = OnLedgerEntity.NonFungibleToken
 
-		func tokenInfo(_ ids: [NonFungibleLocalId], for resourceAddress: ResourceAddress) async throws -> [NonFungibleToken] {
-			if let tokenData = dataOfNewlyMintedNonFungibles[resourceAddress.address] {
-				try extractTokenInfo(tokenData, for: resourceAddress)
```
kugel3, line 873: Why is all this going away? Because of the missing data from RET? If so, shouldn't we just comment it out until we get the data again?

kugel3 (approved):

RadixWallet/Features/TransactionReviewFeature/TransactionReview.swift
GhenadieVP: this will be revisited with the next version, as it will change.

PR #983, opened 2024-01-03
TX Review - Update using dApps section

GhenadieVP:
# What has been done

- Update UsingDapps section to display dApp cards.
- Added the UnknownDappComponents sheet to show the list of unknown components.
- Added "Unauthorized" label on dApps that are not authorized.
- Update the TransferLine to be shown only when there is at least one withdraw and one deposit.
- Added icons to section titles.

# Demo

https://github.com/radixdlt/babylon-wallet-ios/assets/118184705/f2b78a4b-7925-4d41-b2d6-977d36e17a5c


RadixWallet/Core/Resources/Resources/en.lproj/Localizable.strings
```diff
@@ -963,4 +965,4 @@ You will be asked to enter your main seed phrase. This is a set of **24 words**
 "recoverWalletWithoutProfile_complete_headerSubtitle" = "Accounts discovered in the scan have been added to your wallet.
 
 If you have any “Legacy” Accounts (created on the Olympia network) to import - or any Accounts using a Ledger hardware wallet device - please continue and then use the **Account Recovery Scan** option in your Radix Wallet settings under **Account Security**.";
-"recoverWalletWithoutProfile_complete_continueButton" = "Continue";
```
kugel3, line 966: These temporary strings should have a bunch of newlines above them to make sure they are not accidentally added to main. Ideally even a comment line above the first one.

RadixWallet/Features/TransactionReviewFeature/TransactionReview+View.swift
```diff
@@ -312,6 +300,19 @@ private extension View {
 		)
 	}
 
+	private func unknownComponents(with destinationStore: PresentationStoreOf<TransactionReview.Destination>) -> some View {
+		sheet(
+			store: destinationStore,
+			state: /TransactionReview.Destination.State.unknownDappComponents,
+			action: TransactionReview.Destination.Action.unknownDappComponents,
+			content: { store in
+				NavigationStack {
```
kugel3, line 309: You can use the .inNavigationStack modifier, then you don't need `store in`. Also, shouldn't it have a medium detent?

GhenadieVP:
> Looks like the dotted line is gone, but I can add that back later

Dotted line is shown only when there is withdraw and deposit, the presence of message is not taken into account anymore, as per new design.

PR #991, opened 2024-01-18
TX review for pool unit transfer

GhenadieVP:
Show the Pool Unit worth when transferring it between accounts instead of "just" showing the Pool units amount.

## What was done
- Introduced the ability to identify the fungible resource as pool unit in TransactionReview.
- Populate the information about the pool unit depending on the presence of Pool Interaction:
   - If pool interactions are present(in the case of pool contribution and pool redemption), make use of the information extract after manifest was analyzed.
   - If there is no pool interaction(as in the case of transferring the pool unit), resolve the pool unit details by using the OnLedgerEntitiesClient.
- Added a single place/call to load the details for all of the resources of interest, instead of loading those one by one.

### Simple Pool Unit transfer

https://github.com/radixdlt/babylon-wallet-ios/assets/118184705/ee666165-eb0b-48b0-991f-79a32554b31d

### Complex Transfer

https://github.com/radixdlt/babylon-wallet-ios/assets/118184705/c0ae1253-6328-45f5-b8b9-6a96bee16859


RadixWallet/Clients/OnLedgerEntitiesClient/OnLedgerEntitiesClient+Interface.swift
```diff
@@ -389,3 +389,92 @@ extension OnLedgerEntitiesClient {
...
+			[poolAddress],
+			.resourceMetadataKeys,
+			resource.atLedgerState,
+			.useCache
+		)
+		.map(\.resourcePool)
+		.first
+
+		guard let pool else {
+			return false // didn't load any pool
```
kugel3, line 414: What is the difference between the throwing situation and the nil situation?

RadixWallet/Clients/OnLedgerEntitiesClient/OnLedgerEntitiesClient+Interface.swift
```diff
@@ -389,3 +389,92 @@ extension OnLedgerEntitiesClient {
...
+		}
+
+		guard pool?.poolUnitResourceAddress == resource.resourceAddress else {
+			return false // The resource pool declared a different pool unit resource address
```
kugel3, line 418: This indicates an error, right? Perhaps we should throw?

RadixWallet/Clients/OnLedgerEntitiesClient/OnLedgerEntitiesClient+Interface.swift
```diff
@@ -389,3 +389,92 @@ extension OnLedgerEntitiesClient {
...
+		}
+
+		return true // It is a pool unit resource address
+	}
+
+	public func getPoolUnitDetails(_ poolUnitResource: OnLedgerEntity.Resource, forAmount amount: RETDecimal) async throws -> OwnedResourcePoolDetails? {
```
kugel3, line 424: Should this not be combined with `isPoolUnitResource`? It seems we will be loading the some of the same things just to figure out if it is a pool unit resource, we might as well go ahead and load the details? After making the sync checks locally of course. Or maybe there are situations where we only want the true/false answer?

kugel3: Btw, could you take the opportunity to rename `OwnedResourcePoolDetails`?

RadixWallet/Features/TransactionReviewFeature/TransactionReview+View.swift
```diff
@@ -34,7 +34,7 @@ extension TransactionReview.State {
 			showApprovalSlider: reviewedTransaction != nil,
 			canApproveTX: canApproveTX && reviewedTransaction?.feePayingValidation.wrappedValue == .valid,
 			sliderResetDate: sliderResetDate,
-			canToggleViewMode: reviewedTransaction != nil && reviewedTransaction?.isNonConforming != false,
+			canToggleViewMode: reviewedTransaction != nil && reviewedTransaction?.isNonConforming == false,
```
kugel3, line 37: I already changed this in my PR

RadixWallet/Features/TransactionReviewFeature/TransactionReview.swift
```diff
@@ -789,7 +789,7 @@ extension TransactionReview {
 					public let symbol: String?
 					public let address: ResourceAddress
 					public let icon: URL?
-					public var amount: RETDecimal
+					public var amount: String
```
kugel3, line 792: Why? This is not a ViewState, plus the Fungible one has a decimal amount

kugel3 (approved):
I might have preferred to actually do the fetching separately for extractWithdraws and extractDeposits but that' minor. This is a great improvement, let's get it in.

RadixWallet/Features/TransactionReviewFeature/TransactionReview.swift
GhenadieVP: This is mostly about reusing the `poolRedemptionValue` which a returns a string, it will return the special string if we could not calculate the redemption value.

PR #998, opened 2024-01-25
[ABW-2256] Transaction Review Staking and Unstaking

kugel3:
Jira ticket: ABW-2256

## Description
Adds support for staking and unstaking manifests in transaction review

## How to test
Stake/Unstake using https://dashboard.radixdlt.com/network-staking and make sure that the transaction review looks as in Zeplin.

## Screenshot
image

## Video

https://github.com/radixdlt/babylon-wallet-ios/assets/123396602/d55ec2cd-c864-4436-8d84-5eda3b39d900

https://github.com/radixdlt/babylon-wallet-ios/assets/123396602/9e0cd6fc-f99a-4d50-bbfa-93fd09f4ac70

## PR submission checklist

### DIFF
--- a/RadixWallet/Clients/OnLedgerEntitiesClient/OnLedgerEntitiesClient+Interface.swift
+++ b/RadixWallet/Clients/OnLedgerEntitiesClient/OnLedgerEntitiesClient+Interface.swift
@@ -391,6 +391,32 @@ extension OnLedgerEntitiesClient {
 }
 
 extension OnLedgerEntitiesClient {
+	/// Returns the validator of a correctly linked LSU, and `nil` for any other resource
+	public func isLiquidStakeUnit(_ resource: OnLedgerEntity.Resource) async -> OnLedgerEntity.Validator? {
+		guard let validatorAddress = resource.metadata.validator?.asGeneral else {
+			return nil
+		}
+
+		// Fetch validator info
+		let validator = try? await getEntity(
+			validatorAddress,
+			metadataKeys: .resourceMetadataKeys,
+			cachingStrategy: .useCache,
+			atLedgerState: resource.atLedgerState
+		)
+		.validator
+
+		guard let validator else {
+			return nil
+		}
+
+		guard validator.stakeUnitResourceAddress == resource.resourceAddress else {
+			return nil
+		}
+
+		return validator
+	}
+
 	public func isPoolUnitResource(_ resource: OnLedgerEntity.Resource) async -> Bool {
 		guard let poolAddress = resource.metadata.poolUnit?.asGeneral else {
 			return false // no declared pool unit
--- a/RadixWallet/Features/TransactionReviewFeature/TransactionReview+Sections.swift
+++ b/RadixWallet/Features/TransactionReviewFeature/TransactionReview+Sections.swift
@@ -14,6 +14,9 @@ extension TransactionReview {
 		var contributingToPools: TransactionReviewPools.State? = nil
 		var redeemingFromPools: TransactionReviewPools.State? = nil
 
+		var stakingToValidators: ValidatorsState? = nil
+		var unstakingFromValidators: ValidatorsState? = nil
+
 		var accountDepositSetting: DepositSettingState? = nil
 		var accountDepositExceptions: DepositExceptionsState? = nil
 
@@ -54,7 +57,7 @@ extension TransactionReview {
 			return nil
 		case .general, .transfer:
 			let resourcesInfo = try await resourcesInfo(allAddresses.elements)
-			let withdrawals = try? await extractWithdrawals(
+			let withdrawals = try await extractWithdrawals(
 				accountWithdraws: summary.accountWithdraws,
 				newlyCreatedNonFungibles: summary.newlyCreatedNonFungibles,
 				entities: resourcesInfo,
@@ -65,7 +68,7 @@ extension TransactionReview {
 			let dAppAddresses = summary.encounteredEntities.filter { $0.entityType() == .globalGenericComponent }
 			let dAppsUsed: TransactionReviewDappsUsed.State? = try await extractDapps(dAppAddresses, unknownTitle: L10n.TransactionReview.unknownComponents)
 
-			let deposits = try? await extractDeposits(
+			let deposits = try await extractDeposits(
 				accountDeposits: summary.accountDeposits,
 				newlyCreatedNonFungibles: summary.newlyCreatedNonFungibles,
 				entities: resourcesInfo,
@@ -73,7 +76,7 @@ extension TransactionReview {
 				networkID: networkID
 			)
 
-			let proofs = try? await exctractProofs(summary.presentedProofs)
+			let proofs = try await exctractProofs(summary.presentedProofs)
 
 			return Sections(
 				withdrawals: withdrawals,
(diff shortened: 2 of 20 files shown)


RadixWallet/Clients/OnLedgerEntitiesClient/OnLedgerEntitiesClient+Interface.swift
```diff
@@ -391,6 +391,32 @@ extension OnLedgerEntitiesClient {
...
+	public func isLiquidStakeUnit(_ resource: OnLedgerEntity.Resource) async -> OnLedgerEntity.Validator? {
+		guard let validatorAddress = resource.metadata.validator?.asGeneral else {
+			return nil
+		}
+
+		// Fetch validator info
+		let validator = try? await getEntity(
+			validatorAddress,
+			metadataKeys: .liquidStakeUnitMetadataKeys,
+			cachingStrategy: .useCache,
```
CyonAlexRDX, line 404: Hmm I think this should not `useCache` but the `neverUseCache` or what I called it, since it uses the metadata keys set `liquidStakeUnitMetadataKeys` which differs from the "standard one". So for this resource we have already fetched metadata with the standard key set, so the metadata values we need will not be in the cache.

And by using `useCache` we are overwriting the cache with non standard metadata values, thus purging it, from data we might need to populate Asset details for it?

I might be mistaken, but just wanted to check if you have thought about this and tested this?

kugel3: I presumed the cache key would include the keys, but I can look at it

RadixWallet/Features/TransactionReviewFeature/TrackedValidatorStake+Extensions.swift
```diff
@@ -0,0 +1,25 @@
+import Foundation
+
+// extension Collection<TrackedValidatorStake> {
```
CyonAlexRDX, line 3: remove dead code?

RadixWallet/Features/TransactionReviewFeature/TransactionReview+Sections.swift
```diff
@@ -156,10 +159,64 @@ extension TransactionReview {
 			)
 
 		case let .validatorStake(validatorAddresses: validatorAddresses, validatorStakes: validatorStakes):
```
CyonAlexRDX, line 161: those two big code hunks below are near identical? possible to share code?

RadixWallet/Features/TransactionReviewFeature/TransactionReview+Sections.swift
```diff
@@ -361,6 +422,29 @@ extension TransactionReview {
 		return .init(accounts: depositAccounts, enableCustomizeGuarantees: requiresGuarantees)
 	}
 
+	func extractValidators(for addresses: [EngineToolkit.Address]) async throws -> ValidatorsState? {
+		guard !addresses.isEmpty else { return nil }
+
+		let generalAddresses = try addresses.map { try $0.asSpecific() as Address }
```
CyonAlexRDX, line 428: `let GENERAL = ... $0.asSPECIFIC`  sound like an oxymoron?

RadixWallet/Features/TransactionReviewFeature/TransactionReview+View.swift
```diff
@@ -403,6 +432,24 @@ private extension View {
...
+	private func lsuDetails(with destinationStore: PresentationStoreOf<TransactionReview.Destination>) -> some View {
+		sheet(
+			store: destinationStore,
+			state: /TransactionReview.Destination.State.lsuDetails,
+			action: TransactionReview.Destination.Action.lsuDetails,
+			content: { LSUDetails.View(store: $0) }
+		)
+	}
+
+//	private func stakeClaimNFTDetails(with destinationStore: PresentationStoreOf<TransactionReview.Destination>) -> some View {
```
CyonAlexRDX, line 444: remove dead code?

RadixWallet/Features/TransactionReviewFeature/TrackedValidatorStake+Extensions.swift
kugel3: I removed the whole file, i guess it kept the file itself!

RadixWallet/Features/TransactionReviewFeature/TransactionReview.swift
```diff
@@ -596,6 +627,8 @@ extension TransactionReview {
 			let sections = try await sections(for: executionSummary, networkID: networkID)
 			await send(.internal(.updateSections(sections)))
 		} catch: { error, send in
+			print("•  Failed to extract transaction: \(error)")
```
CyonAlexRDX, line 630: loggerGloba.error ?

RadixWallet/Features/TransactionReviewFeature/TransactionReviewAccount/TransactionReviewAccount+View.swift
```diff
@@ -99,29 +99,43 @@ struct TransactionReviewResourceView: View {
...
-				viewState: details,
-				onTap: { stakeClaim in
-					onTap(stakeClaim.token)
-				},
-				onClaimAllTapped: {}
-			).padding()
+			StakeClaimNFTSView(viewState: details) { stakeClaim in
+				onTap(stakeClaim.token)
+			} onClaimAllTapped: {
+				// No action
```
CyonAlexRDX, line 122: make closure `onClaimAllTapped` optional??

RadixWallet/Features/TransactionReviewFeature/TransactionReview+Sections.swift
kugel3: It *is* correct as written, but I should use `.asGeneral` instead, I changed this around a few times and missed it

RadixWallet/Features/TransactionReviewFeature/TransactionReview.swift
kugel3: We already have that, forgot to remove this debug code

CyonAlexRDX (approved):
Oh wow, I have missed that we got this done before this release, bravo, looks great! My most pressing comment is that of the usage of cache.

RadixWallet/Features/TransactionReviewFeature/TransactionReview+Sections.swift
kugel3: Actually we didn't even have an option to convert directly from EngineToolkit.Address to the general address type, but I have added that now

RadixWallet/Features/TransactionReviewFeature/TransactionReview+Sections.swift
kugel3: Not really, without restructuring the whole switch, because they have different associated types in the cases, they initialise the result in different ways, and both deposits and withdrawals are different. Generally speaking we have restructured it to avoid code duplication quite well, but most of the arguments to the extraction methods tend to be the same. Haven't found a good way to make use of that fact!

RadixWallet/Clients/OnLedgerEntitiesClient/OnLedgerEntitiesClient+Interface.swift
GhenadieVP: Just be sure to use the same metadata keys as when fetching the validator details in Account details.

RadixWallet/Features/TransactionReviewFeature/TransactionReview+Sections.swift
```diff
@@ -679,20 +826,34 @@ extension TransactionReview {
...
+				}
 
-		let stakeClaimTokens = try tokens.map { token in
-			guard let data = token.data, let claimAmount = data.claimAmount else {
-				throw InvalidStakeClaimToken()
+				return OnLedgerEntitiesClient.StakeClaim(
+					validatorAddress: stakeClaimValidator.address,
+					token: token,
+					claimAmount: data.claimAmount,
+					reamainingEpochsUntilClaim: Int(data.claimEpoch) - Int(resource.atLedgerState.epoch)
```
GhenadieVP, line 842: for the case of stake claim resource as part of unstake transaction, we can/should set `reamainingEpochsUntilClaim: nil`.

RadixWallet/Features/TransactionReviewFeature/TransactionReview+View.swift
```diff
@@ -648,6 +709,79 @@ extension StrokeStyle {
 	static let transactionReview = StrokeStyle(lineWidth: 2, dash: [5, 5])
 }
 
+extension TransactionReview {
+	public typealias ValidatorsState = ValidatorsView.ViewState
+	public typealias ValidatorState = ValidatorView.ViewState
+
+	public struct ValidatorsView: SwiftUI.View {
```
GhenadieVP, line 716: Should this be here? can it be moved to a separate file?

GhenadieVP (approved):
Awesome work 💯!

RadixWallet/Features/TransactionReviewFeature/TransactionReview+View.swift
kugel3: It definitely could, but it's only used here I think, and lots of other subviews are in this file.

RadixWallet/Clients/OnLedgerEntitiesClient/OnLedgerEntitiesClient+Interface.swift
Sajjon: I dont think the cache key contain metadatakeys no, I think the cache key is the validatorAddress?

PR #1001, opened 2024-01-25
Preselect Single Value

kugel3:
## Description
In any selection list where there is a single element, and the user is asked to select one, or at least one element, then it should be preselected.

The last commit is entirely optional. It simplifies SelectionRequirement. I am not 100% I prefer the new version, but it _is_ more standard.

## Screenshot
image

## PR submission checklist


GhenadieVP (approved):

kugel3:
This needs to be extended to also handle some more complex situations

PR #1003, opened 2024-01-26
Formatted Instant NFT Data

GhenadieVP:
Detect and display properly the Instant NFT Data type.

A given NFT Data field is consider an Instant type if:
- The kind is an I64.
- The type_name is Instant.
- The value is a valid i64.

Formatted the date with system's `format` to inherit all of the users settings regarding date formatting, there are no requirements to display the `Instant` NFT Data in a particular, fixed format.

IMG_0097 2

### DIFF
--- a/RadixWallet/Features/AssetsFeature/Components/NonFungibleAssetList/Components/Details/NonFungibleTokenDetails+View.swift
+++ b/RadixWallet/Features/AssetsFeature/Components/NonFungibleAssetList/Components/Details/NonFungibleTokenDetails+View.swift
@@ -177,6 +177,9 @@ extension NonFungibleTokenDetails {
 				} else {
 					EmptyView()
 				}
+
+			case let .instant(date):
+				KeyValueView(key: field.name, value: date.formatted())
 			}
 		}
 	}
@@ -211,6 +214,7 @@ extension NonFungibleTokenDetails.ViewState.TokenDetails {
 			case decimal(RETDecimal)
 			case `enum`(variant: String)
 			case id(NonFungibleLocalId)
+			case instant(Date)
 		}
 
 		public let kind: Kind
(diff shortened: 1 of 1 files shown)


RadixWallet/Features/AssetsFeature/Components/NonFungibleAssetList/Components/Details/NonFungibleTokenDetails+View.swift
```diff
@@ -279,9 +283,23 @@ private extension String {
...
+			} else {
+				.primitive(self)
+			}
+		}
+	}
 }
 
 private extension GatewayAPI.ProgrammaticScryptoSborValue {
+	enum TypeName: String {
+		case instant = "Instant"
```
kugel3, line 300: So "Instant" is the only typename we support? For non-instants, is `typeName` simply undefined?

kugel3 (approved):

RadixWallet/Features/AssetsFeature/Components/NonFungibleAssetList/Components/Details/NonFungibleTokenDetails+View.swift
```diff
@@ -279,9 +283,23 @@ private extension String {
 			}
 		}
 	}
+
+	var asInstantDataField: ArbitraryDataFieldKind? {
+		nilIfEmpty.map {
+			if let timeInterval = Int64($0) {
```
Sajjon, line 289: Hmm this is an extension on `String`, so the i64 is sent as a String? That sounds strange and wrong?

GhenadieVP: yes, it is returned as String as per GW contract. i64 refers to the [scrypto sbor value kind](https://github.com/radixdlt/babylon-wallet-ios/blob/216b24d52913544c4d471f656224f301112230ef/RadixWallet/Clients/GatewayAPI/CodeGen/Input/gateway-api-schema.yml#L3454), and we then validate the payload as actually being the Int64

RadixWallet/Features/AssetsFeature/Components/NonFungibleAssetList/Components/Details/NonFungibleTokenDetails+View.swift
GhenadieVP: `Instant` it is the only one we do identify and act upon currently. For non-instants we do ignore the typeName, even if it is defined.

CyonAlexRDX (approved):
LGTM

CyonAlexRDX (approved):
LGTM

PR #1009, opened 2024-01-29
[ABW-1913] Apply divisibility to guaranteed amount

kugel3:
Jira ticket: ABW-1913

## Description
After applying a guarantee, we should round the amount according to the divisibility of the resource.

## How to test
- Make an "estimated" transaction with a resource that has divisibility smaller than 18.
- Apply a guarantee
-> The resulting amount should not have more decimals than the divisibility dictates

## Screenshot
Coming

## PR submission checklist


kugel3:
This functionality is partly already existing, partly included in https://github.com/radixdlt/babylon-wallet-ios/pull/1010

PR #1012, opened 2024-01-30
Stake claime transaction aggregate amounts

GhenadieVP:
Aggregate the transfer amounts for the same resource for the same account.

The RET analysis might return multiple withdrawls/deposits for the same resource, instead of showing separate entries for each withdral/deposit, we aggregate the fungible amounts or the non fungible ids.

This aggregates only the guranteed amounts,  predicted amounts cannot be aggregated since each amount will have a specific instruction index attached.

This is only used curently for stake claim transactions, when the Dapp might sent a manifest which is interpreted by RET as having multiple distinct withdrawls/deposits for the same resource.

This should eventually be moved to RET, so it should return the aggregated amounts

## Before

https://github.com/radixdlt/babylon-wallet-ios/assets/118184705/9ff86e0a-53e9-4067-acec-75dd513dce33

## After

https://github.com/radixdlt/babylon-wallet-ios/assets/118184705/d8eeeded-2f6c-4f02-b6e1-8c4c3cdd4648

### DIFF
--- a/RadixWallet/Features/TransactionReviewFeature/TransactionReview+Sections.swift
+++ b/RadixWallet/Features/TransactionReviewFeature/TransactionReview+Sections.swift
@@ -975,12 +975,11 @@ extension [String: [ResourceIndicator]] {
 extension ResourceIndicator {
 	var isGuaranteedAmount: Bool {
 		switch self {
-		case .fungible(_, .guaranteed):
-			true
-		case .nonFungible(_, .byIds):
-			true
+		case .fungible(_, .guaranteed), .nonFungible(_, .byIds):
+			return true
 		default:
-			false // Cannot sum up the predicted amounts, as each predicted amount has a specific instruction index
+			assertionFailure("Cannot sum up the predicted amounts")
+			return false // Cannot sum up the predicted amounts, as each predicted amount has a specific instruction index
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


RadixWallet/Features/TransactionReviewFeature/TransactionReview+Sections.swift
```diff
@@ -934,3 +934,93 @@ extension TransactionReview.ResourceInfo {
 		}
 	}
 }
+
+extension [String: [ResourceIndicator]] {
```
CyonAlexRDX, line 938: nit: I prefer old style: `extension Dictionary where Key == String, Value == ResourceIndicator {` for two reasons:
1. we can search "extension Dictionary" to find ALL dictionary extensions, which is not possible with this, since "extension [" also finds array extension
2. We can use `: Protocol` syntax instead of `==` which is powerful, and only supported by `extension Dictionary` syntax, and thus nice to be consistent.

RadixWallet/Features/TransactionReviewFeature/TransactionReview+Sections.swift
```diff
@@ -934,3 +934,93 @@ extension TransactionReview.ResourceInfo {
...
+}
+
+extension ResourceIndicator {
+	var isGuaranteedAmount: Bool {
+		switch self {
+		case .fungible(_, .guaranteed):
+			true
+		case .nonFungible(_, .byIds):
+			true
+		default:
```
CyonAlexRDX, line 982: is it not feasible to enumerate all cases? I tend to dislike `default` usage (which has bitten us before).

CyonAlexRDX (approved):
LGTM

RadixWallet/Features/TransactionReviewFeature/TransactionReview+Sections.swift
kugel3: I would agree. The trues and the falses can be aggregated so it still only uses two rows

RadixWallet/Features/TransactionReviewFeature/TransactionReview+Sections.swift
```diff
@@ -934,3 +934,93 @@ extension TransactionReview.ResourceInfo {
...
+
+	public mutating func add(_ other: Self) {
+		guard other.resourceAddress == resourceAddress else {
+			assertionFailure("The indicators should have the same resource address")
+			return
+		}
+		switch (self, other) {
+		case (.fungible(_, var fungibleIndicator), let .fungible(_, otherFungibleIndicator)):
+			fungibleIndicator.add(otherFungibleIndicator)
+			self = .fungible(resourceAddress: resourceAddress, indicator: fungibleIndicator)
```
kugel3, line 995: Since you are using the inner `add` in this way, you might as well make it functional and call it `adding`. Saves one row here, and makes it clearer.

kugel3 (approved):

PR #1013, opened 2024-01-30
Linked connector not restoring after iCloud restore

GhenadieVP:
Connect to the P2PLinks during the profile Restore from Backup flow.

# Changes
- Don't remove the P2PLinks from the Profile when deleting the Wallet, so those are not lost when trying to Restore from Backup.
- When user starts the Restore from Backup flow, the Wallet pro-actively connects to the P2PLinks from the selected Profile. 
- While it was possible to connect to P2PLinks only when landing on home screen, connecting pro-actively in Restore from Backup flow, would make the experience of needing to use a Ledger device during the flow a lot better.
- If user leaves the Restore from Backup flow, RadixConnect will remove the established connections.

# Demo

https://github.com/radixdlt/babylon-wallet-ios/assets/118184705/a4d52339-1fce-4195-b6a5-b9fee83f123f

### DIFF
--- a/RadixWallet/Features/ProfileBackupsFeature/RestoreProfileFromBackup/Coordinator/RestoreProfileFromBackupCoordinator.swift
+++ b/RadixWallet/Features/ProfileBackupsFeature/RestoreProfileFromBackup/Coordinator/RestoreProfileFromBackupCoordinator.swift
@@ -61,6 +61,7 @@ public struct RestoreProfileFromBackupCoordinator: Sendable, FeatureReducer {
 	@Dependency(\.factorSourcesClient) var factorSourcesClient
 	@Dependency(\.errorQueue) var errorQueue
 	@Dependency(\.continuousClock) var clock
+	@Dependency(\.radixConnectClient) var radixConnectClient
 	public init() {}
 
 	public var body: some ReducerOf<Self> {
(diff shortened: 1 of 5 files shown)


RadixWallet/Features/ProfileBackupsFeature/RestoreProfileFromBackup/Coordinator/RestoreProfileFromBackupCoordinator.swift
```diff
@@ -88,6 +89,7 @@ public struct RestoreProfileFromBackupCoordinator: Sendable, FeatureReducer {
 			state.profileSelection = .init(snapshot: profileSnapshot, isInCloud: isInCloud)
 			return .run { send in
 				try? await clock.sleep(for: .milliseconds(300))
+				try await radixConnectClient.connectToP2PLinks(profileSnapshot.appPreferences.p2pLinks)
```
Sajjon, line 92: cant remember now, but what are we infact awaiting here? looks wrong? This mean that `send(.internal(.delay..(.importMnemonicsFlow))` event is not emitted until `radixConnectClient.connectToP2PLinks` has finished? So what happens if you quit Chrome and redo your demo in the video? Will this `RestoreProfileFromBackupCoordinator` reducer wait forever until connected? Or are we using unstructured concurrency inside `connectToP2PLinks` ensuring it will not be 'blocking"? Otherwise, perhaps this `try await` inside this `run` must change, e.g. to a `run {}.concat {}` or whatever the syntax is block, or possible even `Task` (unstructured concurrency)?

kugel3 (approved):

RadixWallet/Features/ProfileBackupsFeature/RestoreProfileFromBackup/Coordinator/RestoreProfileFromBackupCoordinator.swift
GhenadieVP: > Will this RestoreProfileFromBackupCoordinator reducer wait forever until connected?

No, `connectToP2PLinks` calls in turn the `try await rtcClients.connect(password, waitsForConnectionToBeEstablished: false)`.

CyonAlexRDX (approved):
LGTM

PR #1015, opened 2024-01-31
[ABW-2875] Transaction Review Asset Buttons

kugel3:
Jira ticket: ABW-2875

## Description
Fixes the appearance and behaviour of all the assets in Transaction Review, and most of the ones used in the account screens etc. Among other things, it turns all of these into actual buttons that highlight when pressed. The highlight is quite subtle though, it's the native one, which turns the view slightly transparent. For white views, this means that the background does not change at all.

We could easily change this behaviour by adding a new button style that slightly darkens the view when the button is pressed.

1. Step 1
2. Step 2
3. Verify

1. Step 1
2. Step 2
3. Verify

## Screenshot

| - | - |

## Video
paste link here

## PR submission checklist

### DIFF
--- a/RadixWallet/Core/DesignSystem/Components/PoolUnitView.swift
+++ b/RadixWallet/Core/DesignSystem/Components/PoolUnitView.swift
@@ -9,7 +9,7 @@ public struct PoolUnitView: View {
 	}
 
 	public let viewState: ViewState
-	public let backgroundColor: Color
+	public let background: Color
 	public let onTap: () -> Void
 
 	public var body: some View {
@@ -55,7 +55,7 @@ public struct PoolUnitView: View {
 				}
 				.padding(.bottom, .small2)
 
-				Text(L10n.TransactionReview.worth)
+				Text(L10n.TransactionReview.worth.uppercased())
 					.textStyle(.body2HighImportance)
 					.foregroundColor(.app.gray2)
 					.padding(.bottom, .small3)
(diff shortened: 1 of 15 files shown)


RadixWallet/Core/DesignSystem/Components/PoolUnitView.swift
```diff
@@ -46,51 +46,49 @@ public struct PoolUnitView: View {
 
 					Spacer(minLength: 0)
 
+					if let isSelected = viewState.isSelected {
+						CheckmarkView(appearance: .dark, isChecked: isSelected)
+					}
+
 					//					AssetIcon(.asset(AssetResource.info), size: .smallest)
```
CyonAlexRDX, line 53: time to remove dead code below?

kugel3: This is actually upcoming code, it will be activated once we have the copy for those info views

RadixWallet/Features/TransactionReviewFeature/TransactionReview+View.swift
```diff
@@ -694,39 +696,41 @@ struct TransactionReviewTokenView: View {
...
 
-				if let fiatAmount = viewState.fiatAmount {
-					// Text(fiatAmount.formatted(.currency(code: "USD")))
-					Text(fiatAmount.formatted())
-						.textStyle(.body2HighImportance)
-						.foregroundColor(.app.gray1)
-						.padding(.top, .small2)
-				}
+					if let fiatAmount = viewState.fiatAmount {
+						// Text(fiatAmount.formatted(.currency(code: "USD")))
```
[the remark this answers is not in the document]
kugel3: Same here, this was added long time ago when I thought we had a pricing service etc

GhenadieVP (approved):
Awesome 👍!

PR #1016, opened 2024-01-31
Stake claim details updated

GhenadieVP:
Update stake details representation of stake claim status, for now it has the same look and feel as in Stakes list. There is planned a another iteration on the design of this, but it will not go into this release.

# Changes
- Extracted the stake claim tokens list in a separate view so it can be reused.
- Made use of the reusable view in NFT details screen.

# Demo

https://github.com/radixdlt/babylon-wallet-ios/assets/118184705/d757478c-b07d-4055-9edf-6b4ced5e53ec

### DIFF
--- a/RadixWallet/Features/AssetsFeature/Components/StakeUnitList/Components/StakeClaimNFTsView.swift
+++ b/RadixWallet/Features/AssetsFeature/Components/StakeUnitList/Components/StakeClaimNFTsView.swift
@@ -164,9 +164,7 @@ public struct StakeClaimTokensView: View {
 			VStack(alignment: .leading, spacing: .small2) {
 				ForEach(claims) { claim in
 					Button {
-						if let onTap {
-							onTap(claim)
-						}
+						onTap?(claim)
 					} label: {
 						HStack {
 							TokenBalanceView(viewState: .xrd(balance: claim.claimAmount))
--- a/RadixWallet/Features/AssetsFeature/Components/StakeUnitList/Components/StakeClaimNFTsView.swift
+++ b/RadixWallet/Features/AssetsFeature/Components/StakeUnitList/Components/StakeClaimNFTsView.swift
@@ -1,7 +1,6 @@
 // MARK: - StakeClaimResourceView
 public struct StakeClaimResourceView: View {
 	public struct ViewState: Sendable, Hashable {
-		public let canClaimTokens: Bool
 		public let validatorName: String?
 		public var stakeClaimTokens: StakeClaimTokensView.ViewState
 		public let stakeClaimResource: OnLedgerEntity.Resource
@@ -16,7 +15,6 @@ public struct StakeClaimResourceView: View {
 			validatorName: String? = nil,
 			selectedStakeClaims: IdentifiedArrayOf<NonFungibleGlobalId>? = nil
 		) {
-			self.canClaimTokens = canClaimTokens
 			self.validatorName = validatorName
 			self.stakeClaimResource = stakeClaimTokens.resource
 			self.stakeClaimTokens = .init(
--- a/RadixWallet/Features/AssetsFeature/Components/StakeUnitList/Components/StakeClaimNFTsView.swift
+++ b/RadixWallet/Features/AssetsFeature/Components/StakeUnitList/Components/StakeClaimNFTsView.swift
@@ -4,6 +4,7 @@ public struct StakeClaimNFTSView: View {
 		public let canClaimTokens: Bool
 		public let validatorName: String?
 		public let stakeClaimTokens: OnLedgerEntitiesClient.NonFunbileResourceWithTokens
+
 		var selectedStakeClaims: IdentifiedArrayOf<NonFungibleGlobalId>?
 
 		var unstaking: IdentifiedArrayOf<OnLedgerEntitiesClient.StakeClaim> {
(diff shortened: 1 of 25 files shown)


RadixWallet/Features/AssetsFeature/Components/StakeUnitList/Components/StakeClaimNFTsView.swift
```diff
@@ -82,24 +65,80 @@ public struct StakeClaimNFTSView: View {
...
+public struct StakeClaimTokensView: View {
+	enum SectionKind {
+		case unstaking
+		case readyToBeClaimed
+		case toBeClaimed
+	}
+
+	public struct ViewState: Sendable, Hashable {
+		public let canClaimTokens: Bool
+		public let stakeClaims: IdentifiedArrayOf<OnLedgerEntitiesClient.StakeClaim>
```
kugel3, line 90: I presume that the purpose of the IdentifiedArrays here is to ensure uniqueness, but I don't think that is needed in the ViewState. It should be enough to do that in State.

RadixWallet/Features/AssetsFeature/Components/StakeUnitList/Components/StakeClaimNFTsView.swift
```diff
@@ -1,26 +1,12 @@
...
-		}
-
-		var readyToBeClaimed: IdentifiedArrayOf<OnLedgerEntitiesClient.StakeClaim> {
-			stakeClaimTokens.stakeClaims.filter(\.isReadyToBeClaimed)
-		}
-
-		var toBeClaimed: IdentifiedArrayOf<OnLedgerEntitiesClient.StakeClaim> {
-			stakeClaimTokens.stakeClaims.filter(\.isToBeClaimed)
-		}
+		public var stakeClaimTokens: StakeClaimTokensView.ViewState
```
kugel3, line 5: I suppose this should actually be a let?

kugel3 (approved):
Nice

RadixWallet/Features/AssetsFeature/Components/StakeUnitList/Components/StakeClaimNFTsView.swift
GhenadieVP: It is var because it is needed to update the selected stake claims array in selection mode.

kugel3: Ah, so we are reusing the ViewState type as State? That would also explain the identified arrays. But for such cases I wonder if we shouldn't do it the other way around, and declare it as a State, and reuse it as ViewState, rather than the other way around. We are not consistent about that, but I think it makes more sense to see the State as primary. Something to put on the todo list.

RadixWallet/Features/AssetsFeature/Components/StakeUnitList/Components/StakeClaimNFTsView.swift
```diff
@@ -82,24 +65,80 @@ public struct StakeClaimNFTSView: View {
...
+		var unstaking: IdentifiedArrayOf<OnLedgerEntitiesClient.StakeClaim> {
+			stakeClaims.filter(\.isUnstaking)
+		}
+
+		var readyToBeClaimed: IdentifiedArrayOf<OnLedgerEntitiesClient.StakeClaim> {
+			stakeClaims.filter(\.isReadyToBeClaimed)
+		}
+
+		var toBeClaimed: IdentifiedArrayOf<OnLedgerEntitiesClient.StakeClaim> {
+			stakeClaims.filter(\.isToBeClaimed)
```
Sajjon, line 102: the name `isToBeClaimed` is too similar to `isReadyToBeClaimed` IMO, thus is `readyToBeClaimed` IdentifiedArrayOf and `toBeClaimed` array variable names also too similar. does "isToBeClaimed" really mean "notYetReadyToBeClaimed"? Then I think we should name it that! Else, what is it? :D

RadixWallet/Features/AssetsFeature/Components/StakeUnitList/Components/StakeClaimNFTsView.swift
```diff
@@ -125,7 +164,9 @@ public struct StakeClaimNFTSView: View {
 			VStack(alignment: .leading, spacing: .small2) {
 				ForEach(claims) { claim in
 					Button {
-						onTap(claim)
+						if let onTap {
+							onTap(claim)
```
Sajjon, line 168: or `onTap?(claim)` ?

CyonAlexRDX (approved):
LGTM,just a bit confused by some variable names

RadixWallet/Features/AssetsFeature/Components/StakeUnitList/Components/StakeClaimNFTsView.swift
GhenadieVP: This is consistent with representation at the UI level, when two different strings are shown "Ready to be claimed" and "To be claimed", I did choose to be consistent in the code.
"To be claimed" means that user will be able to claim it in the future.
"Ready to be claimed" means exactly what it says.
"notYetReadyToBeClaimed" what you have suggested is defined by "isUnstaking"

PR #1022, opened 2024-02-01
Treat only "general" with transfers as "transfer"

kugel3:
## Description
Manifests classified as `.general` should be considered:
- as `.transfer` if it has withdrawals or deposits
- non-conforming if it doesn't

## PR submission checklist


GhenadieVP (approved):

CyonAlexRDX (approved):
LGTM

PR #1025, opened 2024-02-01
[ABW-2902] LSU Guarantee Issue and more

kugel3:
Jira ticket: [ABW-2902](https://radixdlt.atlassian.net/browse/ABW-2902)

## Description
Makes the icon in Customize Guarantees dynamic, so it uses a Fungible Token icon for fungible tokens, etc. 

Fixes the "Unknown dApp" issue. This was actually present also for resources in the DappDetails view.

Also completely overhauls all thumbnail types, replacing them with a single type.

Another issue was that the name/symbol of resource within a pool unit was not displayed in the transaction review context, it just appeared blank if the symbol was missing. Now we fall back first on the name, then on "Unknown", as elsewhere.

Finally it fixes the Pool unit header, that should display "Pool Units" as the fallback name.

## How to test
Go through all icons in the app and make sure they look correct, also when showing the fallback value.

## Notes
In the resources list in pool units, we (currently) don't have any information about what type of resource each row represents. This means that we can't fully use the new Thumbnail type, instead we treat all as standard fungible tokens. This in turn means that the placeholder image will not the the pool unit/lsu icon when contributing these resources, if the url is missing we are forced to fall back on the token placeholder.

It is quite confusing that the Liquid Stake Units row in the pool unit actually does show the pool unit icon, but this is actually set via url on that LSU. 

## Screenshot
Main issue:

Before:

After:

## PR submission checklist

[ABW-2902]: https://radixdlt.atlassian.net/browse/ABW-2902?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


RadixWallet/Clients/OnLedgerEntitiesClient/Helpers/OnLedgerEntitiesClient+CreateEntity.swift
```diff
@@ -541,6 +541,12 @@ extension OnLedgerEntitiesClient {
 	}
 }
 
+extension OnLedgerEntity.Resource {
+	var title: String {
+		metadata.name ?? metadata.symbol ?? L10n.TransactionReview.unknown
```
GhenadieVP, line 546: This is not quite consistent with the [rule for resources display](https://radixdlt.atlassian.net/wiki/spaces/AT/pages/2844753933/Rules+for+Display+of+Assets+in+Account+Detail+View#Token-Display), according to which we should have something line `metadata.symbol ?? metadata.name`, if the resource has neither account or name, no title is shown.

is it meant to be used across the app?

RadixWallet/Core/DesignSystem/Components/PoolUnitView.swift
```diff
@@ -16,24 +16,13 @@ public struct PoolUnitView: View {
...
+					Thumbnail(.poolUnit, url: viewState.poolIcon, size: .extraSmall)
+						.padding(.trailing, .medium3)
 
 					VStack(alignment: .leading, spacing: 0) {
-						if let poolName = viewState.poolName {
-							Text(poolName)
-								.textStyle(.body1Header)
-								.foregroundColor(.app.gray1)
-						}
+						Text(viewState.poolName ?? L10n.TransactionReview.poolUnits)
```
GhenadieVP, line 23: I am not entirely sure we should have the fallback name, but probably better than empty string.

RadixWallet/Features/TransactionReviewFeature/TransactionReviewGuarantees/TransactionReviewGuarantees+View.swift
```diff
@@ -92,9 +92,9 @@ extension TransactionReviewGuarantee.State {
 		.init(
 			id: id,
 			account: account,
-			token: .init(
-				name: resource.title,
-				thumbnail: isXRD ? .xrd : .known(resource.metadata.iconURL),
+			fungible: .init(
+				name: resource.metadata.name ?? resource.metadata.symbol ?? L10n.TransactionReview.unknown,
```
GhenadieVP, line 96: Could use the created `title`

GhenadieVP (approved):
Big and nice update 💯!
Lets double check the soundness of `title` computed property.

RadixWallet/Core/DesignSystem/Components/PoolUnitView.swift
kugel3: Zeplin has a it though

RadixWallet/Features/DappsAndPersonas/DappDetails/DappDetails+View.swift
```diff
@@ -147,12 +147,6 @@ private extension DappDetails.State {
 	}
 }
 
-extension OnLedgerEntity.Resource {
-	var title: String {
-		metadata.name ?? metadata.symbol ?? L10n.DAppRequest.Metadata.unknownName
-	}
-}
-
```
kugel3, line 155: This is where I took it from, but I guess it should be the other way around.

RadixWallet/Clients/OnLedgerEntitiesClient/Helpers/OnLedgerEntitiesClient+CreateEntity.swift
kugel3: Yes that was the idea I suppose, I just moved it here from another location.

RadixWallet/Core/DesignSystem/HitTargetSize.swift
```diff
@@ -9,6 +9,9 @@ public enum HitTargetSize: CGFloat {
 	/// 34
 	case smaller = 34
 
+	/// 40
+	case extraSmall = 40
```
Sajjon, line 13: :D

kugel3: Getting harder to make up names, this needs an overhaul soon

PR #1026, opened 2024-02-02
[ABW-2905] Show Guarantees for Pool Units

kugel3:
Jira ticket: [ABW-2905](https://radixdlt.atlassian.net/browse/ABW-2905)

## Description
This PR makes it so that guaranteed amounts for pool units are shown.

It also updates the guarantee amount view for fungible tokens to use the new design.

## Discussion
With this PR we will also start showing the amount of pool units that are owned, both in the portfolio and when selecting an asset for transfer. Do we want to suppress this?

Also, do we want to add the amount (including estimated/guaranteed amount) for LSUs?

Both of these changes are quite trivial.

## Screenshot
image

## PR submission checklist

[ABW-2905]: https://radixdlt.atlassian.net/browse/ABW-2905?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/RadixWallet/Core/DesignSystem/Components/PoolUnitView.swift
+++ b/RadixWallet/Core/DesignSystem/Components/PoolUnitView.swift
@@ -2,7 +2,7 @@
 public struct PoolUnitView: View {
 	public struct ViewState: Equatable {
 		public let poolName: String?
-		public let amount: RETDecimal
+		public let amount: RETDecimal?
 		public let guaranteedAmount: RETDecimal?
 		public let dAppName: Loadable<String?>
 		public let poolIcon: URL?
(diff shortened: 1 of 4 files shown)


RadixWallet/Core/DesignSystem/Components/PoolUnitView.swift
```diff
@@ -35,8 +37,11 @@ public struct PoolUnitView: View {
 
 					Spacer(minLength: 0)
 
+					TransactionReviewAmountView(amount: viewState.amount, guaranteedAmount: viewState.guaranteedAmount)
```
GhenadieVP, line 40: Better to not show the amount in Account details, because it will have potentially a cascading effect - why not do the same for LSU, have Android to update it also.
It is a good idea, but not to be done now.

GhenadieVP (approved):
LGTM!
IMO we should hide the pool units `amount`, for now.

PR #1032, opened 2024-02-09
Introduce Sargon: Radix Engine Toolkit (RET) removed

CyonAlexRDX:
# STATUS: Ready for review
Do not merge yet, *we must test it more*! Initial testing looks good, I've tried all various "standard transactions" which seems to be working:
* Single Fungible token transfer
* Non Fungible token transfer
* Multiple Fungible
* Multiple Non Fungible
* Stake
* Unstake
* Claim stake
* Use Gumball Club Dapp with Membership
* Mint Fungible
* Update Third Party Deposits
* Mint Non Fungible

(Creating MANY NonFungible and MANY Fungible - from Dev Preferences for DEBUG builds does not work, so I will look into that, but at least mint many NFTs is commented out currently in `main` anyway.)

# Changes
This PR removes [`RET: Radix Engine Toolkit`](https://github.com/radixdlt/swift-engine-toolkit) (which is Swifty wrapper around [Rust impl](https://github.com/radixdlt/radix-engine-toolkit)), and the extensions we had on those types which reside in the [EngineKit folder in this repo](https://github.com/radixdlt/babylon-wallet-ios/tree/main/RadixWallet/EngineKit) - they are all replaces with [`Sargon`](https://github.com/radixdlt/sargon), specifically the [Swift Package / `target` with the same name](https://github.com/radixdlt/sargon/blob/develop/Package.swift#L60), which is a non-trivial Swift wrapper around the UniFFI bindgen "hot garbage" generated file with 22,000 lines of Swift code.

`Sargon` also contains a full implementation of `Profile`, this PR does NOT make use of that here in iOS wallet land, instead we keep using the iOS wallets Profile impl for now, but at a later point in time (hopefully this month) we *will* remove iOS profile and use Sargon profile instead.

`RETDecimal` is now known as `Decimal192` (Not named `Decimal` to avoid naming collision with `Foundation.Decimal`) - and should be identical to what we have today (with a few extra improvements, such `asDecimal` is not a throwing method but a computed property. And more initializers)

## Sendable
All types in Sargon are value types - not references types like they were in RET - with mutable stored properties. All types are `Hashable` (thus `Equatable`) and `Sendable`.

## Samples Values
ALL types in Sargon (in Rust) have sample values. `Self.sample()` and `Self.sample_other()`, almost all types in Swift have got those a exported as static vars, `Self.sample` and `Self.sampleOther` - if you miss one, let me know and I will export it from Rust and "swiftify it".

## RET API's have been upgraded to:
* Not be unnecessarily throwing (e.g. `hash` and `compile` and `intent_hash` etc)
* To not return `Address` (tagged union) but rather use specific addresses, e.g. `AccountAddress` for `accountsWithdrawnFrom`
* Not require `networkID` unnecessarily - instead we read it out from the provided addresses
* *ALL* manifest building has been moved into, and fully isolated in Sargon - Sargon does NOT provide any ManifestBuilder at all
* All addresses have been upgraded with nice shiny API's such as `networkID`, [`poolKind`](https://github.com/radixdlt/sargon/blob/0dbaf885051c6c510cc8b815efdc08ee743107eb/apple/Sources/Sargon/Extensions/Methods/Address/PoolAddress%2BWrap%2BFunctions.swift#L22), [`isLegacyAddress`](https://github.com/radixdlt/sargon/blob/0dbaf885051c6c510cc8b815efdc08ee743107eb/apple/Sources/Sargon/Extensions/Methods/Address/AccountAddress%2BWrap%2BFunctions.swift#L37), [`isFungible`](https://github.com/radixdlt/sargon/blob/0dbaf885051c6c510cc8b815efdc08ee743107eb/apple/Sources/Sargon/Extensions/Methods/Address/ResourceAddress%2BWrap%2BFunctions.swift#L22) etc
* All Address, IntentHashes, NF-LocalID and NF-GlobalID have got the `formatted(_ addressFormat: AddressFormat) -> String` methods

# BROKEN / UNTESTED
Most likely broken are:
* RET errors - they have not yet been migrated - but note that some of them cannot happen any more.

# Test it out
Pre-Alpha 1.5.1 (2)


CyonAlexRDX:
supersedes https://github.com/radixdlt/babylon-wallet-ios/pull/1031

RadixWallet/Features/AccountPreferencesFeature/Children/ThirdPartyDeposits/ThirdPartyDeposits+Reducer.swift
Sajjon, line 1: Translate lowlevel Manifest building -> high level API

RadixWallet/Clients/FaucetClient/FaucetClient+Live.swift
```diff
@@ -80,10 +80,11 @@ extension FaucetClient: DependencyKey {
 			}
 
 			let networkID = await gatewaysClient.getCurrentNetworkID()
-			let manifest = try ManifestBuilder.manifestForFaucet(
+
+			let manifest = try Sargon.manifestForFaucet(
 				includeLockFeeInstruction: true,
 				networkID: networkID,
-				componentAddress: accountAddress.asGeneral
```
Sajjon, line 86: With new Sargon API we use specific addresses 🎉

RadixWallet/Clients/ImportLegacyWalletClient/ImportLegacyWalletClient+Live.swift
```diff
@@ -166,13 +168,10 @@ extension ImportLegacyWalletClient: DependencyKey {
 func convert(
 	parsedOlympiaAccount raw: Olympia.Parsed.Account
 ) throws -> OlympiaAccountToMigrate {
-	let bech32Address = try deriveOlympiaAccountAddressFromPublicKey(publicKey: raw.publicKey.intoEngine(), olympiaNetwork: .mainnet).asStr()
+	let address = try Sargon.deriveOlympiaMainnetAccountAddressFromPublicKey(
```
Sajjon, line 171: Sargon will add support for `LegacyOlympiaAccountAddress` and return it directly.

RadixWallet/Clients/OnLedgerEntitiesClient/Helpers/OnLedgerEntitiesClient+CreateEntity.swift
```diff
@@ -139,7 +139,7 @@ extension OnLedgerEntitiesClient {
 	) async throws -> OnLedgerEntity.Validator? {
 		@Dependency(\.gatewaysClient) var gatewaysClient
 		let networkId = await gatewaysClient.getCurrentNetworkID()
-		let xrdAddress = knownAddresses(networkId: networkId.rawValue).resourceAddresses.xrd.addressString()
+		let xrdAddress = Sargon.xrdAddressOfNetwork(networkId: networkId)
```
Sajjon, line 142: We only used the collection of known addresses to get XRD ResourceAddress, so we use expose API for that.

RadixWallet/Clients/OnLedgerEntitiesClient/Helpers/OnLedgerEntitiesClient+CreateEntity.swift
```diff
@@ -150,7 +150,7 @@ extension OnLedgerEntitiesClient {
 		guard let xrdResource = item
 			.fungibleResources?
 			.items
-			.first(where: { $0.resourceAddress == xrdAddress })
+			.first(where: { $0.resourceAddress == xrdAddress.address })
```
Sajjon, line 153: Gateway returns string... will be VERY nice when Gateway Client lives in Sargon, and we have a wrapper around it translating all Gateway raw models into "Ornamented" models, where all addresses are translated into our own types, and not work with raw Strings...

RadixWallet/Clients/OnLedgerEntitiesClient/OnLedgerEntitiesClient+Live.swift
```diff
@@ -88,7 +88,7 @@ extension OnLedgerEntitiesClient {
 
 		let items = try freshPage.items.map {
 			try NonFungibleGlobalId.fromParts(
-				resourceAddress: request.resource.resourceAddress.intoEngine(),
```
Sajjon, line 91: `intoEngine()` does not exist anymore, since there is not `EngineToolkit.Address`, just `Address`, which will live in Sargon - but limited to a very few APIs, mostly we use specific address types!!

RadixWallet/Clients/ROLAClient/ROLAClient+Live.swift
```diff
@@ -10,16 +10,17 @@ extension ROLAClient {
 			let entity = request.entity
 			let newPublicKey = request.newPublicKey
 
-			let entityAddress: Address = switch entity {
+			let entityAddress = switch entity {
 			case let .account(account):
-				account.address.asGeneral
+				AddressOfAccountOrPersona.account(account.address)
```
Sajjon, line 15: Quite a few API's work with **either** Account **or** Identity (Persona) Address, so will be useful to have a tagged union for it.

RadixWallet/Clients/ROLAClient/ROLAClient+Live.swift
```diff
@@ -144,34 +145,8 @@ extension ROLAClient {
...
-			self = .secp256k1(value: hashBytes)
-		case .eddsaEd25519:
-			self = .ed25519(value: hashBytes)
-		}
-	}
-}
-
 extension OnLedgerEntity.Metadata {
-	public func ownerKeyHashes() throws -> [EngineToolkit.PublicKeyHash]? {
+	public func ownerKeyHashes() throws -> [RETPublicKeyHash]? {
```
Sajjon, line 149: Shall drop the `RET` prefix... but we a naming collision with both `OnLedgerEntity.Metadata` and Gateway client models.

RadixWallet/Clients/SubmitTransactionClient/SubmitTransactionClient+Live.swift
```diff
@@ -52,38 +52,9 @@ extension SubmitTransactionClient: DependencyKey {
 		let submitTransaction: SubmitTransaction = { request in
 			let txID = request.txID
 
-			func debugPrintTX(_ decompiledNotarized: NotarizedTransaction) {
```
Sajjon, line 55: quite a few methods can be dropped if we implement this debug in Sargon, which is nice.

RadixWallet/Clients/TransactionClient/Models/TransactionModels.swift
```diff
@@ -38,14 +38,14 @@ extension GatewayAPI.TransactionPreviewRequest {
 		let notaryIsSignatory = transactionSigners.notaryIsSignatory
 
 		try self.init(
-			manifest: rawManifest.instructions().asStr(),
+			manifest: rawManifest.instructionsString(),
 			blobsHex: rawManifest.blobs().map(\.hex),
-			startEpochInclusive: .init(header.startEpochInclusive),
-			endEpochExclusive: .init(header.endEpochExclusive),
+			startEpochInclusive: .init(header.startEpochInclusive.rawValue),
```
Sajjon, line 43: This is constructing a `Gateway.TransactionHeader` which works with literals and low level types, we see a git diff here since I've upgraded the types of the (RET/Sargon)TransactoinHeader to use strong types, e.g. `NetworkID` instead of `UInt8`.

RadixWallet/Clients/TransactionClient/Models/TransactionModels.swift
```diff
@@ -132,14 +121,14 @@ public struct BuildTransactionIntentRequest: Sendable {
 	public let networkID: NetworkID
 	public let nonce: Nonce
 	public let manifest: TransactionManifest
-	public let message: Message
```
Sajjon, line 135: `Message` does not contain a case called `None` anymore, any since we do not support message encryption yet anyway, I've simplified `Message` => `case plaintext(String)` instead of the more complex type, and made this optional.

RadixWallet/Clients/TransactionClient/TransactionClient+Live.swift
```diff
@@ -49,14 +49,14 @@ extension TransactionClient {
 			func identityFromComponentAddress(_ identityAddress: IdentityAddress) async throws -> Profile.Network.Persona {
 				try await personasClient.getPersona(id: identityAddress)
 			}
-			func mapAccount(_ addresses: [EngineToolkit.Address]) throws -> OrderedSet<Profile.Network.Account> {
-				try .init(validating: addresses.asSpecific().compactMap(accountFromComponentAddress))
+			func mapAccount(_ addresses: [AccountAddress]) throws -> OrderedSet<Profile.Network.Account> {
```
Sajjon, line 52: We use specific addresses 🎉

RadixWallet/Core/Dictionary+Extensions.swift
```diff
@@ -0,0 +1,16 @@
+import Foundation
+
+extension Dictionary {
```
Sajjon, line 3: This was incorrectly put in `EngineKit` folder earlier.

RadixWallet/Exports.swift
```diff
@@ -13,7 +13,6 @@
 @_exported import Dependencies
 @_exported import DependenciesAdditions
 @_exported import Either
-@_exported import EngineToolkit
```
Sajjon, line 16: 🙅🏻‍♀️ away with you

RadixWallet/Features/ManageTrustedContactFactorSource/ManageTrustedContactFactorSource.swift
```diff
@@ -7,13 +7,10 @@ extension AccountAddress {
 	public static func random(
 		networkID: NetworkID = Radix.Gateway.default.network.id
 	) -> Self {
-		let curve25519PublicKey = Curve25519.PrivateKey().publicKey
-		let address = try! deriveVirtualAccountAddressFromPublicKey(
-			publicKey: SLIP10.PublicKey.eddsaEd25519(curve25519PublicKey).intoEngine(),
-			networkId: networkID.rawValue
+		try! Sargon.deriveVirtualAccountAddressFromPublicKey(
```
Sajjon, line 10: DEBUG only

RadixWallet/Features/SettingsFeature/ImportFromOlympiaLegacyWallet/Coordinator/ImportOlympiaWalletCoordinator.swift
```diff
@@ -582,10 +582,10 @@ extension ImportOlympiaWalletCoordinator {
 		existingAccounts: Int
 	) throws -> NonEmpty<[MigratableAccount]> {
 		let result = try scannedAccounts.enumerated().map { index, account in
-			let babylonAddress: AccountAddress = try deriveVirtualAccountAddressFromPublicKey(
-				publicKey: account.publicKey.intoEngine(),
-				networkId: networkID.rawValue
-			).asSpecific()
```
Sajjon, line 588: 🎉 cleaner

RadixWallet/Features/TransactionReviewFeature/TrackedPoolInteraction.swift
```diff
@@ -2,8 +2,8 @@ import Foundation
 
 // MARK: - TrackedPoolInteraction
 public protocol TrackedPoolInteraction {
-	var poolAddress: EngineToolkit.Address { get }
-	var poolUnitsResourceAddress: EngineToolkit.Address { get }
+	var poolAddress: ComponentAddress { get }
```
Sajjon, line 5: specific addresses ✨

RadixWallet/Features/TransactionReviewFeature/TransactionReview+Sections.swift
```diff
@@ -99,7 +98,7 @@ extension TransactionReview {
 			let allAddresses = allAddresses + resourceAddresses.asIdentifiable()
 			let resourcesInfo = try await resourcesInfo(allAddresses.elements)
 
-			let dApps = await extractDappEntities(poolAddresses)
+			let dApps = await extractDappEntities(poolAddresses.map(\.asGeneral))
```
Sajjon, line 101: will need to upgrade code in TX review to make use of the specific address, and not downcast to `Address`....

RadixWallet/Features/TransactionReviewFeature/TransactionReview+View.swift
```diff
@@ -22,10 +22,8 @@ extension TransactionReview.State {
 		.init(
 			message: {
 				// TODO: handle the rest of types
-				if case let .plainText(value) = message,
-				   case let .str(str) = value.message
-				{
-					return str
+				if case let .plainText(value) = message {
```
Sajjon, line 25: Message does not hold an inner `PlainText`type anymore, instead `String` directly. Sargon will handle MimeType for wallets.

RadixWallet/Features/TransactionReviewFeature/TransactionReview.swift
```diff
@@ -729,7 +708,10 @@ extension TransactionReview {
 			}
 		}
 		do {
-			return try addingGuarantees(to: manifest, guarantees: state.allGuarantees)
+			return try Sargon.updatingManifestAddGuarantees(
```
Sajjon, line 711: `mutatingManifest`? `changingManifest`? `modifyManifest`? or `updating` ? which is most clear?

RadixWallet/Profile/Address/AddressProtocol.swift
```diff
@@ -8,6 +6,12 @@ public protocol AddressProtocol {
 	var address: String { get }
 }
 
+extension AddressProtocol {
+	public func networkId() -> NetworkID {
```
Sajjon, line 10: The idea is that ALL Address types will have: 
```
address: String // the bech32 encoded address
bytes: Data // the payload, being data part of bech32, if bech32 decoded
networkID: NetworkID // can be decoded with RET
```

RadixWallet/SARGON/NewSargonTypes/AssetTransferRepresentation.swift
```diff
@@ -0,0 +1,38 @@
+import Foundation
+
+// MARK: - AssetsTransfersTransactionPrototype
+// NOT FINAL, just a sketch, probably wrong...
+public struct AssetsTransfersTransactionPrototype {
```
Sajjon, line 5: Non final representation of the state created by user in UI before she presses "Continue" from AssetTransfer

RadixWallet/SARGON/WrappingRadixEngineToolkit/ExtensionsUsedOnTypes/OnlyPossibleInSwift/DummySargonCodable.swift
```diff
@@ -0,0 +1,13 @@
+import Foundation
+
+// MARK: - DummySargonCodable
+public protocol DummySargonCodable: Codable {}
+extension DummySargonCodable {
```
Sajjon, line 5: Some types from RET are made codable by iOS wallet since our Cache client requires it. When caching is done by Sargon, this can go away. And we really should :)

RadixWallet/SARGON/WrappingRadixEngineToolkit/ExtensionsUsedOnTypes/OnlyPossibleInSwift/Sargon+Swift+Decimal.swift
Sajjon, line 1: these kind of extension we will want to keep, making the Sargon Decimal swifty!!

RadixWallet/SARGON/WrappingRadixEngineToolkit/ExtensionsUsedOnTypes/OnlyPossibleInSwift/Sargon+Swift+PublicKey.swift
```diff
@@ -0,0 +1,12 @@
+import Foundation
+
+extension SLIP10.PublicKey {
+	public var bytes: Data {
```
Sajjon, line 4: already present in Sargon, but can be made into a computed property like this.

RadixWallet/SARGON/WrappingRadixEngineToolkit/HighLevel/RET_Sargon_API_dummy.swift
Sajjon, line 1: This is de facto the API surface of RET inside vendored by Sargon - not at all big! 🎉

RadixWallet/SARGON/WrappingRadixEngineToolkit/ExtensionsUsedOnTypes/OnlyPossibleInSwift/DummySargonCodable.swift
code-reviewers-rdx: Note that caching with Sargon might be harder, as Android uses an actual database, like CoreData, AFAIK.

RadixWallet/Features/TransactionReviewFeature/TransactionReview.swift
GhenadieVP: `modyManifest` sounds good

RadixWallet/SARGON/WrappingRadixEngineToolkit/ExtensionsUsedOnTypes/OnlyPossibleInSwift/DummySargonCodable.swift
Sajjon: We can let caching take place in iOS and in Android for now, and just provide a simple interface for it, and let Sargon be able to tell FFI host "load_cached_value_for_key" - and we can let key be a String for now. I think that is feasible.

GhenadieVP: That, or we can have Sargon ask for a given resource by address, and it is up to the Wallet to use the current infra to return either the value from the GW or from cache.

Sajjon: yes indeed.

RadixWallet/Core/SharedModels/LedgerIdentifiable.swift
```diff
@@ -114,21 +114,22 @@ extension LedgerIdentifiable {
...
-		case _ where PackageEntityType.addressSpace.contains(address.decodedKind):
-			self = .package(.init(address: address.address, decodedKind: address.decodedKind))
-		case _ where ValidatorEntityType.addressSpace.contains(address.decodedKind):
-			self = .validator(.init(address: address.address, decodedKind: address.decodedKind))
-		case _ where ComponentEntityType.addressSpace.contains(address.decodedKind):
-			self = .component(.init(address: address.address, decodedKind: address.decodedKind))
-		default:
-			return nil
-		}
+//		switch address.decodedKind {
```
CyonAlexRDX, line 117: REVERT

RadixWallet/Features/AccountPreferencesFeature/Children/ThirdPartyDeposits/ThirdPartyDeposits+Reducer.swift
```diff
@@ -165,176 +165,11 @@ public struct ManageThirdPartyDeposits: FeatureReducer, Sendable {
...
-		case .acceptKnown:
-			2
-		}
-		return .enumValue(discriminator: discriminator, fields: [])
+	private func prepareForSubmission(
+		newConfig new: ThirdPartyDeposits
+	) throws -> (manifest: TransactionManifest, updatedAccount: Profile.Network.Account) {
+//		let manifest = TransactionManifest.thirdPartyDepositUpdate(to: new)
+		//        return (manifest, updatedAccount)
+		fatalError()
```
CyonAlexRDX, line 173: REVERT / FIX

RadixWallet/Features/AssetsFeature/Components/HelperViews/ResourceBalance/ResourceBalance+Helpers.swift
```diff
@@ -88,7 +88,7 @@ extension ResourceBalance: Comparable {
 // MARK: - ResourceBalance.Amount + Comparable
 extension ResourceBalance.Amount: Comparable {
 	public static func < (lhs: Self, rhs: Self) -> Bool {
-		// If RETDecimal were comparable:
+		// If Decimal192 were comparable:
```
CyonAlexRDX, line 91: REVERT / FIX we can make them comparable!

RadixWallet/Features/AssetsFeature/Components/StakeUnitList/StakeUnitList.swift
```diff
@@ -274,30 +280,40 @@ public struct StakeUnitList: Sendable, FeatureReducer {
...
 				stakeClaims: [
-					.init(
-						validatorAddress: stakeClaim.validatorAddress,
-						resourceAddress: resourceAddress,
-						ids: .init(stakeClaim.id.localId()),
-						amount: stakeClaim.claimAmount.nominalAmount
-					),
+					//					.init(
+//						validatorAddress: stakeClaim.validatorAddress,
+//						resourceAddress: nonFungibleResourceAddress,
```
CyonAlexRDX, line 295: REVERT / FIXME remove deadcode

RadixWallet/Features/DappInteractionFeature/Coordinator/DappInteractionFlow.swift
```diff
@@ -1041,8 +1041,9 @@ extension DappInteractionFlow.Path.State {
 				nonce: .secureRandom(),
 				signTransactionPurpose: .manifestFromDapp,
 				message: item.message.map {
-					Message.plainText(value: .init(mimeType: "text", message: .str(value: $0)))
-				} ?? .none,
+//					Message.plainText(value: $0)
+					Message.plainText(plaintext: .init(mimeType: "application/text", message: .stringMessage(string: $0)))
```
CyonAlexRDX, line 1045: REVERT / FIXME add constructor in Sargon so that we do not need to specify mime type here!

RadixWallet/Clients/AccountPortfoliosClient/AccountPortfoliosClient+Live.swift
```diff
@@ -66,12 +66,17 @@ extension AccountPortfoliosClient: DependencyKey {
 		}
 
 		@Sendable
-		func fetchAccountPortfolios(_ accountAddresses: [AccountAddress], _ forceRefresh: Bool) async throws -> [AccountPortfolio] {
+		func fetchAccountPortfolios(
+			_ accountAddresses: [AccountAddress],
+			_ forceRefreshIgnored: Bool
+		) async throws -> [AccountPortfolio] {
+			let forceRefresh = true
```
CyonAlexRDX, line 73: ‼️ ‼️ ‼️  REVERT! I'm debugging bugs in Cache ‼️ ‼️ ‼️

RadixWallet/Clients/AccountPortfoliosClient/AccountPortfoliosClient+Live.swift
```diff
@@ -128,9 +133,14 @@ extension AccountPortfoliosClient: DependencyKey {
 		}
 
 		@Sendable
-		func fetchAccountPortfolio(_ accountAddress: AccountAddress, _ forceRefresh: Bool) async throws -> AccountPortfolio {
+		func fetchAccountPortfolio(
+			_ accountAddress: AccountAddress,
+			_ forceRefreshIgnored: Bool
+		) async throws -> AccountPortfolio {
+			let forceRefresh = true
```
CyonAlexRDX, line 140: ‼️ ‼️ ‼️ REVERT! I'm debugging bugs in Cache ‼️ ‼️ ‼️

RadixWallet/Clients/ROLAClient/ROLAClient+Live.swift
```diff
@@ -10,35 +10,38 @@ extension ROLAClient {
...
+				}
+			}()
+
+			let metadata = try await onLedgerEntitiesClient.getEntity(entityAddress.embed(), metadataKeys: [.ownerKeys]).genericComponent?.metadata
 			var ownerKeyHashes = try metadata?.ownerKeyHashes() ?? []
 
 			let transactionSigningKeyHash: PublicKeyHash = switch entity.securityState {
 			case let .unsecured(control):
-				try .init(hashing: control.transactionSigning.publicKey)
+				.init(hashing: control.transactionSigning.publicKey.intoSargon())
```
Sajjon, line 29: We have keys in iOS wallet AND Sargon now - so need to map some keys into Sargon ones until we have migrated over to using Profile in Sargon

RadixWallet/Clients/SubmitTransactionClient/SubmitTransactionClient+Interface.swift
```diff
@@ -10,23 +10,23 @@ public struct SubmitTransactionClient: Sendable {
 }
 
 extension SubmitTransactionClient {
-	public typealias SubmitTransaction = @Sendable (SubmitTXRequest) async throws -> TXID
-	public typealias HasTXBeenCommittedSuccessfully = @Sendable (TXID) async throws -> Void
+	public typealias SubmitTransaction = @Sendable (SubmitTXRequest) async throws -> IntentHash
```
Sajjon, line 13: TXID has always been just s typealias for IntentHash, we ought to start using the correct term..

RadixWallet/Clients/SubmitTransactionClient/SubmitTransactionClient+Live.swift
```diff
@@ -10,7 +12,7 @@ extension SubmitTransactionClient: DependencyKey {
 
 			@Sendable func pollTransactionStatus() async throws -> GatewayAPI.TransactionStatusResponse {
 				let txStatusRequest = GatewayAPI.TransactionStatusRequest(
-					intentHash: txID.asStr()
+					intentHash: txID.bech32EncodedTxId
```
Sajjon, line 15: Or `formatted(.raw)` which is the samething - also same: `.description` and `toRawString()`

RadixWallet/Clients/TransactionClient/Models/TransactionFailure.swift
```diff
@@ -151,7 +151,7 @@ extension TransactionFailure {
 			case .failedToCompileTXIntent:
 				"Failed to compile transaction intent"
 			case .failedToGenerateTXId:
-				"Failed to generate TXID"
+				"Failed to generate IntentHash"
```
Sajjon, line 154: ‼️ revert - user facing?

RadixWallet/Clients/TransactionClient/Models/TransactionModels.swift
```diff
@@ -132,14 +136,14 @@ public struct BuildTransactionIntentRequest: Sendable {
...
+	public let message: Message?
 	public let makeTransactionHeaderInput: MakeTransactionHeaderInput
 	public let transactionSigners: TransactionSigners
 
 	public init(
 		networkID: NetworkID,
 		manifest: TransactionManifest,
-		message: Message,
```
Sajjon, line 142: Revert, Message once again has a case None.

RadixWallet/Core/SharedModels/Assets/OnLedgerEntity.swift
```diff
@@ -178,17 +208,27 @@ extension OnLedgerEntity {
 		case dappDefinition = "dapp definition"
 	}
 
-	public struct Resource: Sendable, Hashable, Codable, Identifiable {
+	public struct Resource: Sendable, Hashable, Codable, Identifiable, CustomDebugStringConvertible {
 		public var id: ResourceAddress { resourceAddress }
 		public let resourceAddress: ResourceAddress
 		public let atLedgerState: AtLedgerState
-		public let divisibility: Int?
+		public let divisibility: UInt8?
```
Sajjon, line 215: Negative divisibility makes no sense

RadixWallet/Core/SharedModels/Assets/OnLedgerEntity.swift
```diff
@@ -178,17 +208,27 @@ extension OnLedgerEntity {
...
+			symbol: \(metadata.symbol ?? "???")
+			name: \(metadata.name ?? "???")
+			icon: \(metadata.iconURL?.absoluteString ?? "???")
+			description: \(metadata.description ?? "???")
+			"""
+		}
+
 		public var fungibility: Fungibility {
-			if case .globalFungibleResourceManager = resourceAddress.decodedKind {
+			if resourceAddress.isFungible {
```
Sajjon, line 231: 🥳

RadixWallet/Core/SharedModels/LedgerIdentifiable.swift
```diff
@@ -42,7 +42,7 @@ extension LedgerIdentifiable {
 		public func formatted(_ format: AddressFormat = .default) -> String {
 			switch self {
 			case let .transaction(txId):
-				txId.formatted(format)
+				txId.bech32EncodedTxId
```
Sajjon, line 45: ‼️ revert this re-introduced the incorrect formatting of TXID bug

RadixWallet/Prelude/Logger.swift
```diff
@@ -6,7 +6,7 @@ private let baseLabel = "com.radixpublishing"
 
 private func makeLogger(
 	label: String,
-	level: Logger.Level = .debug
+	level: Logger.Level = .error
```
matiasbzurovski, line 9: is this intentional?

RadixWallet/Features/TransactionReviewFeature/TransactionReviewDapps/TransactionReviewDapps.swift
```diff
@@ -1,13 +1,12 @@
 import ComposableArchitecture
+import Sargon
```
matiasbzurovski, line 2: We already have it on `Exports.swift`

RadixWallet/Clients/FaucetClient/FaucetClient+Live.swift
```diff
@@ -123,22 +122,6 @@ extension FaucetClient: DependencyKey {
...
-//					)
-//				} else {
-//					return try TransactionManifest.manifestForCreateMultipleNonFungibleTokens(
-//						account: request.recipientAccountAddress,
-//						network: networkID
-//					)
-//				}
-//			}()
-
-//			try await signSubmitTX(manifest: manifest)
```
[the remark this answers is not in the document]
Sajjon: Thx for reminding me!

RadixWallet/Clients/OnLedgerEntitiesClient/OnLedgerEntitiesClient+Interface.swift
```diff
@@ -111,7 +111,7 @@ extension OnLedgerEntitiesClient {
 }
 
 extension OnLedgerEntitiesClient {
-	public struct CachingStrategy: Sendable, Hashable {
+	public struct CachingStrategy: Sendable, Hashable, CustomDebugStringConvertible {
```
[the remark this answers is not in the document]
Sajjon: We had it like that, but wanted more flexibility. There might be other dimension being added here later, i.e. expiration date, which is easier to model with a struct.

RadixWallet/Clients/TransactionClient/Models/TransactionFailure.swift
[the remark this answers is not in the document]
Sajjon: already reverted.

RadixWallet/Prelude/Logger.swift
Sajjon: good catch

RadixWallet/Features/TransactionReviewFeature/TransactionReviewDapps/TransactionReviewDapps.swift
Sajjon: Yes, but Xcode is...crap. re-exporting makes code compile, alas, it does NOT give code highlighting :/ :/ at least not in Xcode 15.3

RadixWallet/Features/AssetTransferFeature/AssetTransfer+Reducer.swift
```diff
@@ -166,96 +167,91 @@ extension AssetTransfer {
...
+			nonFungibleResources: nonFungibles
+		)
+
+		return TransactionManifest.assetsTransfers(transfers: perAssetTransfers)
 	}
 }
 
-func instructionForDepositing(
-	bucket: ManifestBuilderBucket,
+func useTryAbortOrDeposit(
```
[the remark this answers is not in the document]
Sajjon: haha `tryAbort` :D :D :D

RadixWallet/Features/AccountPreferencesFeature/Children/DevAccountPreferences+Reducer.swift
```diff
@@ -148,7 +151,11 @@ public struct DevAccountPreferences: Sendable, FeatureReducer {
 		case .createFungibleTokenButtonTapped:
 			return .run { [accountAddress = state.address] send in
 				let accountAddress = try await accountsClient.getAccountByAddress(accountAddress)
-				let manifest = try ManifestBuilder.manifestForCreateFungibleToken(account: accountAddress.address, networkID: accountAddress.networkID)
+
```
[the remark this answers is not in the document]
matiasbzurovski: we can also remove it on the following cases `.createNonFungibleTokenButtonTapped`, `.createMultipleFungibleTokenButtonTapped`, `.createMultipleNonFungibleTokenButtonTapped`.

RadixWallet/Clients/OnLedgerEntitiesClient/OnLedgerEntitiesClient+Interface.swift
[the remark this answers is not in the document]
matiasbzurovski: Following your example, if we ever need an `expirationDate: Date?` property, it would only apply for the `useCache` case. Therefore, it would make more sense to have it as an associated value of the given case, than as an optional variable of the struct. With enum solution we would make sure no `CacheingStrategy` is initialized to `useCache` without an expiration date.

I know that expiration date was only an example, and you may argue that there could be new fields that apply to every case, so I'd have it like this:
<details>
  <summary>Code snippet</summary>
  
```Swift
struct CachingStrategy: String, Sendable, Hashable, CustomDebugStringConvertible {
    let kind: Kind
    let newCommonField: String
    
    enum Kind {
        case forceUpdate
        case useCache(Date)
        case readFromLedgerSkipWrite
        
        public var debugDescription: String {
            rawValue
        }
    }
    
    var read: Read {
        switch kind {
        case .forceUpdate:
            return .fromLedger
        case .useCache:
            return .fromCache
        case .readFromLedgerSkipWrite:
            return .fromLedger
        }
    }
    
    var write: Write {
        switch kind {
        case .forceUpdate:
            return .toCache
        case .useCache:
            return .toCache
        case .readFromLedgerSkipWrite:
            return .skip
        }
    }

    var date: Date? {
        switch kind {
            case let .useCache(value):
                return value
            case .forceUpdate, .readFromLedgerSkipWrite:
                return nil
        }
    }
}

extension CachingStrategy {
    enum Read: Sendable, Hashable {
        case fromCache
        case fromLedger
    }

    enum Write: Sendable, Hashable {
        case toCache
        case skip
    }
}
```
</details>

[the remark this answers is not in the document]
CyonAlexRDX: Ah I remember now, I wanted to be able to have "presets" but still have the freedom to do e.g. `(read: false, save: true)` when debugging. I dont feel strongly about this... but lets keep it out of scope for this PR. Good that you wanna pursue enums by default, I like that!

RadixWallet/Features/TransactionReviewFeature/TransactionReview.swift
```diff
@@ -433,7 +433,9 @@ public struct TransactionReview: Sendable, FeatureReducer {
 			}
 			state.destination = .customizeFees(.init(
 				reviewedTransaction: reviewedTransaction,
-				manifestSummary: reviewedTransaction.transactionManifest.summary(networkId: reviewedTransaction.networkID.rawValue),
+				manifestSummary: reviewedTransaction
+					.transactionManifest
+					.summary,
```
GhenadieVP, line 438: No networkID required, neat

RadixWallet/Features/TransactionReviewFeature/TransactionReview.swift
```diff
@@ -721,17 +699,17 @@ extension TransactionReview {
 		var manifest = reviewedTransaction.transactionManifest
 		if case let .success(feePayerAccount) = reviewedTransaction.feePayer.unwrap()?.account {
 			do {
-				manifest = try reviewedTransaction.transactionManifest.withLockFeeCallMethodAdded(
-					address: feePayerAccount.address.asGeneral,
-					fee: reviewedTransaction.transactionFee.totalFee.lockFee
+				manifest = try reviewedTransaction.transactionManifest.modify(
```
GhenadieVP, line 702: Is this using the modify from RET? I recall that we didn't want to use it because it is extract smart, and would add `lock_fee_and_withdraw` than just `lock_fee` instruction, but maybe it is not the case anymore.

RadixWallet/Profile/Entity/OnLedgerSettings/Account+OnLedgerSettings.swift
```diff
@@ -191,7 +189,7 @@ extension Profile.Network.Account.OnLedgerSettings.ThirdPartyDeposits.DepositorA
 
 		case let .nonFungibleGlobalID(id):
 			try container.encode(Discriminator.nonFungibleGlobalID, forKey: .discriminator)
-			try container.encode(id.asStr(), forKey: .value)
+			try container.encode(id.description, forKey: .value)
```
GhenadieVP, line 192: Again, strange to see that `description` is used.

RadixWallet/Clients/SubmitTransactionClient/SubmitTransactionClient+Interface.swift
Sajjon: They are semantically different things. `IntentHash` *has* `txID`, but *is* not per se. the TXID is the bech32 encoding of the hash contained in `IntentHash`. The `IntentHash` also has a `networkID`. 

We should use `txID` - refering to `intentHash.formatted(.raw)` or `intentHash. bech32EncodedTxId` (they are the same) only when we need the bech32 encoded string - and we ought to do it "as late as possible". Since it is good to retain as much info as possible (i.e. `networkID` of an `IntentHash`).

GhenadieVP: but still, it is wrong to have a property `txID: IntentHash`, right?

RadixWallet/Clients/OnLedgerEntitiesClient/Helpers/OnLedgerEntitiesClient+ComplexTypes.swift
```diff
@@ -1,4 +1,6 @@
 import Foundation
+import Sargon
+import SargonUniFFI
```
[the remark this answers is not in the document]
Sajjon: used before I had Swiftified some SargonUniFFI methods. Even though Sargon re-exports SargonUniFFI, one must import it to get code completion (re-export only makes code compile :/ )

RadixWallet/Features/TransactionReviewFeature/TransactionReview+Sections.swift
```diff
@@ -58,30 +61,31 @@ extension TransactionReview {
...
+				accountWithdraws: summary.withdrawals,
 				newlyCreatedNonFungibles: summary.newlyCreatedNonFungibles,
 				entities: resourcesInfo,
-				userAccounts: userAccounts,
 				networkID: networkID
 			)
 
-			let dAppAddresses = summary.encounteredEntities.filter { $0.entityType() == .globalGenericComponent }
-			let dAppsUsed: TransactionReviewDappsUsed.State? = try await extractDapps(dAppAddresses, unknownTitle: L10n.TransactionReview.unknownComponents)
+			let dAppAddresses = summary.encounteredComponentAddresses
```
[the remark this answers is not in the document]
Sajjon: Oh right, I'm adding `isGlobal` (and `isInternal`) computed properties on `ComponentAddress` in Sargon. Then I can change this line to:
`let dAppAddresses = summary.encounteredComponentAddresses.filter(\.isGlobal)` - good catch!

RadixWallet/Features/TransactionReviewFeature/TransactionReview.swift
Sajjon: I'm using Scrypto's transaction builder, and it does:

```rust
    InstructionV1::CallMethod {
            address: address.into(),
            method_name: "lock_fee",
            <AMOUNT>
     }
```

do you think it is wrong?

RadixWallet/Clients/SubmitTransactionClient/SubmitTransactionClient+Interface.swift
Sajjon: I dont care too much about a few properties with inaccurate name in grand scheme of things. We can rename them as we go if you like.

RadixWallet/Clients/OnLedgerEntitiesClient/Helpers/OnLedgerEntitiesClient+CreateEntity.swift
CyonAlexRDX: Yes my plan is one layer which turns String into Addresses and Decimal192'etc real Sargon types - without performing any application logic mapping - and upon making network call, the response is mapped to those Sargonified types. 

This has the advantage that we catch any bugs in GW early, regarding the fetch as failure if we failed to deserialise and try into Sargonified. 

What we later do with GW types - is application logic. We might chose to save a subset of the data in some
Local cache - similar to what is call OnLedgerEntity in iOS wallet. That code will be trivial, since it is about picking property from already parsed and Sargonified types.

CyonAlexRDX: But I just realised that we can do that we a script that modifies the generated models. 

The openapi generator is completely broken for Rust anyway, so we need to modify the schema file, either we modified the schema file further to generate Decimal192 and typed addresses instead of Strings- not trivially done, but the modifications we need to do just to get models to work is hard however we do it. Or we make a simple post processor script which upgrade strings to Decimal192 and typed addresses.

RadixWallet/Features/TransactionReviewFeature/TransactionReview.swift
CyonAlexRDX: Based on testing I think this is correct...

RadixWallet/Clients/SubmitTransactionClient/SubmitTransactionClient+Interface.swift
GhenadieVP: you sounded somewhat categorical about using the correct term with your initial comment, so I thought you wanted to pursue it to the fullest.

RadixWallet/Features/TransactionReviewFeature/TransactionReview.swift
GhenadieVP: That is why I was asking if you are using the RET manifest modify. And I see that you are not and the behaviour is as expected 👍.

GhenadieVP (approved):
Tremendous work, bravo 💯 🥇!!! So many months of work finally come to fruition, and in what a nice clean way 👏.
Can't wait to have 90% of the codebase in Sargon 😄

RadixWallet/Clients/OnLedgerEntitiesClient/Helpers/OnLedgerEntitiesClient+CreateEntity.swift
GhenadieVP: yes, absolutely, if we are to do this it needs to be automated. Otherwise we might end up maintaining by hand lots of models which are not even used by Sargon/Wallet.

RadixWallet/Clients/SubmitTransactionClient/SubmitTransactionClient+Interface.swift
CyonAlexRDX: That was not my intent (pun intended (double pun intended)) - I forgot to write some emoji.

I just think it is good to avoid confusion - at least on type level, so "TXID" ought to be not used.

RadixWallet/Features/AssetsFeature/Components/HelperViews/ResourceBalance/ResourceBalance+Helpers.swift
kugel3: If you add that, you can uncomment this bit and remove the comment!

kugel3 (approved):
Amazing work, many great improvements. Most comments are style related. Regarding all the new debug strings, shouldn't they be returned by Sargon, or are they all on types that only exist on iOS?

RadixWallet/Clients/AccountPortfoliosClient/AccountPortfoliosClient+Live.swift
```diff
@@ -91,12 +94,12 @@ extension AccountPortfoliosClient: DependencyKey {
...
 				if gateway == .mainnet {
 					/// Only Mainnet resources have prices
-					return (currentAccounts + accounts).flatMap(\.resourcesWithPrices) + [.mainnetXRDAddress]
+					return (currentAccounts + accounts).flatMap(\.resourcesWithPrices) + [.mainnetXRD]
 				} else {
 					#if DEBUG
 					/// Helpful for testing on stokenet
 					return [
-						.mainnetXRDAddress,
+						.mainnetXRD,
```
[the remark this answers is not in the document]
CyonAlexRDX: not my code, done by @GhenadieVP I think to be able to get some prices even for Stokenet - since Stokenet does not have prices.

RadixWallet/Clients/FaucetClient/FaucetClient+Interface.swift
```diff
@@ -3,109 +3,18 @@ public struct FaucetClient: Sendable {
 	public var getFreeXRD: GetFreeXRD
 	public var isAllowedToUseFaucet: IsAllowedToUseFaucet
 
-	#if DEBUG
-	public var createFungibleToken: CreateFungibleToken
-	public var createNonFungibleToken: CreateNonFungibleToken
```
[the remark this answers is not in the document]
CyonAlexRDX: not used in `main` - we use the `TransactionManifest.staticMethodBuildingAManifestInsideOfSargon` directly in the reducer of `DevPreferencesReducer`.

RadixWallet/Clients/CacheClient/CacheClient+Interface.swift
```diff
@@ -69,16 +69,85 @@ extension CacheClient {
...
+		public init(address: some AddressProtocol) {
+			self = .onLedgerEntity(.init(address: address))
+		}
+
 		public enum OnLedgerEntity: Hashable {
-			case account(Address)
-			case resource(Address)
-			case resourcePool(Address)
-			case validator(Address)
-			case genericComponent(Address)
```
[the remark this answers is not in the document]
kugel3: No, as long as it's the case that a specific address can only belong to one (old) case, it should not make a difference. 

Speaking of addresses, is it the case that we now only have one set of addresses, and that the old engine address is now the same as the "general" one of the "specific" addresses? We don't use the general/specific terminology anymore I hope, it's a bit misleading.

RadixWallet/Clients/OnLedgerEntitiesClient/Helpers/OnLedgerEntitiesClient+ComplexTypes.swift
```diff
@@ -204,8 +204,8 @@ extension OnLedgerEntitiesClient {
 			let tokens = try await getNonFungibleTokenData(.init(
 				resource: resourceAddress,
 				nonFungibleIds: existingTokenIds.map {
-					try NonFungibleGlobalId.fromParts(
-						resourceAddress: resourceAddress.intoEngine(),
+					NonFungibleGlobalID(
+						resourceAddress: resourceAddress,
```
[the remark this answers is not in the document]
CyonAlexRDX: No, this cannot fail anymore - RET was overly throwing.

RadixWallet/Clients/OnLedgerEntitiesClient/Helpers/OnLedgerEntitiesClient+CreateEntity.swift
```diff
@@ -87,26 +100,27 @@ extension OnLedgerEntitiesClient {
 
 	@Sendable
 	static func createResource(
+		resourceAddress: ResourceAddress,
```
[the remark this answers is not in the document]
CyonAlexRDX: TODO: sure I can drop the prefixes.

RadixWallet/Clients/OnLedgerEntitiesClient/Helpers/OnLedgerEntitiesClient+CreateEntity.swift
```diff
@@ -116,6 +130,7 @@ extension OnLedgerEntitiesClient {
 
 	@Sendable
 	static func createResourcePool(
+		poolAddress: PoolAddress,
```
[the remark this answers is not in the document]
CyonAlexRDX: TODO: drop prefix

RadixWallet/Clients/OnLedgerEntitiesClient/Helpers/OnLedgerEntitiesClient+CreateEntity.swift
```diff
@@ -217,9 +233,9 @@ extension OnLedgerEntitiesClient {
 		}
 
 		let stakeAndPoolAddresses = Set(
-			stakeUnitCandidates.compactMap(\.metadata.validator?.asGeneral)
-				+ stakeClaimNFTCandidates.compactMap(\.metadata.validator?.asGeneral)
-				+ poolUnitCandidates.compactMap(\.metadata.poolUnit?.asGeneral)
+			stakeUnitCandidates.compactMap(\.metadata.validator).map { $0.embed() }
```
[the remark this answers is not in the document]
CyonAlexRDX: (No UniFFI restrictions)

@kugel3 it is inspired by PF Case Path terminology: https://github.com/pointfreeco/swift-case-paths?tab=readme-ov-file#computed-paths

where `extract` is to try go from Enum -> Case, and inverse operation is `embed`, going from (Associated Value of) Case -> Enum.

But sure we can use `.asGeneral` again - we are "free" to use that now since "integration is complete".

RadixWallet/Clients/TokenPriceClient/TokenPriceClient+Live.swift
```diff
@@ -44,14 +44,14 @@ extension TokenPricesClient.TokenPrices {
 		let formatter = NumberFormatter()
 		formatter.locale = Locale(identifier: "en_US_POSIX") // Just ignore the users locale
 		formatter.numberStyle = .decimal
-		formatter.maximumFractionDigits = Int(RETDecimal.maxDivisibility)
+		formatter.maximumFractionDigits = Int(Decimal192.maxDivisibility)
 		formatter.roundingMode = .down
-		formatter.decimalSeparator = "." // Enforce dot notation for RETDecimal
-		formatter.usesGroupingSeparator = false // No grouping separator for RETDecimal
+		formatter.decimalSeparator = "." // Enforce dot notation for Decimal192
+		formatter.usesGroupingSeparator = false // No grouping separator for Decimal192
```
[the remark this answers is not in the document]
CyonAlexRDX: TODO: @kugel3 thx for comment - we should remove this completely and use `Decimal192.init(_ value: Double)`

RadixWallet/Clients/OnLedgerEntitiesClient/Helpers/OnLedgerEntitiesClient+CreateEntity.swift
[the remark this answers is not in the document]
CyonAlexRDX: TODO: changed `func embed` -> `var asGeneral`

RadixWallet/Clients/TransactionClient/TransactionClient+Live.swift
```diff
@@ -120,38 +120,38 @@ extension TransactionClient {
...
+				startEpochInclusive: epoch,
+				endEpochExclusive: epoch + request.makeTransactionHeaderInput.epochWindow,
+				nonce: request.nonce,
+				notaryPublicKey: SLIP10.PublicKey.eddsaEd25519(request.transactionSigners.notaryPublicKey).intoSargon(),
 				notaryIsSignatory: request.transactionSigners.notaryIsSignatory,
 				tipPercentage: request.makeTransactionHeaderInput.tipPercentage
 			)
 
-			return .init(header: header, manifest: request.manifest, message: request.message)
+			return .init(header: header, manifest: request.manifest, message: request.message ?? Message.none)
```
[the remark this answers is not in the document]
CyonAlexRDX: Best to keep it as is in Sargon (rust side) and simplest in Swift to not work with an Optional - without the need for one.

RadixWallet/Core/SargonExtensions/IntoSargon+Bridge.swift
```diff
@@ -0,0 +1,255 @@
...
+extension DerivationPath {
+	func intoSargon() -> Sargon.DerivationPath {
+		if isGetID {
+			return Sargon.DerivationPath.cap26(value: .getId(value: .default))
+		}
+
+		if let olympiaAccountPath = try? self.asLegacyOlympiaBIP44LikePath() {
+			return Sargon.DerivationPath.bip44Like(value: olympiaAccountPath.intoSargon())
+		} else if let _ = try? self.asAccountPath() {
+			return try! Sargon.DerivationPath.cap26(
```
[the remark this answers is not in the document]
CyonAlexRDX: This is HIGHLY temporary code - hope to remove it in a couple of weeks. So I see no need to make this code pretty :)

If that is alright with you?

RadixWallet/Core/SharedModels/Assets/FiatWorth.swift
```diff
@@ -1,7 +1,7 @@
 // MARK: - FiatWorth
 public struct FiatWorth: Sendable, Hashable {
 	enum Worth: Sendable, Hashable {
```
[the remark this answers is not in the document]
CyonAlexRDX: Not my code :) but I think @GhenadieVP might have had an idea modelling as such - but agree, looks like an optional can be used. 

Maybe semantically `Decimal192.none` would be "have not received any price yet" - vs `Worth.unknown` means "Have fetched prices from GW, but it had no price for this token".

RadixWallet/Core/SharedModels/Assets/OnLedgerEntity.swift
```diff
@@ -240,6 +280,7 @@ extension OnLedgerEntity {
 extension OnLedgerEntity {
 	public struct NonFungibleToken: Sendable, Hashable, Identifiable, Codable {
 		public typealias NFTData = GatewayAPI.ProgrammaticScryptoSborValueTuple
+		public typealias ID = NonFungibleGlobalId
```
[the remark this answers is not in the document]
CyonAlexRDX: I put some logging in the inits I think. I can remove them.

TODO: Remove unnecessary inits

RadixWallet/Features/AssetTransferFeature/AssetTransfer+Reducer.swift
```diff
@@ -166,96 +167,89 @@ extension AssetTransfer {
...
-
-		return try await ManifestBuilder.make {
-			for resource in involvedFungibleResources {
-				let divisibility = resource.divisibility.map(UInt.init) ?? RETDecimal.maxDivisibility
-				try ManifestBuilder.withdrawAmount(
-					accounts.fromAccount.address.intoEngine(),
-					resource.address.intoEngine(),
-					resource.totalTransferAmount.rounded(decimalPlaces: divisibility)
-				)
+		let _involvedFungibleResources = try await extractInvolvedFungibleResources(accounts.receivingAccounts)
```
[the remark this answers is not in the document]
CyonAlexRDX: TODO: remove `_` prefixes from both variables

RadixWallet/Core/SharedModels/Assets/FiatWorth.swift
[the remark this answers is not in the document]
kugel3: Also, you can't make an Optional Comparable. Or you can, but only once.

RadixWallet/Core/SharedModels/Assets/OnLedgerEntity.swift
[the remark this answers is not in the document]
kugel3: I meant the ID typealias actually

RadixWallet/Core/SargonExtensions/IntoSargon+Bridge.swift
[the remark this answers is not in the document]
kugel3: Sure of course

RadixWallet/Features/AssetTransferFeature/AssetTransfer+Reducer.swift
```diff
@@ -166,96 +167,89 @@ extension AssetTransfer {
...
+func useTryDepositOrAbort(
 	resource: ResourceAddress,
-	into receivingAccount: ReceivingAccount.State.Account
-) async throws -> ManifestBuilder.InstructionsChain.Instruction {
-	let recipientAddress = receivingAccount.address
-
-	if case let .left(userAccount) = receivingAccount {
+	into receivingAccount: AssetsTransfersRecipient
+) async throws -> Bool {
+	if case let .myOwnAccount(sargonUserAccount) = receivingAccount {
```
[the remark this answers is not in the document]
CyonAlexRDX: TODO: rename to `usersAccount`

RadixWallet/Features/AssetTransferFeature/Components/ChooseAccount/ChooseReceivingAccount+View.swift
```diff
@@ -128,7 +128,11 @@ extension ChooseReceivingAccount.View {
 			viewStore.chooseAccounts.selectedAccounts?.first?.account,
 			or: viewStore.validateAccountAddress,
 			forAction: { result in
-				viewStore.send(.chooseButtonTapped(result))
+				let recipient: AssetsTransfersRecipient = switch result {
+				case let .left(account): .myOwnAccount(value: account.intoSargon())
+				case let .right(address): .foreignAccount(value: address)
```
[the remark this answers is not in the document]
CyonAlexRDX: sure!
TODO: rename to`otherUser` and `thisUser`

RadixWallet/Features/AssetTransferFeature/Components/TransferAccountList/ReceivingAccount/Asset/NonFungibleResourceAsset+Reducer+View.swift
```diff
@@ -1,16 +1,20 @@
...
 	public struct State: Sendable, Hashable, Identifiable {
 		public typealias ID = String
-		public var id: ID { token.id.asStr() }
+		public var id: ID { nftToken.id.toRawString() }
 
 		public let resourceImage: URL?
 		public let resourceName: String?
 		public let resourceAddress: ResourceAddress
-		public let token: OnLedgerEntity.NonFungibleToken
+		public let nftToken: OnLedgerEntity.NonFungibleToken
```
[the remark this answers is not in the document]
CyonAlexRDX: TODO: revert `nftToken` naming

RadixWallet/Features/AssetTransferFeature/Components/TransferAccountList/ReceivingAccount/ReceivingAccount+Reducer.swift
```diff
@@ -76,65 +72,45 @@ public struct ReceivingAccount: Sendable, FeatureReducer {
 	}
 }
 
-extension ReceivingAccount.State.Account {
+extension AssetsTransfersRecipient {
```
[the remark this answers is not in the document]
CyonAlexRDX: Suggest a better name please :) this is inside Sargon, so must not use `.` in the name. (Rust does not work with `.` type names.).

It is currently only used for transfers. So I think the name is pretty good. And we can change it if we were to use it in other contexts? 

But you might convince me if you suggest an awesome name :) - since we cannot use `.` in Rust, the name cannout be `Account` for that is `Profile.Netork.Account`'s name in Sargon.

RadixWallet/Features/AssetTransferFeature/Components/TransferAccountList/ReceivingAccount/ReceivingAccount+Reducer.swift
```diff
@@ -76,65 +72,45 @@ public struct ReceivingAccount: Sendable, FeatureReducer {
...
 	}
 
 	var address: AccountAddress {
-		switch self {
-		case let .left(account):
-			account.address
-		case let .right(address):
-			address
-		}
+		accountAddress
```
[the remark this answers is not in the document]
CyonAlexRDX: TODO: remove `address`

RadixWallet/Features/AssetsFeature/Components/HelperViews/ResourceBalance/ResourceBalance+Helpers.swift
CyonAlexRDX: TODO: Upgrade this code since Decimal192 ARE Comparable :D

RadixWallet/Features/AssetsFeature/Components/HelperViews/ResourceBalance/ResourceBalanceView.swift
```diff
@@ -206,8 +206,8 @@ extension ResourceBalanceView {
 		public var body: some View {
 			NonFungibleView(
 				thumbnail: .nft(viewState.resourceImage),
-				caption1: viewState.resourceName ?? viewState.id.resourceAddress().formatted(),
-				caption2: viewState.nonFungibleName ?? viewState.id.localId().formatted(),
+				caption1: viewState.resourceName ?? viewState.id.resourceAddress.formatted(),
+				caption2: viewState.nonFungibleName ?? viewState.id.nonFungibleLocalId.formatted(),
```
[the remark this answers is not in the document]
CyonAlexRDX: that is the name inside of Sargon, I might change it later, but all those changes are breaking changes for Android so will need to be coordinated.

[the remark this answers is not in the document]
CyonAlexRDX: but I will add an "alias" - a computed property inside Swift Sargon:
```swift
var localID: NonFungibleLocalID { nonFungibleLocalId }
```
and we can use that...

RadixWallet/Core/SharedModels/Assets/OnLedgerEntity.swift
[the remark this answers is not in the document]
CyonAlexRDX: getting rid of both

RadixWallet/Clients/AccountPortfoliosClient/AccountPortfoliosClient+Live.swift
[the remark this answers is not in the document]
GhenadieVP: and only in DEBUG, to be able to test different states of the views with fiat values shown.

RadixWallet/Core/SharedModels/Assets/FiatWorth.swift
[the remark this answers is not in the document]
GhenadieVP: Sure semantically it is similar to Optional, I felt at the time that I needed to be more explicit with naming.

RadixWalletTests/Clients/TokenPriceCientTests/TokenPriceCientTests.swift
```diff
@@ -19,47 +20,45 @@ final class TokenPriceCientTests: XCTestCase {
...
 
 	func test_belowOne_2() {
-		validateDecimalPriceConversion(0.000099, expected: try! .init(value: "0.000099"))
+		validateDecimalPriceConversion(0.000099, expected: try! Decimal192("0.000099"))
 	}
 
-	// NOTE: All of the below values would be rounded to 14 decimal places
-	//       As it seems that Swift number formatter for Double cannot express more decimals.
-	func test_closeToRETDecimalDivisibility() {
+	func test_closeToDecimal192Divisibility() {
```
Sajjon, line 38: @GhenadieVP @kugel3 we just got two more decimal places of precison, 16 instead of 14, for these tests, Sargon is wise! :)

RadixWallet/Core/SharedModels/Assets/OnLedgerEntity.swift
kugel3: It does in theory, if you only support even hundreds, thousands etc of a coin, but we probably don't support that.

RadixWallet/Features/AssetTransferFeature/Components/TransferAccountList/ReceivingAccount/ReceivingAccount+Reducer.swift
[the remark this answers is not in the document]
kugel3: TransferAccount...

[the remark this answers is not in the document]
Sajjon: I went with `AccountOrAddressOf`:

```rust
```

 - we can always create a typealias for that in the context of Transaction in iOS if we want.

https://github.com/radixdlt/sargon/commit/d2c0166c0a5c499d634ef83d331315ef570694d9

RadixWallet/Features/AssetTransferFeature/AssetTransfer+Reducer.swift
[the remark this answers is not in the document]
Sajjon: `case profileAccount` and `addressOfExternalAccount` it is

RadixWallet/Features/AssetTransferFeature/Components/ChooseAccount/ChooseReceivingAccount+View.swift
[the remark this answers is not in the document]
Sajjon: `case profileAccount` and `addressOfExternalAccount` it is

PR #1041, opened 2024-03-11
NPS Survey

GhenadieVP:


RadixWallet/Clients/NPSSurveyClient/NPSSurveyClient+Live.swift
```diff
@@ -0,0 +1,151 @@
...
+extension NPSSurveyClient {
+	private static let feedbackTransactionCounterThreshold = 10
+	private static let nextFeedbackIntervalThresholdInMonths = 3
+
+	@Sendable
+	static func shouldAskUserForFeedback(_ transactionCounter: Int) -> Bool {
+		@Dependency(\.userDefaults) var userDefaults
+		@Dependency(\.date) var date
+
+		if transactionCounter > Self.feedbackTransactionCounterThreshold {
```
CyonAlexRDX, line 44: this is wrong right?

you have the same condition in both if and else if being `transactionCounter > Self.feedbackTransactionCounterThreshold`.

GhenadieVP: should be `==`, forgot to remove `>` as it was used for quick testing... good catch

RadixWallet/Clients/NPSSurveyClient/NPSSurveyClient+Live.swift
```diff
@@ -0,0 +1,151 @@
...
+
+private extension URLQueryItem {
+	static func userId(_ id: UUID) -> Self {
+		.init(name: "id", value: id.uuidString)
+	}
+
+	#if DEBUG
+	static let formUUID: Self = .init(name: "form_uuid", value: "281622a0-dc6b-11ee-8fd1-23c96056fbd2")
+	#else
+	static let formUUID: Self = .init(name: "form_uuid", value: "3432b6e0-dfad-11ee-a53c-95167f067d9c")
```
CyonAlexRDX, line 130: hmm what is this formUUID? something which is hardcoded for RDX Works, and safe to be open, since also an API secret need to be used ?

RadixWallet/Clients/NPSSurveyClient/NPSSurveyClient+Live.swift
```diff
@@ -0,0 +1,151 @@
...
+	#endif
+
+	static func npsScore(_ score: Int) -> Self {
+		.init(name: "nps", value: "\(score)")
+	}
+
+	static func npsFeedbackReason(_ reason: String) -> Self {
+		.init(name: "what_do_you_value_most_about_our_service", value: reason)
```
CyonAlexRDX, line 138: you can use `QueryItem.feedbackResponse.rawValue` here to avoid duplication, right?

RadixWallet/Clients/OverlayWindowClient/OverlayWindowClient+Live.swift
```diff
@@ -41,4 +41,5 @@ extension OverlayWindowClient.Item.HUD {
 	public static let updatedAccount = Self(text: L10n.AccountSettings.updatedAccountHUDMessage)
 	public static let copied = Self(text: L10n.AddressAction.copiedToClipboard)
 	public static let seedPhraseImported = Self(text: L10n.ImportMnemonic.seedPhraseImported)
+	public static let thankYou = Self(text: "Thank you!")
```
CyonAlexRDX, line 44: localize?

RadixWallet/Features/HomeFeature/Children/NPSSurvey/NPSSurvey+View.swift
```diff
@@ -0,0 +1,127 @@
...
+				}
+			}
+		}
+	}
+}
+
+extension NPSSurvey.View {
+	@ViewBuilder
+	private func headerView() -> some SwiftUI.View {
+		Text("How’s it Going?")
```
CyonAlexRDX, line 48: localize?

RadixWallet/Features/HomeFeature/Children/NPSSurvey/NPSSurvey+View.swift
```diff
@@ -0,0 +1,127 @@
...
+	private func scoreSelectionView() -> some SwiftUI.View {
+		VStack(alignment: .center, spacing: .small2) {
+			HStack {
+				ForEach(0 ..< 6) { score in
+					scoreButton(score)
+				}
+			}
+
+			HStack {
+				ForEach(6 ..< 11) { score in
```
CyonAlexRDX, line 70: I prefer `6 ... 10`, i.e. to not show out of bounds values (11).

CyonAlexRDX: but maybe use a constant for the NPS scale 0-10, and use it everywhere, seems best? i.e. here but also strings, i.e. "0 - Not Likeley" etc. (yes I know, it wont change..., but is semantically most correct)

RadixWallet/Features/HomeFeature/Coordinator/Home+View.swift
```diff
@@ -95,6 +95,9 @@ extension Home {
 			.onFirstAppear {
 				store.send(.view(.onFirstAppear))
 			}
+			.onAppear {
+				debugPrint("Home did Appear")
```
CyonAlexRDX, line 99: remove?

RadixWallet/Clients/NPSSurveyClient/NPSSurveyClient+Live.swift
GhenadieVP: yes, that form is guarded with the API Secret.

PR #1044, opened 2024-03-14
Avoid crashing for RETDecimal.max().formated(...)

CyonAlexRDX:
Jira ticket: paste link here

Complements / depends on / relates to #link to other PR

## Description
This PR solves / fixes / adds / includes functionality 1 and functionality 2.

### Notes
Additional notes.

1. Step 1
2. Step 2
3. Verify

1. Step 1
2. Step 2
3. Verify

## Screenshot

| - | - |

## Video
paste link here

## PR submission checklist

### DIFF
--- a/RadixWallet/EngineKit/RETDecimal+Extensions.swift
+++ b/RadixWallet/EngineKit/RETDecimal+Extensions.swift
@@ -246,7 +246,23 @@ extension RETDecimal {
 		let integerCount = UInt(Swift.max(digits.count - RETDecimal.scale, 1))
 		if integerCount > totalPlaces {
 			let scale = RETDecimal(exponent: integerCount - totalPlaces)
-			return (self / scale).rounded(decimalPlaces: 0) * scale
+			let base = self / scale
+			let decimalPlaces: UInt = 0
+			let base_rounded = base.rounded(decimalPlaces: decimalPlaces)
+
+			do {
+				return try base_rounded.mul(other: scale)
+			} catch {
+				// we overflowed which will happen if we are using RETDecimal.max()
+				// instead of using `base.rounded(decimalPlaces: 0)` which uses
+				// rounding mode `toNearestMidpointAwayFromZero` we will use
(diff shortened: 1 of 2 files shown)


RadixWalletTests/EngineKitTests/DecimalTests.swift
```diff
@@ -274,15 +274,17 @@ final class DecimalTests: TestCase {
...
 			let locale = Locale(identifier: "en_US_POSIX")
-			let decimal = try RETDecimal(value: decimalString)
 			let actual = decimal.formatted(locale: locale, totalPlaces: 8, useGroupingSeparator: false)
 			XCTAssertEqual(actual, expected, line: line)
 		}
+		func doTest(_ decimalString: String, expected: String, line: UInt = #line) throws {
+			try doTest(RETDecimal(value: decimalString), expected: expected, line: line)
+		}
 
+		try doTest(RETDecimal.max(), expected: "3.138e39")
```
Sajjon, line 286: this is a new test, which fails on `main` branch, but not anymore.

RadixWallet/EngineKit/RETDecimal+Extensions.swift
```diff
@@ -246,7 +246,23 @@ extension RETDecimal {
 		let integerCount = UInt(Swift.max(digits.count - RETDecimal.scale, 1))
 		if integerCount > totalPlaces {
 			let scale = RETDecimal(exponent: integerCount - totalPlaces)
-			return (self / scale).rounded(decimalPlaces: 0) * scale
+			let base = self / scale
```
kugel3, line 249: I would say that we don't want to protect ourselves against any kind of error here, only the specific case of max, no? We could just

```
guard self != max else {
  // return the number rounded towards zero
}
```

CyonAlexRDX: No because if the value is close to max we don't want to fail either. Hmm

PR #1045, opened 2024-03-14
TX history performance

GhenadieVP:
Pure SwiftUI implementation for TXHistory be it with using List or ScrollView+LazyVStack has very poor performance.

I did investigate many ways of improving, even using metal optimization with drawingGroup - nada.

But, good old TableView, is magical. The performance is uncomparable.

NOTE: SOME FUNCTIONALITY WAS REMOVED FOR NOW:
- The Accountheader is sticky, it is possible to make it scroll, but it would take more swiftUI work.
- Loading indicator needs to be added back.
- Needs proper event for section did dissapear.

Before:

https://github.com/radixdlt/babylon-wallet-ios/assets/118184705/84f97704-5ace-4f7a-ae78-979f67fc3ccc

After:

https://github.com/radixdlt/babylon-wallet-ios/assets/118184705/1f5e0d6b-11e1-4df1-98af-7a13d739d473


CyonAlexRDX (approved):
LGTM! That performance diff is...mind blowing.

GhenadieVP:
Obsolete, was directly added to the TX history

PR #1054, opened 2024-03-20
A bunch of bugs

GhenadieVP:
Several small bug fixes:
- Proper resources refresh on pull to refresh.
- Properly display the txID.
- Properly display the title of the Choose Asset in Asset Transfer.
- Add back TX message sent from Asset Transfer.


kugel3:
I have a better solution for the id, ResourceBalance shouldn't be ID:ed at all really, it's only in the TransactionReview context that we use that, and that's really only so that we can apply guarantees. I use `Identified<ResourceBalance>` there instead, and remove the id from ResourceBalance. Using the resource address is not enough, since there are situations where you can have lists where the same resource appears multiple times, etc. 

About the other bugs, what are they?

RadixWallet/Clients/AccountPortfoliosClient/AccountPortfoliosClient+Live.swift
GhenadieVP, line 1: applied the proper forcerRefresh.

RadixWallet/Features/AssetTransferFeature/AssetTransfer+Reducer.swift
```diff
@@ -77,7 +77,7 @@ public struct AssetTransfer: Sendable, FeatureReducer {
 				let manifest = try await createManifest(accounts)
 				Task {
 					_ = try await dappInteractionClient.addWalletInteraction(
-						.transaction(.init(send: .init(transactionManifest: manifest))),
+						.transaction(.init(send: .init(transactionManifest: manifest, message: message))),
```
GhenadieVP, line 80: Message was removed in some of older PRs

RadixWallet/Features/AssetTransferFeature/Components/TransferAccountList/TransferAccountList+View.swift
```diff
@@ -90,11 +90,11 @@ private extension View {
 	private func addAsset(with destinationStore: PresentationStoreOf<TransferAccountList.Destination>) -> some View {
 		sheet(store: destinationStore.scope(state: \.state.addAsset, action: \.addAsset)) { assetsStore in
 			AssetsView.View(store: assetsStore)
+				.navigationTitle(L10n.AssetTransfer.AddAssets.navigationTitle)
```
GhenadieVP, line 93: With latest iOS updates the order matters now.

kugel3 (approved):

PR #1068, opened 2024-04-02
[ABW-3015] AccountDetails - Collapse header on scroll

danvleju-rdx:
Jira ticket: [ABW-3015](https://radixdlt.atlassian.net/browse/ABW-3015)

## Description
Adds a generic component that creates a collapsible header with a list, used in the Account Details screen.

## Demo
https://github.com/radixdlt/babylon-wallet-ios/assets/163979791/700e9806-48ce-42da-8bb9-98c3215e6908

[ABW-3015]: https://radixdlt.atlassian.net/browse/ABW-3015?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


CyonAlexRDX:
should we really hide the bar "Tokens" / "NFT" / "Staking" / "Pool" here? I would I expect it to be sticky, but maybe Matt did not want that?

Screenshot 2024-04-02 at 15 03 06

RadixWallet/Features/AccountDetailsFeature/HeaderListViewContainer.swift
```diff
@@ -0,0 +1,119 @@
...
+	public func makeUIView(context: Context) -> UIView {
+		let containerView = UIView()
+		addSubviews(in: containerView)
+
+		return containerView
+	}
+
+	public func updateUIView(_ uiView: UIViewType, context: Context) {
+		uiView.subviews.forEach { $0.removeFromSuperview() }
+		addSubviews(in: uiView)
```
matiasbzurovski, line 42: do we need to remove and add every subview each time? It is adding a small glitch every time the view appears/disappears (it shows the activity indicator for an instant).


https://github.com/radixdlt/babylon-wallet-ios/assets/164921079/8649d9b9-4f3d-40eb-9a8f-2c15af5aff69


Screenshot 2024-04-02 at 16 27 46

matiasbzurovski (changes requested):

kugel3 (changes requested):
This looks very nice, but we do have as a goal to remove all usage of Introspect. Could this not be done instead with an GeometryReader/Preference combo until we can require iOS 17? We already have some helpers defined for that.

If you place it on the upmost view in the scrollview, you will be able to get the content offset as long the it is (more or less) visible, but we also know that when it disappears, the header should also be collapsed.

Another option we should consider is to use a UITableView here, we do that in AccountHistory, mostly for performance reasons, but it also makes things like this easier.

GhenadieVP:
Looks slick, though there are some rough edges, which I am not sure easy are fixable. Here I have few resources in my account, so the scroll content is reduced.
Additionally, some deceleration would be nice when user stops scrolling, but probably I am getting to fancy :)


https://github.com/radixdlt/babylon-wallet-ios/assets/118184705/86a3b176-7db4-49a7-b789-25c8a3a2baea

danvleju-rdx:
@GhenadieVP Reasonably fancy :). I did some improvements [here](https://github.com/radixdlt/babylon-wallet-ios/pull/1068/commits/58074b8de2d18bcd5910c4c552828e788ae3aa19). Feels much better.

https://github.com/radixdlt/babylon-wallet-ios/assets/163979791/7b575475-2cb9-4a7a-b486-2c5d31d9d01e

matiasbzurovski (approved):

kugel3 (approved):

PR #1070, opened 2024-04-03
1.5.0 Hot fixes

GhenadieVP:
1.5.0 Hot fixes

- Properly truncate price before initialising the RETDecimal.
- Apply empty vaults when refreshing the portfolio for a given account.
- TXID truncation.

FLOOP worth:
Screenshot 2024-04-03 at 09 07 07

Tokens with near zero value are still shown
Screenshot 2024-04-03 at 09 07 43

TXID truncation
Screenshot 2024-04-03

### DIFF
--- a/RadixWallet/Clients/AccountPortfoliosClient/AccountPortfoliosClient+Live.swift
+++ b/RadixWallet/Clients/AccountPortfoliosClient/AccountPortfoliosClient+Live.swift
@@ -81,7 +81,7 @@ extension AccountPortfoliosClient: DependencyKey {
 			await state.setSelectedCurrency(preferences.fiatCurrencyPriceTarget)
 			await state.setIsCurrencyAmountVisble(preferences.isCurrencyAmountVisible)
 
-			let accounts = try await onLedgerEntitiesClient.getAccounts(accountAddresses).map(\.nonEmptyVaults)
+			let accounts = try await onLedgerEntitiesClient.getAccounts(accountAddresses)
 
 			let portfolios = accounts.map { AccountPortfolio(account: $0) }
 			await state.handlePortfoliosUpdate(portfolios)
@@ -120,7 +120,7 @@ extension AccountPortfoliosClient: DependencyKey {
 			await applyTokenPrices(Array(allResources), forceRefresh: forceRefresh)
 
 			// Load additional details
-			_ = await accounts.parallelMap {
+			_ = await accounts.map(\.nonEmptyVaults).parallelMap {
 				await fetchPoolAndStakeUnitsDetails($0, cachingStrategy: forceRefresh ? .forceUpdate : .useCache)
 			}
 
@@ -133,7 +133,7 @@ extension AccountPortfoliosClient: DependencyKey {
 				cacheClient.removeFolder(.onLedgerEntity(.account(accountAddress.asGeneral)))
 			}
 
-			let account = try await onLedgerEntitiesClient.getAccount(accountAddress).nonEmptyVaults
+			let account = try await onLedgerEntitiesClient.getAccount(accountAddress)
 			let portfolio = AccountPortfolio(account: account)
 
 			let currentResources = await state.tokenPrices.keys
@@ -143,7 +143,7 @@ extension AccountPortfoliosClient: DependencyKey {
 			)
 
 			await state.handlePortfolioUpdate(portfolio)
-			await fetchPoolAndStakeUnitsDetails(account, cachingStrategy: forceRefresh ? .forceUpdate : .useCache)
+			await fetchPoolAndStakeUnitsDetails(account.nonEmptyVaults, cachingStrategy: forceRefresh ? .forceUpdate : .useCache)
 
 			return portfolio
 		}
--- a/RadixWallet/Clients/AccountPortfoliosClient/AccountPortfoliosClient+Live.swift
+++ b/RadixWallet/Clients/AccountPortfoliosClient/AccountPortfoliosClient+Live.swift
@@ -133,7 +133,7 @@ extension AccountPortfoliosClient: DependencyKey {
 				cacheClient.removeFolder(.onLedgerEntity(.account(accountAddress.asGeneral)))
 			}
 
-			let account = try await onLedgerEntitiesClient.getAccount(accountAddress)
+			let account = try await onLedgerEntitiesClient.getAccount(accountAddress).nonEmptyVaults
 			let portfolio = AccountPortfolio(account: account)
 
 			let currentResources = await state.tokenPrices.keys
--- a/RadixWallet/Clients/TokenPriceClient/TokenPriceClient+Live.swift
+++ b/RadixWallet/Clients/TokenPriceClient/TokenPriceClient+Live.swift
@@ -45,10 +45,12 @@ extension TokenPricesClient {
 extension TokenPricesClient.TokenPrices {
 	fileprivate init(_ tokenPricesResponse: TokensPriceResponse) {
 		let formatter = NumberFormatter()
+		formatter.locale = Locale(identifier: "en_US_POSIX") // Just ignore the users locale
 		formatter.numberStyle = .decimal
 		formatter.maximumFractionDigits = Int(RETDecimal.maxDivisibility)
 		formatter.roundingMode = .down
-		formatter.decimalSeparator = "." // Enfore dot notation for RETDecimal
+		formatter.decimalSeparator = "." // Enforce dot notation for RETDecimal
+		formatter.usesGroupingSeparator = false // No grouping separator for RETDecimal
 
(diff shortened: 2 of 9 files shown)


RadixWallet/Clients/AccountPortfoliosClient/AccountPortfoliosClient+Live.swift
```diff
@@ -133,7 +133,7 @@ extension AccountPortfoliosClient: DependencyKey {
 				cacheClient.removeFolder(.onLedgerEntity(.account(accountAddress.asGeneral)))
 			}
 
-			let account = try await onLedgerEntitiesClient.getAccount(accountAddress)
+			let account = try await onLedgerEntitiesClient.getAccount(accountAddress).nonEmptyVaults
```
GhenadieVP, line 136: This was causing users to see tokens with zero values when they did performa pull to refresh in Account details

RadixWallet/Clients/TokenPriceClient/TokenPriceClient+Live.swift
```diff
@@ -4,11 +4,8 @@ extension TokenPricesClient {
 		@Dependency(\.jsonDecoder) var jsonDecoder
 		@Dependency(\.jsonEncoder) var jsonEncoder
 		@Dependency(\.cacheClient) var cacheClient
-		#if DEBUG
```
GhenadieVP, line 7: Opt to always use prod token price service to facilitate testing. Dev endpoint might not have all prices available or latest ones.

RadixWallet/Clients/TokenPriceClient/TokenPriceClient+Live.swift
```diff
@@ -45,10 +42,12 @@ extension TokenPricesClient {
 extension TokenPricesClient.TokenPrices {
 	fileprivate init(_ tokenPricesResponse: TokensPriceResponse) {
 		let formatter = NumberFormatter()
+		formatter.locale = Locale(identifier: "en_US_POSIX") // Just ignore the users locale
 		formatter.numberStyle = .decimal
 		formatter.maximumFractionDigits = Int(RETDecimal.maxDivisibility)
 		formatter.roundingMode = .down
-		formatter.decimalSeparator = "." // Enfore dot notation for RETDecimal
+		formatter.decimalSeparator = "." // Enforce dot notation for RETDecimal
+		formatter.usesGroupingSeparator = false // No grouping separator for RETDecimal
```
GhenadieVP, line 50: Not having this caused the prices > 1000 to be formatted with a thousand separator which then couldn't be converted to RETDecimal.

RadixWallet/Clients/TokenPriceClient/TokenPriceClient+Live.swift
```diff
@@ -45,10 +42,12 @@ extension TokenPricesClient {
 extension TokenPricesClient.TokenPrices {
 	fileprivate init(_ tokenPricesResponse: TokensPriceResponse) {
 		let formatter = NumberFormatter()
```
GhenadieVP, line 44: Would be nice to have a RETDecimal initializer in Sargon to handle the case of providing a Double with more than 18 decimal places.

RadixWallet/Clients/AccountPortfoliosClient/AccountPortfoliosClient+Live.swift
kugel3: If this is done here, will that not affect all the endpoints on the client? In transaction history I need the full portfolio, including empty vaults.

RadixWallet/Clients/TokenPriceClient/TokenPriceClient+Live.swift
Sajjon: yeah I can fix one!

RadixWallet/Clients/TokenPriceClient/TokenPriceClient+Live.swift
Sajjon: there is another initialiser than `value: String`, there is `formatted: String` as well. which should be able to accept strings with a thousands separator.

RadixWallet/Core/FeaturePrelude/AddressView/AddressFormat.swift
```diff
@@ -74,3 +74,16 @@ extension NonFungibleLocalId {
 		}
 	}
 }
+
+extension TXID {
```
Sajjon, line 78: Aha! I suspected we had a regression, because I started adding this in Sargon for transaction hashes but then tested main branch and saw we did not truncated. I should have spoken up :) @micbakos-rdx we can you add a PR implementing this in Sargon?

RadixWalletTests/Clients/TokenPriceCientTests/TokenPriceCientTests.swift
```diff
@@ -0,0 +1,82 @@
+@testable import Radix_Wallet_Dev
+import XCTest
+
+final class TokenPriceCientTests: XCTestCase {
```
Sajjon, line 4: tests! 🙂

RadixWallet/Clients/TokenPriceClient/TokenPriceClient+Live.swift
kugel3: It looks like we are rounding the Double that represents the price, by first converting the price Double to a string (using a fixed decimal separator) and then converting it back to RETDecimal? 

1. I presume that `RETDecimal(value: String)` is the version that takes a machine readable string, so that it ignores the user preference on decimal separator, because otherwise this wouldn't work, but even so, I don't like this, it doesn't seem nice or necessary to convert the price to a Double, then a String and the RETDecimal, just in order to round it.

2. If we do want to round it here, can't we convert it (back) to RETDecimal and just round that, without going by way of String?

3. Why do we want `price` to be a Double at all? This operation here would be even simpler if it was kept a RETDecimal until it's to be displayed.

4. Is the behaviour actually correct? What is the purpose of clamping the number of decimals in the **price** according to the divisibility, isn't the price here given in a fiat currency?

RadixWallet/EngineKit/TXID.swift
```diff
@@ -3,10 +3,6 @@ import EngineToolkit
 public typealias TXID = TransactionHash
 
 extension TXID {
-	public func formatted(_ format: AddressFormat = .default) -> String {
```
Sajjon, line 6: @GhenadieVP merge conflict imminent, but I see, it must happen, since you needed to implement this a method with this signature.

RadixWallet/Clients/AccountPortfoliosClient/AccountPortfoliosClient+Live.swift
Sajjon: right should be able to do it later, in view layer is Gustafs meaning?

GhenadieVP: probably will it will,
will move this at UI level

RadixWallet/Clients/TokenPriceClient/TokenPriceClient+Live.swift
kugel3: >there is another initialiser than value: String, there is formatted: String as well. which should be able to accept strings with a thousands separator.

Presumably (hopefully), the latter respects the user's locale, and the former does not, and in this case we don't want to respect the locale, since we hard code the decimal separator.

GhenadieVP: @kugel3 note this whole shenanigans is not about rounding, it is about truncating to RETDecimal max divisibility, otherwise we cannot initialize the RETDecimal with a double that has more decimals than 18.

The price is double as it comes from the TokenPrice service, here we do immediately convert it to RETDecimal.

kugel3: Alright, then I understand your other comment about an initialiser for RETDecimal that takes a Double with arbitrary number of decimals, that would be highly preferable. Since NumberFormatter is pretty expensive to initialise, we probably don't want to make that initialiser in Swift.

But while we have this workaround, I think we should be more explicit about what is going on, so that nobody accidentally changes to using the other string based initialiser for example. Because what I wrote before is correct, right? The `RETDecimal(formatted: String)` could not actually be used here?

Perhaps change `// Enforce dot notation for RETDecimal` to something like `// Use a period as decimal separator, for the sake of RETDecimal(formatted: String)`.

CyonAlexRDX: Yes `RETDecimal.init(formattedStrinf:locale)` is respects locale

PR #1073, opened 2024-04-04
[ABW-2735] AppsFlyer integration

matiasbzurovski:
Jira ticket: [ABW-2735](https://radixdlt.atlassian.net/jira/software/c/projects/ABW/issues/ABW-2735)

## Description
Adds AppsFlyer support.

### Notes
The AppsFlyer logic is very simple, but this PR also includes a solution for handling sensitive information (such as AppsFlyer tokens). To do so, the app has a new `SensitiveInfoClient` that will fetch the values for each key specified on its interface.

The file from which the app will read its values is called `SensitiveInfo.plist`. Given its content is sensitive, it has been added to the `.gitgnore` so that its content is never committed. However, a sample file (with placeholder values) has been added, and a custom run script phase will make sure to create the `SensitiveInfo` file from the sample one if it doesn't exist.

Then, on CD, we will replace the placeholder values with the real ones that are safely stored on secrets.

[ABW-2735]: https://radixdlt.atlassian.net/browse/ABW-2735?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.gitignore
```diff
@@ -13,7 +13,8 @@ DerivedData/
 .netrc
 fastlane/report.xml
 fastlane/test_output/
-fastlane/.env.secret
+.env.secret
```
matiasbzurovski, line 16: This ignores both `.env.secret` files: the one at root level and the existing one under `fastlane`.

RadixWallet/Clients/SensitiveInfoClient/SensitiveInfoClient+Live.swift
```diff
@@ -0,0 +1,19 @@
+extension SensitiveInfoClient: DependencyKey {
+	static var liveValue: SensitiveInfoClient {
+		.init(
+			read: { key in
+				guard let filePath = Bundle.main.path(forResource: "SensitiveInfo", ofType: "plist") else {
+					fatalError("Couldn't find file 'SensitiveInfo.plist'")
+				}
+				let plist = NSDictionary(contentsOfFile: filePath)
+				guard let value = plist?.object(forKey: key.rawValue) as? String else {
+					fatalError("Couldn't find value for key '\(key.rawValue)' in 'SensitiveInfo.plist'.")
```
GhenadieVP, line 10: `fatalError` is intentional?

RadixWallet/Clients/SensitiveInfoClient/SensitiveInfoClient+Live.swift
```diff
@@ -0,0 +1,19 @@
+extension SensitiveInfoClient: DependencyKey {
+	static var liveValue: SensitiveInfoClient {
+		.init(
+			read: { key in
+				guard let filePath = Bundle.main.path(forResource: "SensitiveInfo", ofType: "plist") else {
+					fatalError("Couldn't find file 'SensitiveInfo.plist'")
```
GhenadieVP, line 6: Same here I guess, though it is 99.99% percent guaranteed that the file exists. In general we should not break the app if we cannot configure the AppsFlyer SDK.

RadixWallet.xcodeproj/project.pbxproj
```diff
@@ -7850,6 +7887,28 @@
 		};
 /* End PBXResourcesBuildPhase section */
 
+/* Begin PBXShellScriptBuildPhase section */
```
GhenadieVP, line 7890: Do we still want this?

RadixWallet/Clients/AppsFlyerClient/AppsFlyerClient+Live.swift
```diff
@@ -0,0 +1,20 @@
+import AppsFlyerLib
+
+extension AppsFlyerClient: DependencyKey {
+	static var liveValue: AppsFlyerClient {
+		@Dependency(\.sensitiveInfoClient) var sensitiveInfoClient
+
+		return .init(
+			start: {
+				AppsFlyerLib.shared().appsFlyerDevKey = sensitiveInfoClient.read(.appsFlyerDevKey)
```
GhenadieVP, line 9: We should probably not start the AppsFlyer if the necessary sensitive info is missing.
Applicable for CI, where we don't care to start and run AppsFlyer.

RadixWallet/Clients/SensitiveInfoClient/SensitiveInfoClient+Live.swift
matiasbzurovski: Will change to `assertionFailure` instead

RadixWallet.xcodeproj/project.pbxproj
matiasbzurovski: Yes, otherwise developers outside RDX Works team won't have the placeholder `SensitiveInfo.plist` created (as they won't run fastlane).

GhenadieVP: why wouldn't they run the fastlane?

matiasbzurovski: well, until now fastlane only makes sense if you have access to secrets. We can remove this script phase and update development set up so that they always run the `create_sensitive_info` lane though. wdyt?

GhenadieVP: well, the respective lane that you have added those not require secrets, unless I am missing something. Those are needed when the app has to be built.

fastlane/Fastfile
```diff
@@ -29,9 +30,33 @@ platform :ios do
...
+        if File.exists?("../Aux/SensitiveInfo.plist")
+            puts "SensitiveInfo.plist already available"
+        elsif
+            sh("cp -v '../Aux/SensitiveInfo-Sample.plist' '../Aux/SensitiveInfo.plist'")
+        end
+    end
+
+    desc "Populates SensitiveInfo.plist with environment values"
+    lane :populate_sensitive_info do |options|
+        appIdKey = "APPS_FLYER_APP_ID_%s" % options[:flavour]
```
GhenadieVP, line 44: What happens if the flavour is Pre-Alpha or Beta?

GhenadieVP: I guess you can find out by running Pre-Alpha build from your branch :)

matiasbzurovski: It leaves an empty value on the `APPS_FLYER_APP_ID`, which is fine since we don't care about AppsFlyer for those builds.

I can update the `AppsFlyerClient` to use `nilIfEmpty` on each key, but tbh I don't think there is any issue with starting the AppsFlyer SDK with invalid keys.

GhenadieVP (approved):

PR #1076, opened 2024-04-08
Describing use of required reason API

GhenadieVP:
## Description
We required to implement the [Describing use of required reason API](https://developer.apple.com/documentation/bundleresources/privacy_manifest_files/describing_use_of_required_reason_api)

From the documentation and our codebase, the only thing that requires to declare the reason is UserDefaults.
Even though the Apple email suggested that we do also need to provide the description for the following:
- [File timestamp APIs](https://developer.apple.com/documentation/bundleresources/privacy_manifest_files/describing_use_of_required_reason_api#4278393)
- [System boot time APIs](https://developer.apple.com/documentation/bundleresources/privacy_manifest_files/describing_use_of_required_reason_api#4278394)
- [Disk space APIs](https://developer.apple.com/documentation/bundleresources/privacy_manifest_files/describing_use_of_required_reason_api#4278397)

I didn't find exact usage for these in our codebase, and I am not sure how the analysis is made by Apple. Possible our third-party dependencies might use these, need to investigate them. I guess the only way to find out is to send the app again for review.

### DIFF
--- a/Aux/PrivacyInfo.xcprivacy
+++ b/Aux/PrivacyInfo.xcprivacy
@@ -1,0 +1,17 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<dict>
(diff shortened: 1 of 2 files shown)


CyonAlexRDX (approved):
LGTM

PR #1083, opened 2024-04-11
Fix 400 error in transaction history

kugel3:
Slack discussion: [thread](https://rdxworks.slack.com/archives/C05V4202J90/p1712242660583689)

## Description
Some users have experienced an HTTP error, that may be due to them having their clock going a little fast. When we make the requests in transaction history, the Gateway will complain if the upper bound date is more than a few seconds into the future.

This PR removes the upper bound when it would have otherwise been *now*. 

I tested setting the clock forward and on main that does cause the 400 error, and it is indeed fixed by this PR.

It also improves the handling when there are no transactions at all.

## PR submission checklist

### DIFF
--- a/RadixWallet/Features/TransactionHistory/TransactionHistory+Reducer.swift
+++ b/RadixWallet/Features/TransactionHistory/TransactionHistory+Reducer.swift
@@ -207,12 +207,10 @@ public struct TransactionHistory: Sendable, FeatureReducer {
 
 	public func reduce(into state: inout State, internalAction: InternalAction) -> Effect<Action> {
 		switch internalAction {
-		case let .loadedFirstTransactionDate(date):
+		case let .loadedFirstTransactionDate(firstDate):
 			state.loading.isLoading = false
-			let lastDate: Date = .now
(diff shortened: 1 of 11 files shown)


RadixWallet/Features/TransactionHistory/TransactionHistory+Reducer.swift
```diff
@@ -209,8 +209,8 @@ public struct TransactionHistory: Sendable, FeatureReducer {
 		switch internalAction {
 		case let .loadedFirstTransactionDate(date):
 			state.loading.isLoading = false
-			let lastDate: Date = .init(timeIntervalSinceNow: -10)
-			let firstDate = date ?? lastDate
+			let lastDate: Date = .now
+			let firstDate = date ?? lastDate.addingTimeInterval(-24 * 3600)
```
kugel3, line 213: This is just an arbitrary amount of time to get a non-empty time window when you have no transactions

CyonAlexRDX (approved):
LGTM!

GhenadieVP (approved):

PR #1084, opened 2024-04-11
custom-dump version bump

GhenadieVP:
swift-custom-dump is transient dependency, had to pin in Package.resolved.
Fixes the build error in Release mode - https://github.com/pointfreeco/swift-custom-dump/pull/108


danvleju-rdx (approved):

matiasbzurovski (commented):
As I don't see any changes in the `project.pbxproj`, I assume you manually changed the `Package.resolved` file.

This should work for the next archive, but if we ever resolve packages/add a new package, we will introduce this issue again. I think we should manually add `swift-custom-dump` as a dependency of our project and point it to this exact version.

CyonAlexRDX (approved):

PR #1090, opened 2024-04-17
Bump Sargon to version which works under RELEASE.

CyonAlexRDX:
Bump Sargon to a version which builds for `RELEASE`.

### DIFF
--- a/RadixWallet/Core/SargonExtensions/IntoSargon+Bridge.swift
+++ b/RadixWallet/Core/SargonExtensions/IntoSargon+Bridge.swift
@@ -166,8 +166,8 @@ extension ThirdPartyDeposits {
 }
 
 extension OrderedSet<EntityFlag> {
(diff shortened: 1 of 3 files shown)


RadixWallet/Core/SargonExtensions/IntoSargon+Bridge.swift
```diff
@@ -166,8 +166,8 @@ extension ThirdPartyDeposits {
 }
 
 extension OrderedSet<EntityFlag> {
-	func intoSargon() -> [Sargon.EntityFlag] {
-		map { $0.intoSargon() }
+	func intoSargon() -> Sargon.EntityFlags {
```
CyonAlexRDX, line 169: N.B. not really used... only used to be able to used `AccountOrAddressOf` enum and pass `Account` to Sargon, for AssetsTransfers, but the entity flags are not being read/ used by that....

GhenadieVP (approved):

PR #1093, opened 2024-04-18
Issue with entering and validating a 3 letter word

GhenadieVP:
Words strip is broken(not showing) in latest iOS versions, so we don't count on it to be used by users selecting the word disambiguate between the words. Rather the Wallet will validate the word eagerly if it is a valid one. 

Previously the import seed phrase screen would wait for the user to select a word from the words strip to disambiguate between the words. And since the strip was not shown, the word was never validated. For example when users entered `end` there were two matches `end` and `endless`.

Behaviour:
- User enters the first two characters, having the word `en` - the word is incomplete.
- User enters another character, having the word `end` - the word is valid.
- User enters additional character, having the word `endl` - the word is incomplete.
- User enters additional characters, having the word `endless` - the word is considered valid.
- User removes some characters, having the word `endle` - the word is incomplete.
- User removes more characters, having the word `end` - the word is valid.
- User removes another character, having the word `en` - the word is incomplete.

https://github.com/radixdlt/babylon-wallet-ios/assets/118184705/546e6c13-dc72-45d4-967f-1ed3b3c48d8a


RadixWallet/Features/ImportMnemonic/ImportMnemonic.swift
```diff
@@ -660,14 +660,32 @@ extension ImportMnemonic {
...
-			state.words[id: id]?.value = .incomplete(text: input, hasFailedValidation: false)
-			return .none
+			guard let userInput = candidates.first(where: { $0.word == nonEmptyInput }) else {
+				state.words[id: id]?.value = .incomplete(text: input, hasFailedValidation: false)
+				return .none
+			}
 
-		case let .known(.unambiguous(word, match, _)):
-			return completeWith(word: word, completion: .auto(match: match), id: id, input: input, &state)
+			return completeWith(word: userInput, completion: .user, id: id, input: input, &state)
```
Sajjon, line 681: I think `completion: .auto(match: match)` is more accurate than `.user`.

RadixWallet/Features/ImportMnemonic/ImportMnemonic.swift
```diff
@@ -660,14 +660,32 @@ extension ImportMnemonic {
...
+
+		case let .known(.unambiguous(word, _, _)):
+			guard word.word.rawValue == input else {
+				state.words[id: id]?.value = .incomplete(text: input, hasFailedValidation: false)
+				return .none
+			}
+			return completeWith(word: word, completion: .user, id: id, input: input, &state)
```
Sajjon, line 688: probably `completion: .auto(match: match)`

RadixWallet/Features/ImportMnemonic/ImportWord/ImportMnemonicWord.swift
```diff
@@ -125,29 +125,30 @@ public struct ImportMnemonicWord: Sendable, FeatureReducer {
...
-					// allow lookup if current word is invalid
-					return .send(.delegate(.lookupWord(input: input)))
-				default:
-					// We don't perform lookup when we decrease character count if the
-					// state is not currently "incomplete(_, hasFailedValidation: true)
-					state.value = .incomplete(text: input, hasFailedValidation: false)
-					return .none
-				}
-			}
+			// FIXME: - No autocompletion, so this is disabled for now.
```
Sajjon, line 128: 😢 too it broke with some IOS updates and because of our shaky navigation depth. A lot of users - including myself - loved the autocompletion. Lets try to bring it back later once we have time to build it properly!

CyonAlexRDX (approved):

RadixWallet/Features/ImportMnemonic/ImportMnemonic.swift
GhenadieVP: not so sure, maybe you know better, but doesn't `auto` mean that the word was autocompleted? say users entered `endle` and we do autocomplete with `endless`? 
If so, note that my fix removes autocompletion(intent is to remove the clever part and make it plain simple), so user needs to enter the `endless` word by themselves, in which case the completion is `user`.

CyonAlexRDX: `user` meant that user presses any word in the toolbar. But yeah we can user if user inputs all chars manually, that it is user... sure

PR #1096, opened 2024-04-18
[ABW-3164/3168/3174/3180] Redesign of App Settings

matiasbzurovski:
Jira tickets: 
- [ABW-3164](https://radixdlt.atlassian.net/browse/ABW-3164)
- [ABW-3168](https://radixdlt.atlassian.net/browse/ABW-3168)
- [ABW-3174](https://radixdlt.atlassian.net/browse/ABW-3174)
- [ABW-3180](https://radixdlt.atlassian.net/browse/ABW-3180)

## Description
This PR implements the redesign of the App Settings.

### Notes
One of the components added in this PR are the `Security Factors`. This view is going to be shown from the `Security Center` that @kugel3 is working on. Also, after his work, we will be able to reuse its logic to show the warnings (if any) on the `Settings` row, as shown [here](https://app.zeplin.io/project/639c7e7ab391bb15ac5d25fa/screen/66210084a17f5c375302a6c3). I have already implemented the changes to allow multiple hints per row.

## Video
In the following video you will notice that when tapping on `Security Center`, the `Security Factors` are shown. This was only added for the video to show how the view looks.

https://github.com/radixdlt/babylon-wallet-ios/assets/164921079/d7c68c53-1ad2-4120-acca-450c6a473cde

[ABW-3164]: https://radixdlt.atlassian.net/browse/ABW-3164?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[ABW-3168]: https://radixdlt.atlassian.net/browse/ABW-3168?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[ABW-3174]: https://radixdlt.atlassian.net/browse/ABW-3174?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[ABW-3180]: https://radixdlt.atlassian.net/browse/ABW-3180?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


RadixWallet/Clients/DeviceFactorSourceClient/DeviceFactorSourceClient+Live.swift
```diff
@@ -72,6 +72,14 @@ extension DeviceFactorSourceClient: DependencyKey {
 			)
 		}
 
+		let isSeedPhraseNeededToRecoverAccounts: @Sendable () async throws -> Bool = {
+			let deviceFactorSources = try await factorSourcesClient.getFactorSources(type: DeviceFactorSource.self)
+			let entities = try await deviceFactorSources.asyncMap {
+				try await entitiesControlledByFactorSource($0, nil)
+			}
+			return entities.contains(where: { !$0.isMnemonicPresentInKeychain })
```
matiasbzurovski, line 80: I don't know if this is the most optimal way to check this.

RadixWallet/Core/DesignSystem/Components/AssetIcon.swift
```diff
@@ -27,6 +27,7 @@ public struct AssetIcon: View {
 	public var body: some View {
 		image
 			.resizable()
+			.scaledToFit()
```
matiasbzurovski, line 30: Reason why system images were looking bad on Debug Settings 🙂

RadixWallet/Core/DesignSystem/Components/Hint.swift
```diff
@@ -73,9 +82,7 @@ public struct Hint: View, Equatable {
 		switch viewState.kind {
 		case .info:
 			nil
-		case .error:
-			AssetResource.error
-		case .warning:
+		case .error, .warning:
```
matiasbzurovski, line 85: `AssetResource.error` is 17x15 (not squared), which doesn't look great on Hints. Given we are always rendering the image as template and colorizing it, makes more sense to use `AssetResource.warningError` (which is squared).

RadixWallet/Core/DesignSystem/Components/Hint.swift
```diff
@@ -1,9 +1,16 @@
 
 // MARK: - Hint
 public struct Hint: View, Equatable {
-	public struct ViewState: Equatable {
+	public struct ViewState: Equatable, Identifiable {
 		public let kind: Kind
 		public let text: Text?
+		public let id: UUID
```
kugel3, line 7: In general this is pretty risky, you need to be very careful so that you only create these once when using UUID as ID

RadixWallet/Core/DesignSystem/Components/AssetIcon.swift
matiasbzurovski: | Before | After |
| -- | -- |
| IMG_0015 | IMG_0014 |

RadixWallet/Core/DesignSystem/Components/Hint.swift
matiasbzurovski: Forgot adding a comment on this class, but it was related to this in particular so will answer you here.

We are using `Hint` all over the app as a `View` that gets sent by parameter. This isn't very optimal as it is hard to compare, iterate, etc. Similar thing happens with the `let text: Text?` that the `Hint.ViewState` has.

I would recommend moving into a `ViewState` that instead has a `let text: String?`. Then, it is the Hint's `var body: some View` responsibility to create the `Text` from the `String`.

Same approach for the `Hint`: instead of sending a `Hint`, send a `Hint.ViewState` (example on `AppTextField`).

I didn't want to get into this change because it involves quite a lot of classes unrelated to this PR (which is big enough already). But if you like the approach I am happy to work on it.

RadixWallet/Core/DesignSystem/Components/PlainListRow.swift
```diff
@@ -49,45 +53,82 @@ public struct PlainListRow<Icon: View>: View {
...
+			icon
+				.padding(.trailing, .medium3)
+		}
+	}
+
+	@ViewBuilder
+	private var accessory: some View {
+		if let accessory = viewState.accessory {
+			Image(asset: accessory)
+		}
```
kugel3, line 110: I would prefer the inner variable not having the same name as the outer

RadixWallet/Core/DesignSystem/Components/PlainListRow.swift
```diff
@@ -109,27 +150,86 @@ public struct PlainListRowCore: View {
...
+			.zero
+		case .settings:
+			.medium1
+		}
+	}
+}
+
+// MARK: - PlainListRowCore.ViewState.Kind
+extension PlainListRowCore.ViewState {
+	public enum Kind {
```
kugel3, line 227: Perhaps Context or Mode make more sense as names for this?

RadixWallet/Core/DesignSystem/Styles/SecondaryRectangularButtonStyle.swift
```diff
@@ -1,6 +1,7 @@
 // MARK: - SecondaryRectangularButtonStyle
 public struct SecondaryRectangularButtonStyle: ButtonStyle {
 	@Environment(\.controlState) var controlState
+	let backgroundColor: Color
```
kugel3, line 4: I'd say we should avoid this, in my mind the background color is a core feature of this particular style. What other color do you need for this button style?

RadixWallet/Features/SettingsFeature/Preferences/Preferences+View.swift
```diff
@@ -0,0 +1,182 @@
...
+		.withSeparator
+		.eraseToAnyView()
+	}
+}
+
+private extension StoreOf<Preferences> {
+	var destination: PresentationStoreOf<Preferences.Destination> {
+		func scopeState(state: State) -> PresentationState<Preferences.Destination.State> {
+			state.$destination
+		}
```
kugel3, line 142: Just FYI: The only reason we use this pattern is because XCode get incredibly slow when it has to typecheck it otherwise, to the point of being unusable.

RadixWallet/Features/SettingsFeature/Preferences/Preferences+View.swift
```diff
@@ -0,0 +1,182 @@
+private typealias S = L10n.Preferences
```
kugel3, line 1: I don't like this, though I think Alex does. Since you don't seem to be using these strings in single row contexts anyway, there is no space constraint, better to use the full version, saves some cognitive load.

RadixWallet/Features/SettingsFeature/Preferences/Preferences+View.swift
```diff
@@ -0,0 +1,182 @@
...
+}
+
+extension Preferences.View {
+	@MainActor
+	private var content: some View {
+		WithViewStore(store, observe: \.viewState, send: { .view($0) }) { viewStore in
+			ScrollView {
+				VStack(spacing: .zero) {
+					ForEach(rows(viewStore: viewStore)) { kind in
+						kind.build(viewStore: viewStore)
```
kugel3, line 63: I don't think that a model should have a func that returns a view, I would greatly prefer having say a RowView that takes a kind, and have the switch in there.

RadixWallet/Features/SettingsFeature/Preferences/Preferences+View.swift
```diff
@@ -0,0 +1,182 @@
...
+				icon: .systemImage("eye.fill"),
+				action: .hiddenEntitiesButtonTapped
+			),
+			.header(S.advancedPreferences),
+			.model(
+				title: S.gateways,
+				icon: .asset(AssetResource.gateway),
+				action: .gatewaysButtonTapped
+			),
+			.custom(developerMode(viewStore: viewStore), id: "developerMode"),
```
kugel3, line 115: What is custom and why does it take an AnyView? Let's make the id at least a private let.

RadixWallet/Features/SettingsFeature/Preferences/Preferences.swift
```diff
@@ -0,0 +1,130 @@
...
+	public init() {}
+
+	public var body: some ReducerOf<Self> {
+		Reduce(core)
+			.ifLet(destinationPath, action: /Action.destination) {
+				Destination()
+			}
+	}
+
+	private let destinationPath: WritableKeyPath<State, PresentationState<Destination.State>> = \.$destination
```
kugel3, line 65: Same story here, typechecking can't handle it otherwise

RadixWallet/Features/SettingsFeature/Preferences/Preferences.swift
```diff
@@ -0,0 +1,130 @@
...
+		switch internalAction {
+		case let .loadedAppPreferences(appPreferences):
+			state.appPreferences = appPreferences
+			return .none
+		}
+	}
+
+	public func reduceDismissedDestination(into state: inout State) -> Effect<Action> {
+		if
+			case let .depositGuarantees(depositGuarantees) = state.destination,
```
kugel3, line 115: I don't think `if case` should be broken up, and the ending brace does not need a newline. Some ex-Swift developers in the team may disagree with me...

RadixWallet/Features/SettingsFeature/SecurityFactors/SecurityFactors+View.swift
```diff
@@ -0,0 +1,143 @@
...
+		}
+	}
+
+	func rows(viewStore: ViewStoreOf<SecurityFactors>) -> [SettingsRow<SecurityFactors>] {
+		[
+			.header(S.subtitle),
+			.model(
+				title: S.SeedPhrases.title,
+				subtitle: S.SeedPhrases.subtitle,
+				detail: seedPhrasesDetail(viewStore),
```
kugel3, line 65: I personally really dislike passing in ViewStore, I only do it when absolutely needed, which nowadays is only when you need a binding - since you can send directly to store. Keep in mind also that viewstore is going away very soon.

RadixWallet/Features/SettingsFeature/SecurityFactors/SecurityFactors+View.swift
```diff
@@ -0,0 +1,143 @@
...
+		}
+		return [.init(kind: .warning, text: .init(S.SeedPhrases.enterSeedPhrase))]
+	}
+
+	func ledgerWalletsDetail(_ viewStore: ViewStoreOf<SecurityFactors>) -> String? {
+		guard let count = viewStore.ledgerWalletsCount else {
+			return nil
+		}
+		return count == 1 ? S.LedgerWallet.counterSingular : S.LedgerWallet.counterPlural(count)
+	}
```
kugel3, line 99: All these that take the viewstore and return a simple value, I really like it better when those are just computed properties on the state.

RadixWallet/Features/SettingsFeature/SecurityFactors/SecurityFactors+View.swift
```diff
@@ -0,0 +1,143 @@
...
+	private func ledgerHardwareWallets(with destinationStore: PresentationStoreOf<SecurityFactors.Destination>) -> some View {
+		navigationDestination(
+			store: destinationStore,
+			state: /SecurityFactors.Destination.State.ledgerWallets,
+			action: SecurityFactors.Destination.Action.ledgerWallets,
+			destination: {
+				LedgerHardwareDevices.View(store: $0)
+					.background(.app.gray5)
+					.navigationTitle(L10n.AccountSecuritySettings.LedgerHardwareWallets.title)
+					.toolbarBackground(.visible, for: .navigationBar)
```
kugel3, line 139: Is this an Introspection modifier or native?

RadixWallet/Features/SettingsFeature/SettingsRow.swift
```diff
@@ -0,0 +1,115 @@
+import SwiftUI
+
+// MARK: - SettingsRow
+enum SettingsRow<Feature: FeatureReducer> {
+	/// A standard tappable row with the details specified on the `Model`
+	case model(Model)
+
+	/// A custom row with its own UI. Useful, for example, when we want a `ToggleView`.
+	case custom(AnyView, id: String)
```
kugel3, line 9: I suppose the AnyView is unavoidable if we want a completely custom row type. But do we actually need that? Seems to go somewhat against the core idea, will we really have a bunch of completely custom rows in settings?

RadixWallet/Features/SettingsFeature/SettingsRow.swift
```diff
@@ -0,0 +1,115 @@
...
+
+	/// A small row acting as a section header with the provided title.
+	case header(String)
+
+	/// Similar to the `.header`, but with no title.
+	case separator
+
+	@ViewBuilder
+	func build(viewStore: ViewStoreOf<Feature>) -> some View {
```
kugel3, line 18: I would prefer this to not be a property on the model, I want the model to just represent the idea of the thing. So I think this enum should be called SettingsRowModel, and then you have SettingsRow which is a View that is created with a model. Or something along those lines. As long as the model is more separate from the view!

RadixWallet/Features/SettingsFeature/SettingsRow.swift
```diff
@@ -0,0 +1,115 @@
...
+			"separator"
+		case let .header(value):
+			value
+		}
+	}
+}
+
+// MARK: - Helper
+extension SettingsRow {
+	static func model(
```
kugel3, line 94: This is a bit confusing for me, does it add anything compared to just having an init on the model type?

kugel3 (approved):
Very solid PR, great work. Just a number of style questions. Also, when you are going over and updating or creating features, please take the opportunity to:

1. Use at least CasePathable whenever applicable, so we can get away from the old hideous CasePaths in scopes etc
2. Use ImageResource instead of ImageAsset / AssetResource

Not only can we write `Image(.arrowBack)` in place of `Image(asset: AssetResource.arrowBack)`, it is native, doesn't rely on shell scripts.

RadixWallet/Core/DesignSystem/Components/Hint.swift
kugel3: I am sure your changes to Hint are thought through, I was only really talking about the practice of using UUID as ID for Identifiable. When I say risky I meant that SwiftUI (and also TCA) may sometimes create new copies of things, and if we happen to then also give them a new UUID, the ID system will break down. So it's important to make sure that we are really only creating the UUID once, for example when we fetch something from an API or something. Definitely not in say a computed property.

Anyway, that was just a general comment, I don't know if you are doing that here, and in any case even if things did go wrong here, it wouldn't be a big issue, at most a visual glitch. But for example in TransactionHistory, I use UUID in one place, and there it would indeed be pretty bad if we accidentally created a new ID for some entry.

RadixWallet/Core/DesignSystem/Components/AssetIcon.swift
kugel3: Wow that was ugly, good catch. I wonder if it's been like that for long.

RadixWallet/Core/DesignSystem/Components/Hint.swift
matiasbzurovski: Yes, I total see where you come from and why you highlighted the risks of using an UUID for something that isn't actually a model. Let me explain why I proposed this Hint refactor solution.

With the possibility of having multiple hints chained (as we discussed on that thread with Aftab and Matt), I needed to add the `UUID` in order to allow looping with a `ForEach` an array of `[Hint.ViewState]`. Given that its state is defined by a `View` (a `Text`), it isn't possible to identify them. The cleanest solution would be changing the entire View and uses of it, but as the current solution isn't critical (as you say, worst case it has a UI glitch), I wanted to ask for opinions before diving into it.

RadixWallet/Features/SettingsFeature/Preferences/Preferences+View.swift
matiasbzurovski: Yeah, I learned that from having my own personal Xcode freeze experience 😁  thanks

RadixWallet/Features/SettingsFeature/Preferences/Preferences+View.swift
matiasbzurovski: maybe we can put it up to a vote then? cc @CyonAlexRDX @GhenadieVP @danvleju-rdx 
I see two benefits of this approach:
- Save saves spaces and makes usage of localization shorter
- Prevents from adding the wrong string by mistake. For example, it is easy to confuse and autocomplete with:
```Swift
.navigationTitle(L10n.AppSettings.title)
```
when you actually needed:
```Swift
.navigationTitle(L10n.AccountSettings.title)
```

A similar case like this happened to me while working on this PR.

RadixWallet/Features/SettingsFeature/Preferences/Preferences+View.swift
matiasbzurovski: If you go into `SettingsRow.custom` it explains what custom is. It allows to adding a different type of row that one defined by `PlainListRow`.

RadixWallet/Features/SettingsFeature/SecurityFactors/SecurityFactors+View.swift
matiasbzurovski: native

RadixWallet/Features/SettingsFeature/SettingsRow.swift
matiasbzurovski: not right now at least. We could delete the entire concept and in the usage of it, instead of simply iteration over an array of `[SettingsRow]`, we would iterate over a first array of `[SettingsRow]`, then add the custom one, then iterate over an array of remaining `[SettingsRow]`. 

I can remove it for now and if in the future we have more custom rows we consider adding it back.

RadixWallet/Features/SettingsFeature/SettingsRow.swift
```diff
@@ -0,0 +1,115 @@
...
+		title: String,
+		subtitle: String? = nil,
+		detail: String? = nil,
+		hints: [Hint.ViewState] = [],
+		icon: AssetIcon.Content,
+		accessory: ImageAsset? = AssetResource.chevronRight,
```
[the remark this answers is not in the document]
matiasbzurovski: This was code moved around, but will do.

RadixWallet/Features/SettingsFeature/SettingsRow.swift
matiasbzurovski: No, just a helper method to do `.model(title:)` instad of `.init(.model(title:))`. Will be removed any way.

RadixWallet/Core/DesignSystem/Components/PlainListRow.swift
```diff
@@ -49,45 +53,82 @@ public struct PlainListRow<Icon: View>: View {
...
-				icon
-					.padding(.trailing, .medium3)
-			}
+		VStack(alignment: .leading) {
+			top
+			hints
+		}
+		.padding(.vertical, viewState.rowCoreViewState.verticalPadding)
+		.padding(.horizontal, .medium3)
+		.frame(minHeight: .settingsRowHeight)
```
GhenadieVP, line 66: hmm, is there actually a requirement to have a min height? Isn't it about having the same top and bottom padding?

RadixWallet/Features/SettingsFeature/DebugSettings/Coordinator/DebugSettingsCoordinator+View.swift
```diff
@@ -1,8 +1,16 @@
...
 
+extension DebugSettingsCoordinator.State {
+	var viewState: DebugSettingsCoordinator.ViewState {
+		.init()
+	}
+}
+
 // MARK: - DebugSettingsCoordinator.View
 extension DebugSettingsCoordinator {
+	public struct ViewState: Equatable {}
```
GhenadieVP, line 12: Redundant? why was this added?

RadixWallet/Features/SettingsFeature/Preferences/Preferences.swift
```diff
@@ -0,0 +1,130 @@
...
+	public func reduce(into state: inout State, internalAction: InternalAction) -> Effect<Action> {
+		switch internalAction {
+		case let .loadedAppPreferences(appPreferences):
+			state.appPreferences = appPreferences
+			return .none
+		}
+	}
+
+	public func reduceDismissedDestination(into state: inout State) -> Effect<Action> {
+		if
```
GhenadieVP, line 114: Prefer `guard` instead

RadixWallet/Features/SettingsFeature/SecurityFactors/SecurityFactors+View.swift
```diff
@@ -0,0 +1,143 @@
...
+				title: S.LedgerWallet.title,
+				subtitle: S.LedgerWallet.subtitle,
+				detail: ledgerWalletsDetail(viewStore),
+				icon: .asset(AssetResource.ledger),
+				action: .ledgerWalletsButtonTapped
+			),
+		]
+	}
+
+	func seedPhrasesDetail(_ viewStore: ViewStoreOf<SecurityFactors>) -> String? {
```
GhenadieVP, line 80: As @kugel3 also mentioned, there is really no point as it seems to passing the viewStore, you could just pass the propert, like here, it seems that just passing `seedPhrasesCount` to this function should be enough.

You may want to pass the viewStore if you want to send actions, which doesn't seem to be the case.

RadixWallet/Features/SettingsFeature/SettingsRow.swift
GhenadieVP: Is this added to be able to have a `toggle` in a settings row?

RadixWallet/Features/SettingsFeature/Settings+View.swift
```diff
@@ -176,28 +116,46 @@ extension Settings.View {
 	}
 
 	@MainActor
-	private func normalRows(viewStore: ViewStoreOf<Settings>) -> [SettingsRowModel<Settings>] {
+	private func normalRows(viewStore: ViewStoreOf<Settings>) -> [SettingsRow<Settings>] {
```
GhenadieVP, line 119: ViewStore does not seem to be used anymore in this funciton.

RadixWallet/Features/SettingsFeature/Troubleshooting/Troubleshooting+View.swift
```diff
@@ -0,0 +1,117 @@
...
+
+extension Troubleshooting.State {
+	var viewState: Troubleshooting.ViewState {
+		.init()
+	}
+}
+
+// MARK: - Troubleshooting.View
+public extension Troubleshooting {
+	struct ViewState: Equatable {}
```
GhenadieVP, line 11: Redundant?

GhenadieVP (commented):
Amazing work, some things need to be clarified before merging.

GhenadieVP:
Additional thing - according to this [screen](https://app.zeplin.io/project/639c7e7ab391bb15ac5d25fa/screen/66210084a17f5c375302a6c3), there should be an X button in top right.

RadixWallet/Core/DesignSystem/Components/PlainListRow.swift
```diff
@@ -49,45 +53,82 @@ public struct PlainListRow<Icon: View>: View {
...
-				Image(asset: accessory)
+	@ViewBuilder
+	private var hints: some View {
+		if !viewState.hints.isEmpty {
+			HStack(spacing: .zero) {
+				icon
+					.hidden() // to leave the same leading padding than on top view
+
+				VStack(alignment: .leading, spacing: .small1) {
+					ForEach(viewState.hints) { hint in
```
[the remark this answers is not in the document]
matiasbzurovski: Great example on how I was overcomplicating a simple problem. thanks!

RadixWallet/Core/DesignSystem/Components/PlainListRow.swift
matiasbzurovski: This min height actually is necessary for `PlainListRows` that have `kind: .general`. This is, every other row except those used in Settings views. The `minHeight` was already set, and doesn't actually applies for rows with `kind: .settings` as with the vertical padding they will take more space than the min provided. However, if I remove this line, other rows will look smalled and look bad. 

I do agree that the name of the constant isn't appropiate (it is called `settingsRowHeight` when it actually applies to every row but settings ones 🙂). Maybe we rename it to `plainRowMinHeight`?

RadixWallet/Features/SettingsFeature/DebugSettings/Coordinator/DebugSettingsCoordinator+View.swift
matiasbzurovski: The reason I added this empty `ViewState` is because the new `SettingsRow` now requires a `ViewStore` in order to send the corresponding action when the rows are tapped. Using an empty `ViewState` is the easiest way I've found to pass a `ViewStore` (and not the `Store` we already have), but open to suggestions if you have a better way of doing so.

matiasbzurovski:
@GhenadieVP thanks for your review, will tackle all comments now. 

Regarding the extra `X`, the designs are outdated and it should be removed according to this [thread](https://rdxworks.slack.com/archives/C031A0V1A1W/p1712839821629559?thread_ts=1712838388.964059&cid=C031A0V1A1W) I read from Matt

CyonAlexRDX (approved):
Great job! Looks much nicer!

RadixWallet/Features/SettingsFeature/DebugSettings/Coordinator/DebugSettingsCoordinator+View.swift
matiasbzurovski: Well, actually we could always send `Store` and just do `store.send(.view(action))`. Will do that

GhenadieVP: yup, you can just use store :)

RadixWallet/Clients/DeviceFactorSourceClient/DeviceFactorSourceClient+Live.swift
matiasbzurovski: But that logic is related to a particular account, isn't it? I want to know if there is any account for which the user needs to enter the seed phrase, in order to show the warning like this [mockup](https://app.zeplin.io/project/639c7e7ab391bb15ac5d25fa/screen/6616fbde81ea923357799a43)

RadixWallet/Core/DesignSystem/Components/Hint.swift
matiasbzurovski: In that case, maybe we can update the `error` asset to be a `24x24` (instead of the current `17x15`)? It is already a template image so it wouldn't break on other places

RadixWallet/Clients/DeviceFactorSourceClient/DeviceFactorSourceClient+Live.swift
GhenadieVP: I mostly meant to be sure it is the same logic but for many accounts,
but it is fine, I just recalled that `isMnemonicPresentInKeychain` uses the same code to check for mnemonic presence.
So to me, your check here is fine 👍.

RadixWallet/Core/DesignSystem/Styles/SecondaryRectangularButtonStyle.swift
matiasbzurovski: I see your point. This is needed for this view, which has a background of `gray3` (and not `gray4`).

Screenshot 2024-04-19 at 13 16 15

I believe it is better having the `backgroundColor` of this style as modifiable as any of its other properties than repeating this logic on a `TertiaryRectangularButtonStyle`.

RadixWallet/Features/SettingsFeature/SettingsRow.swift
matiasbzurovski: Yes @GhenadieVP, will remove it though.

GhenadieVP: ok, @matiasbzurovski I am thinking that it could be worth extending PlainListRow to accept arbitrary accessory view, unless you are doing exactly this alread.

matiasbzurovski: not really, I am reusing the `ToggleView` which internally uses `PlainListRowCore`

RadixWallet/Core/DesignSystem/Components/Hint.swift
kugel3: sounds good

RadixWallet/Core/DesignSystem/Components/Hint.swift
kugel3: The downside with using index is that the view will not behave correctly if the array ever changes. To be precise, if the change is "add new rows at the end" it will work, but if you insert some new element at say row 3, SwiftUI will not know that this is different from the old index 3. Sometimes that will give incorrect results.

RadixWallet/Core/DesignSystem/Styles/SecondaryRectangularButtonStyle.swift
kugel3: Then I think you should first check if that difference was actually intentional, it could also be a mistake. If it is intentional, and perhaps it's because the view background is not white, then I think there should rather be either a boolean flag like "hasNonWhiteBackground" or a Context enum and parameter that is either something like default, some clever description of this situation here. In short, I think we should strive for "semantic" parameters when possible.

RadixWallet/Features/SettingsFeature/Preferences/Preferences+View.swift
kugel3: The downside is that it makes it hard to read, you have to scroll up to see what this means. In this case, there is no reason to use it since space is not at a premium.

For the second argument, that is good as far as it goes, but the relation between screens and string groups is not entirely 1-to-1.

But the biggest downside is that it makes some searches impossible. I often search for parts of the string key.

RadixWallet/Features/SettingsFeature/SecurityFactors/SecurityFactors+View.swift
kugel3: Aha, that is great

RadixWallet/Features/SettingsFeature/DebugSettings/Coordinator/DebugSettingsCoordinator+View.swift
kugel3: Yeah, I am doing that for all new features, and I have stopped making ViewState in many cases, since it is going away.

RadixWallet/Core/DesignSystem/Components/PlainListRow.swift
kugel3: The plain list row has evolved a bit since I introduced it, it's not so plain anymore. Feel free to change the height stuff, just keep in mind that we never want to set absolute heights since we want Dynamic Type to work. Instead we rely on padding outside the text bit as much as possible. An absolute minimum height is fine though, since it won't interfere with Dynamic Type. 

I don't remember what `settingsRowHeight` was back then but it might have broken this rule, so it's great if you clean it up.

RadixWallet/Features/SettingsFeature/Preferences/Preferences+View.swift
kugel3: Actually the biggest reason I am against this is that it makes strings less visually distinctive, so if you guys really insist on this I think we should use something better than `S`. Perhaps `L13g` for "localisedstring" or something clever.

RadixWallet/Core/DesignSystem/Components/Hint.swift
matiasbzurovski: I just pushed the following changes:
- `error` image was updated to be a square one
- `warning_error` was renamed to `error_large`. The asset is the same than `error`, but it is 4.3x times its size, so it makes sense to have the same name and just the `large` suffix.
- Updated usages of `error_large` (that were actually resizing the image to have `.frame(.smallest)`) to just use `.error` (whose size matches `.smallest`).

RadixWallet/Features/SettingsFeature/Preferences/Preferences+View.swift
kugel3: Yes that's why I wrote "some" searches. But even here, it gives you the file, not the line, so you can't tell from the left pane search results, you need to go into the file.

GhenadieVP (approved):

RadixWallet/Features/SettingsFeature/Preferences/Preferences+View.swift
kugel3: Since it's not needed for this screen - there is no space constraint, let's postpone this and use the full key for now?

PR #1089, opened 2024-04-17
[ABW-2879] CAP-36 New Linking Flow

danvleju-rdx:
Jira ticket: [ABW-2879](https://radixdlt.atlassian.net/browse/ABW-2879)

## Description
Implements the new linking flow, allowing for the linking of multiple phones using one browser.

### Notes
`P2PLinks` were moved from the `Profile` to the `SecureStorage`. The "migration" logic (which will contain displaying messages to users to address the drop of the old links) will be handled in a separate PR. The main idea is to decode `Profile` as `JSONValue` and check if the user has old p2pLinks. In future, this logic could be transferred to Sargon.

Sending the `AccountListMessage` to the CE will also be handled in a separate PR.

## Videos
<details closed>
<summary>New connection</summary>

https://github.com/radixdlt/babylon-wallet-ios/assets/163979791/f78783ee-786a-4ac5-ac89-cca20ccb6c37

</details>
<details closed>
<summary>Existing connection</summary>

https://github.com/radixdlt/babylon-wallet-ios/assets/163979791/6dc16c59-d9e5-4cb7-884b-ae6e42c2012d

</details>

[ABW-2879]: https://radixdlt.atlassian.net/browse/ABW-2879?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #1098, opened 2024-04-24
CAP-36 New Linking Flow Updates

danvleju-rdx:
Updates for PR [#1089](https://github.com/radixdlt/babylon-wallet-ios/pull/1089)

## Description
What was updated:
- After 1.6.0 is out, we will for sure encounter cases when users scan the old QR code, we may want to identify this and tell them that they are using old CE
- Updated the Wallet to have only one public key for all CE links, so CE can identify the Wallet as being the same client in the same way the Wallet sees the CE being the same client

### DIFF
--- a/RadixWallet/Core/Resources/Resources/en.lproj/Localizable.strings
+++ b/RadixWallet/Core/Resources/Resources/en.lproj/Localizable.strings
@@ -1167,5 +1167,5 @@ Only continue if you are linking to the **official Radix Connector browser exten
 "linkedConnectors_linkFailedErrorTitle" = "Link Failed";
 "linkedConnectors_unknownPurposeErrorMessage" = "This type of Connector link is not supported.";
 "linkedConnectors_changingPurposeNotSupportedErrorMessage" = "Changing a Connector’s type is not supported.";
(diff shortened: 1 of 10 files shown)


RadixWallet/Core/Resources/Resources/en.lproj/Localizable.strings
```diff
@@ -1167,4 +1167,5 @@ Only continue if you are linking to the **official Radix Connector browser exten
 "linkedConnectors_linkFailedErrorTitle" = "Link Failed";
 "linkedConnectors_unknownPurposeErrorMessage" = "This type of Connector link is not supported.";
 "linkedConnectors_changingPurposeNotSupportedErrorMessage" = "Changing a Connector’s type is not supported.";
+"linkedConnectors_oldQRErrorMessage" = "This is an old version of the Radix Connector browser extension. Please update to the latest Connector and try linking again.";
```
GhenadieVP, line 1170: is there a related PR on strings repo?

GhenadieVP (approved):
LGTM!

RadixWallet/Core/Resources/Resources/en.lproj/Localizable.strings
danvleju-rdx: yes

PR #1100, opened 2024-04-26
[ABW-3213] Address Details

matiasbzurovski:
Jira ticket: [ABW-3213](https://radixdlt.atlassian.net/browse/ABW-3213)
Jira ticket: [ABW-3256](https://radixdlt.atlassian.net/browse/ABW-3256)

## Description
This PR adds a sheet presented every time the user taps on an `AddressView` with a `LedgerIdentifiable.Address` associated.

## Video

https://github.com/radixdlt/babylon-wallet-ios/assets/164921079/09c5f6f7-c4a7-459e-b3cb-673da8772178

[ABW-3213]: https://radixdlt.atlassian.net/browse/ABW-3213?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[ABW-3256]: https://radixdlt.atlassian.net/browse/ABW-3256?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


RadixWallet/Core/DesignSystem/LinearGradients.swift
```diff
@@ -31,6 +31,21 @@ extension Gradient {
 	]
 }
 
+extension LinearGradient.App {
+	public var brand1: LinearGradient {
```
matiasbzurovski, line 35: Ended up not being used on these designs but decided to keep it as we will probably need it somewhere else (it is part of the [design system](https://www.figma.com/file/6MaKVYek0m9GKkd1xujB8P/Babylon-Wallet?type=design&node-id=405%3A3070&mode=design&t=dn5fkwpAWwUic1bK-1).

RadixWallet/Core/FeaturePrelude/AddressView/AddressDetails+View.swift
```diff
@@ -0,0 +1,238 @@
...
+				loadable(title) {
+					ProgressView()
+				} successContent: { title in
+					Text(title)
+						.textStyle(.sheetTitle)
+						.foregroundColor(.app.gray1)
+				}
+
+				if showQrCode {
+					Text("Address QR Code")
```
matiasbzurovski, line 83: Will update & use localisaton once https://github.com/radixdlt/apps-localization-src/pull/367 gets merged

RadixWallet/Core/FeaturePrelude/AddressView/AddressDetails.swift
```diff
@@ -0,0 +1,187 @@
+// MARK: - AddressDetails
+public struct AddressDetails: Sendable, FeatureReducer {
+	public struct State: Sendable, Hashable {
+		let address: LedgerIdentifiable.Address
+		let closeAction: (() -> Void)?
```
matiasbzurovski, line 5: It really bothers me that we need to send a non-sendable closure to this state.. But it is the only way I was able to handle dismissal when the view is presented from `AddressView` (which is a plain SwiftUI vanilla view). Open to better suggestions.

GhenadieVP:
Note: This is to be merged for the next 1.6.0 release if Android manages to implement it, otherwise it will go in 1.7.0 release.

RadixWallet/Core/FeaturePrelude/AddressView/AddressDetails+View.swift
```diff
@@ -0,0 +1,266 @@
...
+			qrImage: qrImage,
+			showEnlarged: showEnlarged,
+			showShare: showShare
+		)
+	}
+}
+
+// MARK: - AddressDetails.View
+
+public extension AddressDetails {
```
kugel3, line 15: New features should ideally use the observable style. You will need to add at least one `WithPerceptionTracking`, sometimes multiple.

RadixWallet/Core/FeaturePrelude/AddressView/AddressDetails+View.swift
```diff
@@ -0,0 +1,266 @@
...
+				Group {
+					if viewStore.showEnlarged {
+						enlargedView(text: viewStore.enlargedAddress)
+					}
+				}
+				.animation(.easeInOut, value: viewStore.showEnlarged)
+			}
+		}
+
+		private func top(title: Loadable<String?>, showQrCode: Bool, qrImage: Loadable<CGImage>) -> some SwiftUI.View {
```
kugel3, line 84: I personally prefer to make these into nested View structs when they take a bunch of parameters. But with observation you shouldn't have to pass anything in, since you can access `store` from the functions.

RadixWallet/Core/FeaturePrelude/AddressView/AddressDetails+View.swift
```diff
@@ -0,0 +1,266 @@
...
+			.frame(width: UIScreen.main.bounds.width * 2, height: UIScreen.main.bounds.height * 2)
+			.contentShape(Rectangle())
+			.onTapGesture {
+				store.send(.view(.hideEnlargedView))
+			}
+	}
+}
+
+private extension AddressDetails.ViewState {
+	static func colorised(address: LedgerIdentifiable.Address) -> AttributedString {
```
kugel3, line 210: I think this should be either a local private function, or, if you might want to use it somewhere else, a `var` on `LedgerIdentifiable.Address` or something.

RadixWallet/Core/FeaturePrelude/AddressView/AddressDetails.swift
kugel3: Doesn't `@Environment(\.dismiss) var dismiss` work (in the View)? Or `@Dependency(\.dismiss) var dismiss`?

kugel3: I mean the reducer doesn't need to know that we closed it, right? It could just be done on the View side?

RadixWallet/Core/FeaturePrelude/AddressView/AddressDetails.swift
```diff
@@ -0,0 +1,190 @@
...
+	@Dependency(\.gatewaysClient) var gatewaysClient
+	@Dependency(\.openURL) var openURL
+	@Dependency(\.ledgerHardwareWalletClient) var ledgerHardwareWalletClient
+
+	public init() {}
+
+	public func reduce(into state: inout State, viewAction: ViewAction) -> Effect<Action> {
+		switch viewAction {
+		case .task:
+			return loadTitle(state: &state)
```
kugel3, line 52: Let's use the `-Effect` suffix for these, just to make it clearer

RadixWallet/Core/FeaturePrelude/AddressView/AddressDetails.swift
```diff
@@ -0,0 +1,190 @@
...
+			return .none
+		}
+	}
+
+	private func loadTitle(state: inout State) -> Effect<Action> {
+		state.title = .loading
+		return .run { [address = state.address] send in
+			let result: TaskResult<String?> = switch address {
+			case let .account(address, _):
+				await TaskResult {
```
kugel3, line 115: Can't we just wrap the whole thing in a `TaskResult`?

RadixWallet/Core/FeaturePrelude/AddressView/AddressDetails.swift
```diff
@@ -0,0 +1,190 @@
...
+			}
+			await send(.internal(.loadedQrImage(result)))
+		}
+	}
+}
+
+// MARK: - Helpers
+
+private extension OnLedgerEntity.Resource {
+	var resourceTitle: String? {
```
kugel3, line 169: We used to have something with a very similar name, which returned `metadata.symbol ?? metadata.name`, but I can't find it right now. If it still exists somewhere it might be a bit confusing.

RadixWallet/Core/FeaturePrelude/AddressView/AddressDetails.swift
matiasbzurovski: I literally don't understand what was the problem back then what I added this.. removing the whole closure concept.

RadixWallet/Core/FeaturePrelude/AddressView/AddressDetails+View.swift
kugel3: We usually put `// FIXME: Strings` on these so that we can find them later in case we forget

RadixWallet/Core/FeaturePrelude/AddressView/AddressDetails.swift
matiasbzurovski: Probably [this](https://github.com/radixdlt/babylon-wallet-ios/blob/main/RadixWallet/Clients/OnLedgerEntitiesClient/Helpers/OnLedgerEntitiesClient%2BCreateEntity.swift#L582-L585) `title` defined on `OnLedgerEntity.Metadata`. 

The one I defined is private anyway, and is an improved version of it. `smartResourceTitle`?

RadixWallet/Core/FeaturePrelude/AddressView/AddressDetails+View.swift
matiasbzurovski: The idea was to computate it only once

RadixWallet/Core/FeaturePrelude/AddressView/AddressDetails+View.swift
```diff
@@ -1,121 +1,93 @@
...
 		}
 
-		private func fullAddress(address: AttributedString) -> some SwiftUI.View {
-			VStack(spacing: .medium2) {
-				VStack(spacing: .small2) {
-					Text(L10n.AddressDetails.fullAddress)
-						.foregroundColor(.app.gray1)
-					Text(address)
+		private func qrCode(qrImage: Loadable<CGImage>) -> some SwiftUI.View {
+			ZStack {
```
kugel3, line 73: What does the ZStack do here? It looks like each branch only returns a single element, so it would seem that it would be enough to put `@ViewBuilder` on the function

RadixWallet/Core/FeaturePrelude/AddressView/AddressDetails+View.swift
```diff
@@ -231,11 +170,40 @@ private extension AddressDetails.ViewState {
 
 		return result
 	}
+}
 
-	static func enlarged(address: LedgerIdentifiable.Address) -> AttributedString {
+// MARK: - Enlarged view
+private extension AddressDetails.View {
+	var enlargedView: some SwiftUI.View {
+		Group {
```
kugel3, line 178: What is the purpose of the Group?

RadixWallet/Core/FeaturePrelude/AddressView/AddressDetails.swift
```diff
@@ -99,51 +106,42 @@ public struct AddressDetails: Sendable, FeatureReducer {
...
+			await send(.internal(.loadedTitle(TaskResult {
+				try await loadTitle(address: address)
+			})))
+		}
+	}
+
+	private func loadTitle(address: LedgerIdentifiable.Address) async throws -> String? {
+		switch address {
+		case let .account(address, _):
+			let account = try await accountsClient.getAccountByAddress(address)
```
kugel3, line 121: I suppose it's a matter of taste but it looks like all these branches could be written as one-liners, without a `return`.
```
try await accountsClient.getAccountByAddress(address).displayName.rawValue
```

RadixWallet/Core/FeaturePrelude/AddressView/AddressDetails.swift
```diff
@@ -99,51 +106,42 @@ public struct AddressDetails: Sendable, FeatureReducer {
...
-				await TaskResult {
-					let resource = try await onLedgerEntitiesClient.getResource(globalId.resourceAddress)
-					return resource.resourceTitle
-				}
-			}
-			await send(.internal(.loadedTitle(result)))
+			await send(.internal(.loadedTitle(TaskResult {
+				try await loadTitle(address: address)
+			})))
+		}
```
kugel3, line 115: Much cleaner!

RadixWallet/Core/FeaturePrelude/AddressView/AddressDetails.swift
```diff
@@ -1,18 +1,21 @@
 // MARK: - AddressDetails
 public struct AddressDetails: Sendable, FeatureReducer {
```
kugel3, line 2: You can add `@Reducer` to this. Then you need to also put `public typealias Action = FeatureAction<Self>` explicitly, otherwise the compiler gets confused.

RadixWallet/Core/FeaturePrelude/AddressView/AddressDetails.swift
kugel3: I wouldn't say it's improved, it's different, and only makes sense here

RadixWallet/Core/FeaturePrelude/AddressView/AddressDetails+View.swift
kugel3: The ViewState is actually re-computed every time anything changes, as is the entire SwiftUI View, so I don't think it makes any difference. 

But either way, my idea was just to change the ergonomics to something like this:

`let addressString = address.colorizedString`

It might be that Observable is smart enough to only recompute it when the underlying address changes if you do it like that. When it's in the ViewState, any change to any property on the ViewState would recompute everything.

Seems like it could be useful more generally, btw.

RadixWallet/Core/FeaturePrelude/AddressView/AddressDetails.swift
kugel3: So I would just make it a private computed property in the View

kugel3:
Now the dismiss doesn't actually work for me in your PR. But putting`@Environment(\.dismiss) var dismiss` in the View instead does work.

RadixWallet/Core/FeaturePrelude/AddressView/AddressDetails+View.swift
matiasbzurovski: probably left over from old implementation, removing

RadixWallet/Core/FeaturePrelude/AddressView/AddressDetails.swift
matiasbzurovski: and what's the benefit of doing so if we still need to implement every `func reduce` we currently have?

RadixWallet/Core/FeaturePrelude/AddressView/AddressDetails.swift
matiasbzurovski: It is already a private computed property, but it is an extension of the `OnLedgerEntity.Resource`, and not of the View, since it is used in multiple places
```Swift
try await onLedgerEntitiesClient.getResource(address).resourceTitle
try await onLedgerEntitiesClient.getResource(globalId.resourceAddress).resourceTitle
```

Given it is already private, I will leave it as is.

RadixWallet/Core/FeaturePrelude/AddressView/AddressDetails.swift
kugel3: It's usually good to stay up to date with the latest generation of TCA, so we can enjoy any future improvements coming to `@Reducer` they make.

GhenadieVP (approved):
LGTM!

PR #1103, opened 2024-04-29
Sargon profile expanded.

CyonAlexRDX:
Blocked by https://github.com/radixdlt/sargon/pull/110

After that PR is merged the SPM of Sargon should be updated, this PR currently uses local Sargon, so will fail CI/CD.

Replaced `AccountOrPersona` with impl in Sargon, replaces both model and logic impl of `AuthorizedPersonaDetailed` with impl in Sargon

Removes `EntityProtocol` since it has been fully migrated into Sargon.

# Demo
Demo of `AuthorizedDappDetailed`

https://github.com/radixdlt/babylon-wallet-ios/assets/116169792/14da79ff-0ae2-4564-899f-0a6b14506d30


RadixWallet/Clients/ImportLegacyWalletClient/ImportLegacyWalletClient+Live.swift
```diff
@@ -41,10 +41,10 @@ extension ImportLegacyWalletClient: DependencyKey {
 
 				let appearanceID = await accountsClient.nextAppearanceID(networkID, accountOffset)
 
-				let babylon = try Account(
+				let babylon = Account(
```
Sajjon, line 44: many previously throwing methods and inits are no longer throwing in Sargon. We should all help out cascading those changes to calling scopes (and scopes calling those calling scopes...)

RadixWallet/Clients/ROLAClient/ROLAClient+Live.swift
```diff
@@ -9,18 +9,7 @@ extension ROLAClient {
...
-				case let .account(account):
-					assert(account.networkID == request.entity.networkID)
-					return AddressOfAccountOrPersona.account(account.address)
-				case let .persona(persona):
-					assert(persona.networkID == request.entity.networkID)
-					return AddressOfAccountOrPersona.identity(persona.address)
-				}
-			}()
-
+			let entityAddress = entity.address
```
Sajjon, line 12: 🎉

RadixWallet/Features/DappInteractionFeature/Children/AccountPermissionChooseAccounts/AccountPermissionChooseAccounts+Reducer.swift
```diff
@@ -167,9 +168,7 @@ struct AccountPermissionChooseAccounts: Sendable, FeatureReducer {
 
 				var accountsLeftToVerifyDidSign: Set<Account.ID> = Set(selectedAccounts.map(\.id))
 				let walletAccountsWithProof: [P2P.Dapp.Response.Accounts.WithProof] = signedAuthChallenge.entitySignatures.map {
-					guard case let .account(account) = $0.signerEntity else {
-						fatalError()
-					}
+					let account = try! $0.signerEntity.asAccount()
```
Sajjon, line 171: this `fatalError` -> `try!` ought to be reworked ofc... outside scope of Profile integration.

RadixWallet/Features/DappsAndPersonas/PersonaDetails/PersonaDetails+View.swift
```diff
@@ -174,7 +174,7 @@ extension PersonaDetails.State {
 
 	struct AccountSection: Equatable {
 		let dAppName: String
-		let sharingAccounts: OrderedSet<ProfileNetwork.AccountForDisplay>
+		let sharingAccounts: AccountsForDisplay
```
Sajjon, line 177: `AccountsForDisplay` is its own type now, like `Accounts` and `Personas`...

RadixWallet/Features/Signing/Signer.swift
```diff
@@ -0,0 +1,39 @@
+import Foundation
+import Sargon
+
+// MARK: - Signer
```
Sajjon, line 4: Just moved out from `AccountOrPersona` file. Ought to be moved into Sargon later.

RadixWallet/Features/AccountRecoveryScan/Children/AccountRecoveryScanInProgress/AccountRecoveryScanInProgress.swift
```diff
@@ -210,13 +210,13 @@ public struct AccountRecoveryScanInProgress: Sendable, FeatureReducer {
...
-					let accounts = try await publicHDKeys.enumerated().asyncMap { localOffset, publicHDKey in
+					let accounts = await publicHDKeys.enumerated().asyncMap { localOffset, publicHDKey in
 						let offset = localOffset + globalOffset
 						let appearanceID = await accountsClient.nextAppearanceID(networkID, offset)
-						return try Account(
+						return Account(
 							networkID: networkID,
 							factorInstance: .init(factorSourceId: id, publicKey: publicHDKey),
-							displayName: .init(rawValue: L10n.AccountRecoveryScan.InProgress.nameOfRecoveredAccount) ?? "Unnamed",
+							displayName: (try? .init(validating: L10n.AccountRecoveryScan.InProgress.nameOfRecoveredAccount)) ?? DisplayName(value: "Unnamed"),
```
matiasbzurovski, line 219: Help me understand this. We are using `DisplayName`, which basically is a struct that validates the associated String isn't empty nor longer than 30 characters. And in this case we are trying to init it with the localised version of `nameOfRecoveredAccount`, and if it fails we default to one that we know meets that criteria.

If that's the case, and the only thing we do on the `init(validating:)` is check those two conditions:
1) what would you think about making the `init?` an optional one instead of a throwing method? Maybe I am missing something, but it seems to me that every time the `try?` throws, we are gonna want to deal with a null value (as here) rather than catch an error.

2) when calling `init(nonEmpty:)`, as in the next changed file, how do we know that it meets the length criteria? In other words, why is such initializer not throwing as this one?

RadixWallet/Features/Signing/Signer.swift
```diff
@@ -0,0 +1,39 @@
...
+// FIXME: move elsewhere. not really part of Profile... but.. where? We need some kind of shared target for higher level models that can depend on Profile. We lack such a package right now.
```
matiasbzurovski, line 5: Why not under `MIGRATE_TO_SARGON`?

matiasbzurovski (commented):
Code looks good, but logically would need to test once dependency is unblocked

RadixWallet/Features/Signing/Signer.swift
Sajjon: not planned to do it "soon", basically every single type which is not a `View`, `Reducer` or a `Client` will be moved into Sargon.

RadixWallet/Features/AccountRecoveryScan/Children/AccountRecoveryScanInProgress/AccountRecoveryScanInProgress.swift
Sajjon: `nonEmpty` takes a `NonEmptyString` which is already guaranteed to be non-empty.

There are many improvements to be made for sure... I'm trying to get my monster PR merged ASAP and we will all have to help out improving code base :)

matiasbzurovski: > nonEmpty takes a NonEmptyString which is already guaranteed to be non-empty.

Right, but if I understood correctly the `DisplayName` also validates that it isn't larger than 30 chars. Are we performing such validation?

CyonAlexRDX: yes correct

matiasbzurovski: Then I don't understand why the `init(nonEmpty:)` (performs 1 validation) doesn't throw while the `init(validating:)` (performs 2 validations) does throw.

CyonAlexRDX: I will force unwrap. the L10n should not be empty.

CyonAlexRDX: Or rather use a non throwing (should have been hidden, but here convenient it is not) init `init:value` (memberwise init exported by UniFFI.

CyonAlexRDX:
Added video demo.

PR #1104, opened 2024-04-29
GW 1.5.1 update

GhenadieVP:
Updated the GW models. Mostly new models were added, other models have very minor changes.

Main update is `variant_id` which is now a String, still decoding from Int is supported.


CyonAlexRDX (approved):
Blanket approval... no I did not look at the generated models and I have not tested this build
Local

PR #1105, opened 2024-05-02
Profile Not `Codable`

CyonAlexRDX:
Bumps Sargon to `0.7.4` - containing github.com/radixdlt/sargon/119 changes - which removes conformance to `Codable` on `Profile` since it is too slow.

### DIFF
--- a/RadixWallet/Features/DebugInspectProfile/DebugInspectProfile.swift
+++ b/RadixWallet/Features/DebugInspectProfile/DebugInspectProfile.swift
@@ -16,11 +16,8 @@ public struct DebugInspectProfile: Sendable, FeatureReducer {
 		}
 
 		public var json: String? {
-			@Dependency(\.jsonEncoder) var jsonEncoder
-			let encoder = jsonEncoder()
-			encoder.outputFormatting = [.prettyPrinted, .withoutEscapingSlashes, .sortedKeys]
 			guard
-				let json = try? encoder.encode(profile),
+				case let json = profile.profileSnapshot(),
 				let jsonString = String(data: json, encoding: .utf8)
 			else { return nil }
 			return jsonString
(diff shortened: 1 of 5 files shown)


RadixWallet/Features/DebugInspectProfile/DebugInspectProfile.swift
```diff
@@ -16,11 +16,8 @@ public struct DebugInspectProfile: Sendable, FeatureReducer {
 		}
 
 		public var json: String? {
-			@Dependency(\.jsonEncoder) var jsonEncoder
-			let encoder = jsonEncoder()
-			encoder.outputFormatting = [.prettyPrinted, .withoutEscapingSlashes, .sortedKeys]
 			guard
-				let json = try? encoder.encode(profile),
+				case let json = profile.profileSnapshot(),
```
kugel3, line 20: Maybe it's just me but I hate this usage of `case`. Also, what does "snapshot" mean here? And, how is it different from .jsonData?

Sajjon: "snapshot" is something we have used to not that "this is a snapshot in time of the every changing organism that is Profile". It is now somewhat of a legacy term, but used in many places in the code base. in the coming weeks/months more and more of code touching it will move into Sargon, so will be refactored eventually.

Big picture now Gustaf... :D is this really blocking merge?

GhenadieVP (approved):

RadixWallet/Features/DebugInspectProfile/DebugInspectProfile.swift
kugel3: If you look closely, you will notice that I have not requested any changes. Since I am serializing and deserializing profiles, it is relevant for me to understand the difference between .jsonData or .profileSnapshot, even if it's temporary.

danvleju-rdx (approved):

RadixWallet/Features/DebugInspectProfile/DebugInspectProfile.swift
CyonAlexRDX: @kugel3 the method `jsonData()` calls `profileSnapshot()` so they are the same.

PR #1107, opened 2024-05-07
Disable Create new persona button before checking primacy

matiasbzurovski:
## Description
This PR solves a crash on debug builds reproduced if user tapped on `Create New Persona` too soon.

### Notes
The crash has been identified through Xcode organizer, showing the issue described.

Screenshot 2024-05-07 at 13 30 11

It has happened when using a very large profile with lots of accounts. I haven't been able to reproduce this issue (even with a similar profile) but the added solution should have no harm.


CyonAlexRDX (approved):
LGTM

PR #1110, opened 2024-05-09
Bump Sargon

matiasbzurovski:
Use latest version of Sargon with checksum optimization.


CyonAlexRDX (approved):
Nice, I will merge it

PR #1114, opened 2024-05-10
Security Center Client

kugel3:
## Description
Breaks out the Security Center Client and all related changes, from https://github.com/radixdlt/babylon-wallet-ios/pull/1106


RadixWallet/Clients/AppPreferencesClient/AppPreferencesClient+Live.swift
```diff
@@ -35,6 +33,15 @@ extension AppPreferencesClient: DependencyKey {
 					profile.id,
 					isEnabled ? .enable : .disable
 				)
+			},
+			setIsCloudBackupEnabled: { isEnabled in
```
matiasbzurovski, line 37: When would we use `setIsCloudBackupEnabled` and not `setIsCloudProfileSyncEnabled`?

RadixWallet/Clients/BackupsClient/BackupsClient+Interface.swift
```diff
@@ -31,11 +34,10 @@ public struct BackupsClient: Sendable {
 extension BackupsClient {
 	public typealias SnapshotOfProfileForExport = @Sendable () async throws -> Profile
 	public typealias LoadProfileBackups = @Sendable () async -> Profile.HeaderList?
-
+	public typealias LookupProfileSnapshotByHeader = @Sendable (Profile.Header) async throws -> Profile?
 	public typealias ImportProfileSnapshot = @Sendable (Profile, Set<FactorSourceIDFromHash>) async throws -> Void
+	public typealias DidExportProfileSnapshot = @Sendable (Profile) throws -> Void
```
matiasbzurovski, line 39:
```suggestion
	public typealias DidExportProfile = @Sendable (Profile) throws -> Void
```

to follow the pattern on the changes done [here](https://github.com/radixdlt/babylon-wallet-ios/pull/1114/files#diff-36d3a26d72311238ab6bd57ce5b40bf982f396903cafaf1c9faa97360312bd10R15)

RadixWallet/Clients/CloudBackupClient/CloudBackupClient+Interface.swift
```diff
@@ -0,0 +1,47 @@
+import CloudKit
+import ComposableArchitecture
+import DependenciesAdditions
+import os
+
+// MARK: - CloudBackupClient
+public struct CloudBackupClient: DependencyKey, Sendable {
```
matiasbzurovski, line 7: The 3 files regarding `CloudBackupClient` are not part of the project, so I guess we don't need them on this PR?

RadixWallet/Clients/SecureStorageClient/SecureStorageClient+Interface.swift
```diff
@@ -11,6 +11,7 @@ public struct SecureStorageClient: Sendable {
 	public var loadProfileSnapshotData: LoadProfileSnapshotData
 	public var loadProfileSnapshot: LoadProfileSnapshot
 	public var loadProfile: LoadProfile
+	public var deleteProfile: DeleteProfile
```
matiasbzurovski, line 14: So we already have a `DeleteProfileAndMnemonicsByFactorSourceIDs`, which actually is something as `DeleteProfileAndMnemonicsByProfileId`. 
Do we actually want to ONLY delete the Profile and keep the mnemonics on some scenario?

RadixWallet/Clients/SecurityCenterClient/SecurityCenterClient+Interface.swift
```diff
@@ -0,0 +1,49 @@
...
+	public typealias LastCloudBackup = @Sendable () async -> AnyAsyncSequence<BackupStatus>
+}
+
+// MARK: - SecurityProblem
+public enum SecurityProblem: Hashable, Sendable, Identifiable {
+	case problem3(accounts: Int, personas: Int)
+	case problem5
+	case problem6
+	case problem7
+	case problem9
```
matiasbzurovski, line 26: I would add as a comment of each the description on [Confluence](https://radixdlt.atlassian.net/wiki/spaces/AT/pages/3392569357/Security-related+Problem+States+in+the+Wallet)

RadixWallet/Clients/SecurityCenterClient/SecurityCenterClient+Interface.swift
```diff
@@ -0,0 +1,49 @@
+import CloudKit
+import ComposableArchitecture
+import DependenciesAdditions
+import os
+
+// MARK: - SecurityCenterClient
+public struct SecurityCenterClient: DependencyKey, Sendable {
```
matiasbzurovski, line 7: Not added in project (applies to the 3 files)

RadixWallet/Clients/UserDefaults+Dependency+Extension/UserDefaults+Dependency+Extension.swift
```diff
@@ -138,4 +141,72 @@ extension UserDefaults.Dependency {
...
+	public func setLastCloudBackup(_ result: BackupResult.Result, of profile: Profile) throws {
+		var backups: [UUID: BackupResult] = getLastCloudBackups
+		backups[profile.id] = .init(
+			backupDate: .now,
+			profileHash: profile.hashValue,
+			result: result
+		)
+
+		try save(codable: backups, forKey: .lastCloudBackups)
+		print("  •• did setLastCloudBackup \(result)")
```
matiasbzurovski, line 166: logger?

RadixWallet/Clients/UserDefaults+Dependency+Extension/UserDefaults+Dependency+Extension.swift
```diff
@@ -10,6 +10,9 @@ public enum UserDefaultsKey: String, Sendable, Hashable, CaseIterable {
 	case transactionsCompletedCounter
 	case dateOfLastSubmittedNPSSurvey
 	case npsSurveyUserID
+	case didMigrateKeychainProfiles
+	case lastCloudBackups
+	case lastManualBackups
```
matiasbzurovski, line 15: Do we need the `last` prefix for these? At first I thought we were only setting the last one, but as they correspond to a dictionary of `[UUID: BackupResult]` I would just name them:
```Swift
case cloudBackups
case manualBackups
```

RadixWallet/Clients/UserDefaults+Dependency+Extension/UserDefaults+Dependency+Extension.swift
```diff
@@ -138,4 +141,72 @@ extension UserDefaults.Dependency {
...
+
+	public func setDidMigrateKeychainProfiles(_ value: Bool) {
+		set(value, forKey: Key.didMigrateKeychainProfiles.rawValue)
+	}
+
+	public var getLastCloudBackups: [UUID: BackupResult] {
+		(try? loadCodable(key: .lastCloudBackups)) ?? [:]
+	}
+
+	public func setLastCloudBackup(_ result: BackupResult.Result, of profile: Profile) throws {
```
matiasbzurovski, line 157:
```suggestion
	public func addCloudBackup(_ result: BackupResult.Result, of profile: Profile) throws {
```

RadixWallet/Clients/UserDefaults+Dependency+Extension/UserDefaults+Dependency+Extension.swift
```diff
@@ -138,4 +141,72 @@ extension UserDefaults.Dependency {
...
+			backupDate: .now,
+			profileHash: profile.hashValue,
+			result: result
+		)
+
+		try save(codable: backups, forKey: .lastCloudBackups)
+		print("  •• did setLastCloudBackup \(result)")
+	}
+
+	public func lastCloudBackupValues(for profileID: ProfileID) -> AnyAsyncSequence<BackupResult> {
```
matiasbzurovski, line 169:
```suggestion
	public func cloudBackupValues(for profileID: ProfileID) -> AnyAsyncSequence<BackupResult> {
```

RadixWallet/Clients/UserDefaults+Dependency+Extension/UserDefaults+Dependency+Extension.swift
```diff
@@ -138,4 +141,72 @@ extension UserDefaults.Dependency {
...
+
+	public var getDidMigrateKeychainProfiles: Bool {
+		bool(key: .didMigrateKeychainProfiles)
+	}
+
+	public func setDidMigrateKeychainProfiles(_ value: Bool) {
+		set(value, forKey: Key.didMigrateKeychainProfiles.rawValue)
+	}
+
+	public var getLastCloudBackups: [UUID: BackupResult] {
```
matiasbzurovski, line 153:
```suggestion
	public var getCloudBackups: [UUID: BackupResult] {
```

RadixWallet/Clients/UserDefaults+Dependency+Extension/UserDefaults+Dependency+Extension.swift
```diff
@@ -138,4 +141,72 @@ extension UserDefaults.Dependency {
...
+
+		try save(codable: backups, forKey: .lastCloudBackups)
+		print("  •• did setLastCloudBackup \(result)")
+	}
+
+	public func lastCloudBackupValues(for profileID: ProfileID) -> AnyAsyncSequence<BackupResult> {
+		lastBackupValues(for: profileID, key: .lastCloudBackups)
+	}
+
+	public var getLastManualBackups: [UUID: BackupResult] {
```
matiasbzurovski, line 173:
```suggestion
	public var getManualBackups: [UUID: BackupResult] {
```

RadixWallet/Clients/UserDefaults+Dependency+Extension/UserDefaults+Dependency+Extension.swift
```diff
@@ -138,4 +141,72 @@ extension UserDefaults.Dependency {
...
+		(try? loadCodable(key: .lastManualBackups)) ?? [:]
+	}
+
+	/// Only call this on successful manual backups
+	public func setLastManualBackup(of profile: Profile) throws {
```
matiasbzurovski, line 178:
```suggestion
	public func addManualBackup(of profile: Profile) throws {
```

RadixWallet/Clients/BackupsClient/BackupsClient+Interface.swift
kugel3: yes

RadixWallet/Clients/UserDefaults+Dependency+Extension/UserDefaults+Dependency+Extension.swift
kugel3: It only gets the last one, the ID corresponds to the profile

RadixWallet/Clients/UserDefaults+Dependency+Extension/UserDefaults+Dependency+Extension.swift
kugel3: Why

RadixWallet/Clients/UserDefaults+Dependency+Extension/UserDefaults+Dependency+Extension.swift
kugel3: not necessary I think, this is just a leftover

RadixWallet/Clients/CloudBackupClient/CloudBackupClient+Interface.swift
kugel3: Sort of I guess, since they are never called. But the security center client reads the values that are written only here, so it made sense to put them in the same pr. but the files should also be added to the project for sure.

RadixWallet/Clients/BackupsClient/BackupsClient+Interface.swift
kugel3: But there are lots of other places where we still call it snapshot, including the other APIs on this client, so let's fix that all later and avoid merge issues.

RadixWallet/Clients/AppPreferencesClient/AppPreferencesClient+Live.swift
kugel3: But as of this PR, it is still in use, so it shouldn't be deleted yet

RadixWallet/Clients/SecureStorageClient/SecureStorageClient+Interface.swift
kugel3: This one I don't know, @CyonAlexRDX will tell us what is needed.

RadixWallet/Clients/CloudBackupClient/CloudBackupClient+Live.swift
```diff
@@ -0,0 +1,153 @@
...
+
+extension CloudBackupClient {
+	struct IncorrectRecordTypeError: Error {}
+	struct NoProfileInRecordError: Error {}
+	struct ProfileMissingFromKeychainError: Error { let id: UUID }
+
+	public static let liveValue: Self = .live()
+
+//	private static let container = CKContainer(identifier: "iCloud.com.radixpublishing.radixwallet.ios.dev.cloudBackup")
+	private static let container = CKContainer(identifier: "iCloud.com.radixpublishing.radixwallet.ios.pre-alpha.cloudBackup")
```
Sajjon, line 22: whats the plan for PROD? To use XCConfig variables?

RadixWallet/Clients/CloudBackupClient/CloudBackupClient+Live.swift
```diff
@@ -0,0 +1,153 @@
...
+			let data = try Data(contentsOf: fileURL)
+			let profile = try Profile(jsonData: data)
+			try FileManager.default.removeItem(at: fileURL)
+
+			return profile
+		}
+
+		@discardableResult
+		@Sendable
+		func saveProfile(_ profile: Profile, existingRecord: CKRecord?) async throws -> CKRecord {
```
Sajjon, line 81: I think it would be more clear to name it `uploadProfileToCloud` which is what we are doing. At least while reading code in this file, `saveProfile` could have meant `downloadProfileToLocalKeychain`.

RadixWallet/Clients/CloudBackupClient/CloudBackupClient+Live.swift
kugel3: Yes I suppose that is the best way

RadixWallet/Clients/CloudBackupClient/CloudBackupClient+Live.swift
kugel3: Yes `save` is a bit confusing, but since the client is called `CloudBackupClient`, it is perhaps enough to name it `backupProfile`.

RadixWallet/Clients/SecureStorageClient/SecureStorageClient+Interface.swift
Sajjon: It is more a product owner question... But I think we should NEVER delete mnemonics without user knowing it. So safest to just delete profile and keep mnemonics. 

This can, however, lead to "dangling mnemonics", given current impl in Swift. We we ought to later see if we can changed that, because we only ever do the query "load EVERYTHING" in the `DebugKeychainContents` debug screen. Since otherwise, it is only a profile knowing the exact lookup-key (`FactorSourceID`) of a FactorSource (Mnemonic).

RadixWallet/Clients/SecurityCenterClient/SecurityCenterClient+Interface.swift
```diff
@@ -0,0 +1,50 @@
...
+		case .problem7: 7
+		case .problem9: 9
+		}
+	}
+}
+
+// MARK: - SecurityCenterClient.BackupStatus
+extension SecurityCenterClient {
+	// MARK: - BackupStatus
+	public struct BackupStatus: Codable, Sendable {
```
Sajjon, line 45: move `BackupStatus` to Sargon?

RadixWallet/Clients/SecurityCenterClient/SecurityCenterClient+Live.swift
```diff
@@ -0,0 +1,93 @@
...
+			await combineLatest(profileStore.values(), results).map { profile, backup in
+				let upToDate = backup.profileHash == profile.hashValue
+				let success = backup.result == .success
+				return .init(backupDate: backup.backupDate, upToDate: upToDate, success: success)
+			}
+			.eraseToAnyAsyncSequence()
+		}
+
+		return .init(
+			problems: {
```
Sajjon, line 41: The logic here too ought to done in Sargon IMO.

RadixWallet/Clients/SecureStorageClient/SecureStorageClient+Interface.swift
kugel3: For context, this is called after migrating the profile backup to icloud.

CyonAlexRDX (approved):
I don't know what has been previously agreed with Ghenadie, but there are a bunch of new undocumented and untested models and functions, which imo should be moved into Sargon: and be documented and tested, and both iOS and Android benefits from shared, tested, documented modes and logic. 

But maybe this is the last PR with models and logic and Swift and after the Rust course you will move it into Rust? 

Contents looks good otherwise (except lack of documentation)

RadixWallet/Clients/SecurityCenterClient/SecurityCenterClient+Interface.swift
```diff
@@ -0,0 +1,50 @@
...
+// MARK: SecurityCenterClient.Problems
+extension SecurityCenterClient {
+	public typealias Problems = @Sendable () async -> AnyAsyncSequence<[SecurityProblem]>
+	public typealias LastManualBackup = @Sendable () async -> AnyAsyncSequence<BackupStatus>
+	public typealias LastCloudBackup = @Sendable () async -> AnyAsyncSequence<BackupStatus>
+}
+
+// MARK: - SecurityProblem
+/// As outlined in https://radixdlt.atlassian.net/wiki/spaces/AT/pages/3392569357/Security-related+Problem+States+in+the+Wallet
+public enum SecurityProblem: Hashable, Sendable, Identifiable {
```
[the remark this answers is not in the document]
kugel3: As long as I can use it in swift. But when will that be available?

RadixWallet/Clients/SecurityCenterClient/SecurityCenterClient+Interface.swift
kugel3: It's iOS specific, so I am not sure of the benefit of that

RadixWallet/Clients/SecurityCenterClient/SecurityCenterClient+Live.swift
kugel3: I am not sure I see the benefit of that, this uses iOS specific APIs like user defaults and the logic is specific to iCloud backups. Also, can uniffi expose AsyncSequences?

RadixWallet/Clients/SecureStorageClient/SecureStorageClient+Interface.swift
matiasbzurovski: but if it isn't used on new Security Center, why are we exposing the `deleteProfile` as public?

RadixWallet/Clients/SecurityCenterClient/SecurityCenterClient+Interface.swift
```diff
@@ -0,0 +1,49 @@
...
+// MARK: SecurityCenterClient.Problems
+extension SecurityCenterClient {
+	public typealias Problems = @Sendable () async -> AnyAsyncSequence<[SecurityProblem]>
+	public typealias LastManualBackup = @Sendable () async -> AnyAsyncSequence<BackupStatus>
+	public typealias LastCloudBackup = @Sendable () async -> AnyAsyncSequence<BackupStatus>
+}
+
+// MARK: - SecurityProblem
+public enum SecurityProblem: Hashable, Sendable, Identifiable {
+	case problem3(accounts: Int, personas: Int)
```
matiasbzurovski, line 23:
```suggestion
	/// User has at least one shield, but has no XRD. User has an account (or accounts) without a shield applied, meaning it is only secured by a single factor.
	/// That factor (as is often, but not always the case) is a phone key factor and the user has never viewed the seed phrase and confirmed that they wrote it down.
	/// Losing the phone means losing that account forever.
	case problem3(accounts: Int, personas: Int)
```

This is what I meant @kugel3

RadixWallet/Clients/SecureStorageClient/SecureStorageClient+Interface.swift
kugel3: It is used [in CloudBackupClient](https://github.com/radixdlt/babylon-wallet-ios/pull/1106/files#diff-c4ae62b3f1ebe8cd50673c681cd202bc3b22760a37e90016f7781f9fd94acdc8R126)

RadixWallet/Clients/SecurityCenterClient/SecurityCenterClient+Interface.swift
kugel3: I know, but it's better to link it, since that Confluence page might update in the future

RadixWallet/Clients/SecureStorageClient/SecureStorageClient+Interface.swift
matiasbzurovski: so we need to consult with product if it is desired to keep mnemonics on that delete?

RadixWallet/Clients/BackupsClient/BackupsClient+Interface.swift
matiasbzurovski: Yes but this is a new method, why add it with an outdated name (`Snapshot`) when we can have it correctly?

CyonAlexRDX:
> Is the plan to stop writing Swift? I thought the point of Sargon was to put shared core logic there, not all logic? In many cases it will be a lot quicker to write things in Swift.

Let's merge it as is, implementing logic in Swift, and let's migrate later..,

RadixWallet/Clients/BackupsClient/BackupsClient+Interface.swift
kugel3: Does it matter? The purpose of this PR is too let you build your stuff on top of it.

kugel3: The import function still uses "Snapshot"

RadixWallet/Clients/SecurityCenterClient/SecurityCenterClient+Interface.swift
matiasbzurovski: Right, but if that confluence page updates and the implementation doesn't, then the linked Confluence will have logic that doesn't correspond to what we have on code. That's why I believe it is better to keep the link on top, but also reference on each case so that any developer can see what `problemX` means in code.

RadixWallet/Clients/BackupsClient/BackupsClient+Interface.swift
matiasbzurovski: yeah, just wanted to avoid merging code that we already want to change.. but we can fix later

matiasbzurovski (approved):

PR #1117, opened 2024-05-14
Fix P2P links loading

danvleju-rdx:
Fix for those devices that have created P2P links from `pre-alpha 1.6.0 (3)` build.
Solution is to ignore `loadP2PLinks` decoding error when updating P2P links.


CyonAlexRDX (approved):
LGTM

PR #1118, opened 2024-05-14
[ABW-3309] Factory Reset + Settings fixes

matiasbzurovski:
Jira ticket: [ABW-3309](https://radixdlt.atlassian.net/browse/ABW-3309)
Jira ticket: [ABW-3311](https://radixdlt.atlassian.net/browse/ABW-3311)
Jira ticket: [ABW-3319](https://radixdlt.atlassian.net/browse/ABW-3311)

## Description
This PR takes care of the following items:
- Add `Factory Reset` flow under `Troubleshooting` section in Settings.
- Allow `Import From Legacy Wallet` flow only on mainnet.
- Add missing warnings to Settings views based on `SecurityCenterClient`.

## Video

https://github.com/radixdlt/babylon-wallet-ios/assets/164921079/4553063b-c20e-4e2a-a94c-27b166df8e4c

[ABW-3309]: https://radixdlt.atlassian.net/browse/ABW-3309?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[ABW-3311]: https://radixdlt.atlassian.net/browse/ABW-3311?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[ABW-3319]: https://radixdlt.atlassian.net/browse/ABW-3319?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


RadixWallet/Clients/DeviceFactorSourceClient/DeviceFactorSourceClient+Live.swift
```diff
@@ -79,6 +79,22 @@ extension DeviceFactorSourceClient: DependencyKey {
 			return entities.contains(where: { !$0.isMnemonicPresentInKeychain })
 		}
 
+		let unrecoverableEntitiesCount: @Sendable () async throws -> (accounts: Int, personas: Int) = {
```
matiasbzurovski, line 82: This may go away if we don't need to be smart about amount of entities that are unrecoverable. Slack [thread](https://rdxworks.slack.com/archives/C03QFAWBRNX/p1715679454159999?thread_ts=1715353183.606979&cid=C03QFAWBRNX)

RadixWallet/Features/SettingsFeature/Preferences/Preferences+View.swift
```diff
@@ -68,7 +68,7 @@ extension Preferences.View {
 					#endif
 				}
 			}
-			.background(Color.app.gray4)
+			.background(Color.app.gray5)
```
matiasbzurovski, line 71: Background colors in Settings views were incorrect, so updated it in every screen.

CyonAlexRDX (approved):
Looks great! Good job!

RadixWallet/Clients/SecurityCenterClient/SecurityCenterClient+Interface.swift
```diff
@@ -50,6 +49,16 @@ public enum SecurityProblem: Hashable, Sendable, Identifiable {
...
+
+	public var message: String {
+		switch self {
+		case .problem3: L10n.SecurityCenter.Problem3.text
+		case .problem5: L10n.SecurityCenter.Problem5.text
+		case .problem6: L10n.SecurityCenter.Problem6.text
+		case .problem7: L10n.SecurityCenter.Problem7.text
+		case .problem9: L10n.SecurityCenter.Problem9.text
+		}
+	}
```
kugel3, line 61: This should not be here, it's view related. I think we already have it there.

RadixWallet/Clients/SecurityCenterClient/SecurityCenterClient+Live.swift
```diff
@@ -68,6 +72,11 @@ extension SecurityCenterClient {
 						await (try? deviceFactorSourceClient.isSeedPhraseNeededToRecoverAccounts()) ?? false
 					}
 
+					let problem3 = await problem3()
```
kugel3, line 75: I would use the same style as the others, `hasProblem3`, but this one can return an optional tuple. So it will be used like `if let (personas, accounts) = hasProblem3() { ... }`

RadixWallet/Features/SettingsFeature/Settings+Reducer.swift
```diff
@@ -175,4 +183,13 @@ extension Settings {
 			)))
 		}
 	}
+
+	private func loadSecurityProblems() -> Effect<Action> {
```
kugel3, line 187: I'd call this securityProblemsEffect

RadixWallet/Features/SettingsFeature/Settings+Reducer.swift
```diff
@@ -175,4 +183,13 @@ extension Settings {
...
+		.run { send in
+			for try await problems in await securityCenterClient.problems() {
+				guard !Task.isCancelled else { return }
+				await send(.internal(.loadedSecurityProblems(problems)))
```
kugel3, line 191: and this one I'd call setSecurityProblems

RadixWallet/Clients/SecurityCenterClient/SecurityCenterClient+Interface.swift
matiasbzurovski: It is used in multiple views, not just in one.

RadixWallet/Clients/SecurityCenterClient/SecurityCenterClient+Live.swift
matiasbzurovski: In that case it would be
``` Swift
if let (personas, accounts) = hasProblem3(), (personas.count + accounts.count) > 0 {
}
```

meaning the `hasProblem3()` doesn't actually mean that it HAS the problem.

RadixWallet/Features/SettingsFeature/Settings+Reducer.swift
matiasbzurovski: I think it is more descriptive with the `load` prefix, plus seems to be the standard followed
Screenshot 2024-05-15 at 15 44 22

RadixWallet/Clients/SecurityCenterClient/SecurityCenterClient+Live.swift
kugel3: no, as I said I think the tuple should be optional, and return nil when the problem is not present. So the check would be inside hasProblem3

RadixWallet/Features/SettingsFeature/Settings+Reducer.swift
kugel3: It isn't loading anything though, and we already have exactly these mechanics in many other places, so we should use the same naming conventions. Please model it on SecurityCenter+Reducer, where I do exactly the same thing.

matiasbzurovski: Updated both methods from this class, but to me the `securityProblemsEffect` isn't descriptive enough. We could call it `read` instead of `load` if you want, but the name of the method should describe what's actually happening.

kugel3: The convention is that for streams we just use the plural of the thing that the stream produces. This is based on the fact that apple explicitly discourages using things like "get..." as prefix for async methods.

kugel3: So load or read would make sense for something you fetch once, or at least something you fetch using "pull" logic. For something that is "push", like a stream, load does not make much sense in my mind, and in any case we started using the convention above so we better stick to it.

PR #1121, opened 2024-05-15
[CAP-36] Accounts sharing fix

danvleju-rdx:
Fixes the scenarios when `sendAccountListMessage` throws an error and stops the execution of the loop.


CyonAlexRDX (approved):
LGTM

PR #1126, opened 2024-05-17
[ABW-3238] Fix navigation bar titles

matiasbzurovski:
Jira ticket: [ABW-3238](https://radixdlt.atlassian.net/browse/ABW-3238)
Jira ticket: [ABW-3239](https://radixdlt.atlassian.net/browse/ABW-3239)

## Description
This task started only as a fix to the situation described on [ABW-3238](https://radixdlt.atlassian.net/browse/ABW-3238), where some views wouldn't show the navigation bar title after certain steps. It ended up being a unification of the font and color used for every navigation bar in the app.

| - | - |
| <video src=https://github.com/radixdlt/babylon-wallet-ios/assets/164921079/82bbf301-8e13-4f0e-881f-eb8e45c66178> | <video src=https://github.com/radixdlt/babylon-wallet-ios/assets/164921079/d838bb98-110e-4f48-a70d-15d2ebf2961b> |

[ABW-3238]: https://radixdlt.atlassian.net/browse/ABW-3238?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[ABW-3239]: https://radixdlt.atlassian.net/browse/ABW-3239?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[ABW-3238]: https://radixdlt.atlassian.net/browse/ABW-3238?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


RadixWallet/Core/DesignSystem/ViewModifiers/NavigationBarAppearenceViewModifier.swift
```diff
@@ -0,0 +1,39 @@
+import SwiftUI
+
+extension View {
+	func configureNavigationBarAppearence() -> some View {
```
kugel3, line 4: I am really not sure that we want to do UIKit magic for this, we are trying move away from non-standard UI solutions.

matiasbzurovski: I also don't love this, but SwiftUI doesn't have a way of setting a default back button for every view. Only way to do it would be to manually add it on every view, and that would also involve checking for the case we want a close button on the leading bar (and should not add the back button). 

Also, this PR moves away from some UIKit + Introspect magic so I still consider it a win in that side

Screenshot 2024-05-17 at 15 07 30

CyonAlexRDX (approved):
LGTM, I agree with Gustaf that the goal in general is to limit UIKit use, but this PR looks good and hopefully Apple can allow more customisation of NavBar with SwiftUI this WWDC? Fingers crossed 🤞

matiasbzurovski:
hopefully they do, although we would only be able to use it once N and N-1 iOS versions support it, so we probably won't have the custom ⬅️ back icon for a while.

RadixWallet/Core/DesignSystem/Extensions/View+Extra.swift
```diff
@@ -9,21 +9,46 @@ extension View {
...
 					Text(title)
 						.foregroundColor(.app.gray1)
-						.textStyle(.secondaryHeader)
+						.textStyle(.body1Header)
 				}
 			}
 			.navigationBarTitleDisplayMode(.inline)
-			.toolbarBackground(.app.background, for: .navigationBar)
-			.toolbarBackground(.visible, for: .navigationBar)
+			.applyIf(setBackgroundColor) {
```
kugel3, line 22: Can't we just use the Visibility parameter for `toolbarBackground` to enable/disable it, rather than `applyIf`?

kugel3 (approved):
Great to simplify and improve this stuff

PR #1131, opened 2024-05-22
[CAP-36] Fix P2P links connection

danvleju-rdx:
## Description
Fixes the issue when P2P links were not connected after creating a new profile.


RadixWallet/Features/CreateAccount/Coordinator/CreateAccountCoordinator+Reducer.swift
```diff
@@ -179,6 +180,7 @@ extension CreateAccountCoordinator {
 			return .run { send in
 				await send(.internal(.handleAccountCreated(TaskResult {
 					try await accountsClient.saveVirtualAccount(account)
+					_ = await radixConnectClient.loadP2PLinksAndConnectAll()
```
CyonAlexRDX, line 183: what is the rationale for loading and connecting here?

danvleju-rdx: We need to connect links after a new profile is created, but I realize now that this might not be a good place to do it.
Maybe a better place will be in `OnboardingCoordinator` -> `.createAccountCoordinator(.delegate(.completed))`?

CyonAlexRDX (approved):
LGTM

PR #1132, opened 2024-05-22
[ABW-3328] Fix NPSSurvey presentation bug

matiasbzurovski:
Jira ticket: [ABW-3328](https://radixdlt.atlassian.net/browse/ABW-3328?atlOrigin=eyJpIjoiM2FlOWQwMWRlZDg0NDgxOGExZTA3MDE0OTlmZmFiYjYiLCJwIjoiaiJ9)

## Problem
Until now, `Home` was in charge of presenting the `NPSSurvey` view whenever the conditions were met (10th successful transaction or 3 months since last time shown). However, if the `DappInteractor` was currently presenting its flow on the root view, we could have an overlap of modal presentations. This causes unexpected behaviors on SwiftUI, by presenting only one of the two modals and leaving modal presentation broken during the app run. 

### Solution
Trigger the NPS Survey counter increment after completion is dismissed.

## How to test
You can either run on a fresh install and perform 10 transactions, or just update the logic under `NPSSurveyClient+Live`:
```Swift
static func shouldAskUserForFeedback(_ transactionCounter: Int, dateOfLastSubmittedNPSSurvey: Date?) -> Bool {
	return true
}
```

## Bug before & after
You can see on the first video how the completion view wasn't presented, while on the second video both views (NPSSurvey & Completion) are presented sequentially.

| Before | After |
| - | - |
| <video src=https://github.com/radixdlt/babylon-wallet-ios/assets/164921079/ad202ace-b49a-4d08-ad39-ad8ae2f1a06f> | <video src=https://github.com/radixdlt/babylon-wallet-ios/assets/164921079/78d77490-171e-4277-99d4-f666604e7426> |

[ABW-3328]: https://radixdlt.atlassian.net/browse/ABW-3328?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/RadixWallet/Features/DappInteractionFeature/Interactor/DappInteractor.swift
+++ b/RadixWallet/Features/DappInteractionFeature/Interactor/DappInteractor.swift
@@ -19,6 +19,8 @@ struct DappInteractor: Sendable, FeatureReducer {
 
 		@PresentationState
 		public var destination: Destination.State?
+
+		fileprivate var shouldIncrementOnCompletionDismiss = false
 	}
 
 	enum ViewAction: Sendable, Equatable {
@@ -128,7 +130,9 @@ struct DappInteractor: Sendable, FeatureReducer {
 			}
 
 		case .completionDismissed:
-			npsSurveyClient.incrementTransactionCompleteCounter()
+			if state.shouldIncrementOnCompletionDismiss {
+				npsSurveyClient.incrementTransactionCompleteCounter()
+			}
 			return .none
 		}
 	}
@@ -209,6 +213,7 @@ struct DappInteractor: Sendable, FeatureReducer {
 			return .none
 
 		case let .presentResponseSuccessView(dappMetadata, txID, p2pRoute):
+			state.shouldIncrementOnCompletionDismiss = txID != nil
 			state.destination = .dappInteractionCompletion(
 				.init(
 					txID: txID,
--- a/RadixWallet/Features/DappInteractionFeature/Interactor/DappInteractor.swift
+++ b/RadixWallet/Features/DappInteractionFeature/Interactor/DappInteractor.swift
@@ -31,6 +31,7 @@ struct DappInteractor: Sendable, FeatureReducer {
 		case task
 		case moveToBackground
 		case moveToForeground
+		case completionDismissed
 		case responseFailureAlert(PresentationAction<ResponseFailureAlertAction>)
 		case invalidRequestAlert(PresentationAction<InvalidRequestAlertAction>)
 
@@ -106,6 +107,7 @@ struct DappInteractor: Sendable, FeatureReducer {
 	@Dependency(\.rolaClient) var rolaClient
 	@Dependency(\.appPreferencesClient) var appPreferencesClient
 	@Dependency(\.dappInteractionClient) var dappInteractionClient
+	@Dependency(\.npsSurveyClient) var npsSurveyClient
 
 	var body: some ReducerOf<Self> {
 		Reduce(core)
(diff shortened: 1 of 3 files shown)


RadixWallet/Features/DappInteractionFeature/Interactor/DappInteractor.swift
```diff
@@ -18,30 +18,32 @@ struct DappInteractor: Sendable, FeatureReducer {
...
-		case invalidRequestAlert(PresentationAction<InvalidRequestAlertAction>)
-
-		enum ResponseFailureAlertAction: Sendable, Hashable {
-			case cancelButtonTapped(RequestEnvelope)
-			case retryButtonTapped(P2P.Dapp.Response, for: RequestEnvelope, DappMetadata)
-		}
-
-		enum InvalidRequestAlertAction: Sendable, Hashable {
-			case ok(P2P.RTCOutgoingMessage.Response, origin: P2P.Route)
-		}
```
matiasbzurovski, line 44: Moved into `Destination`

RadixWallet/Features/DappInteractionFeature/Interactor/DappInteractor.swift
```diff
@@ -106,52 +123,30 @@ struct DappInteractor: Sendable, FeatureReducer {
...
-			case let .presented(.ok(response, route)):
-				return .run { send in
-					do {
-						try await dappInteractionClient.completeInteraction(.response(response, origin: route))
-					} catch {
-						errorQueue.schedule(error)
-					}
-					await send(.internal(.presentQueuedRequestIfNeeded))
-				}
-			}
```
matiasbzurovski, line 147: Moved into `func reduce(into state: inout State, presentedAction: Destination.Action)`
[link](https://github.com/radixdlt/babylon-wallet-ios/pull/1132/files#diff-e0da21a3c53a16216da44c31e8e799e020a72c57b6f04d332e0c20306cc49711R284)

danvleju-rdx (approved):
LGTM, but I would wait for @GhenadieVP's input, as he worked on the NPS survey.
 Perhaps there was an intentional reason for not placing the NPS survey in `DappInteractor`.

RadixWallet/Features/DappInteractionFeature/Interactor/DappInteractor.swift
```diff
@@ -483,4 +525,28 @@ extension DappInteractor {
...
+		.run { send in
+			for try await shouldShow in await npsSurveyClient.shouldAskForUserFeedback() {
+				guard !Task.isCancelled else { return }
+				await send(.internal(.shouldShowNpsSurvey(shouldShow)))
+			}
+		}
+	}
+
+	private func uploadUserFeedbackEffect(_ feedback: NPSSurveyClient.UserFeedback?) -> Effect<Action> {
+		if feedback != nil {
```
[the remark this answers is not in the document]
CyonAlexRDX: I think this is a good idea. But maybe just check with Matt! If you have not done so already?

CyonAlexRDX (approved):
Nice! LGTM! Good job!

RadixWallet/Features/DappInteractionFeature/Interactor/DappInteractor.swift
```diff
@@ -156,18 +126,22 @@ struct DappInteractor: Sendable, FeatureReducer {
 			return .run { _ in
 				_ = await radixConnectClient.loadP2PLinksAndConnectAll()
 			}
+
+		case .completionDismissed:
+			npsSurveyClient.incrementTransactionCompleteCounter()
```
GhenadieVP, line 131: Note that just `completionDismissed` is not enough, you need to also check that the completion is for a transaction, and not any other interaction(like Login).

You could potentially make use of the presence of `txId` from the completion state.

GhenadieVP (changes requested):
Looks good, but the condition of increment needs to be updated.

GhenadieVP (approved):

PR #1133, opened 2024-05-23
Show fiat worth error state when price retrieval fails.

GhenadieVP:
Jira ticket: https://radixdlt.atlassian.net/browse/ABW-3336

## Description
Previously the Wallet considered that there is no token prices if the retrieval of these failed. With this PR the Wallet will show error state when the Wallet fails to fetch token prices.

## Video

https://github.com/radixdlt/babylon-wallet-ios/assets/118184705/7aca69f1-4d17-4f7d-b89c-9fd7ab14e835


CyonAlexRDX (approved):
Nice!

danvleju-rdx (approved):

matiasbzurovski (approved):

PR #1136, opened 2024-05-24
Fix FactoryReset to not remove CloudKit related cache

GhenadieVP:
Jira ticket: https://radixdlt.atlassian.net/browse/ABW-3342

## Description
Our CacheClient did create all of the cache folders/files at the root of `caches` directory, thus when it was needed to clear the cache on FactoryReset, the whole `caches` directory was cleared out.

CloudKit is using `caches` directory for its operations, and it would create specific folders for its operations. Thus when performing FactoryReset those were deleted - this caused the bug of not being able to see the Cloud backups when restoring from backup immediately after FactoryReset.

The error thrown by CloudKit was:
```
CKError 0x302a87e10: "Internal Error" (1/1000); "Moving downloaded asset failed"
```
Which correlates well with the CloudKit directories being removed from `caches` - CloudKit didn't know anymore where to store the temporary assets after downloading those from the container.

The fix is about moving all of the CacheClient's entries under one and the same folder, thus when doing factory reset and calling `cacheClient.removeAll()`, only the entries created by the CacheClient directly are removed.

## Video

https://github.com/radixdlt/babylon-wallet-ios/assets/118184705/872963a4-59a0-4abe-8e31-4627ca415baf

### DIFF
--- a/RadixWallet/Clients/CacheClient/CacheClient+Live.swift
+++ b/RadixWallet/Clients/CacheClient/CacheClient+Live.swift
@@ -61,7 +61,7 @@ extension CacheClient: DependencyKey {
 			@Dependency(\.diskPersistenceClient) var diskPersistenceClient
 
 			do {
-				try diskPersistenceClient.removeAll()
+				try diskPersistenceClient.remove(Entry.root)
 				loggerGlobal.debug("💾 Data successfully cleared from disk")
 			} catch {
 				loggerGlobal.debug("💾❌ Could not clear cached data from disk: \(error.localizedDescription)")
--- a/RadixWallet/Clients/CacheClient/CacheClient+Interface.swift
+++ b/RadixWallet/Clients/CacheClient/CacheClient+Interface.swift
@@ -109,6 +109,8 @@ extension Address {
 
 extension CacheClient {
 	public enum Entry: Equatable {
+		static let root: String = "RadixWallet"
+
 		public init(address: some AddressProtocol) {
 			self = .onLedgerEntity(.init(address: address))
 		}
@@ -161,7 +163,7 @@ extension CacheClient {
 		case dateOfFirstTransaction(_ accountAddress: AccountAddress)
 
 		var filesystemFilePath: String {
-			switch self {
+			let path = switch self {
 			case let .onLedgerEntity(entity):
 				entity.filesystemFilePath
 			case let .networkName(url):
@@ -177,6 +179,8 @@ extension CacheClient {
 			case let .dateOfFirstTransaction(address):
 				"\(filesystemFolderPath)/account-\(address.address)"
 			}
+
+			return "\(Self.root)/\(path)"
 		}
 
 		var filesystemFolderPath: String {


RadixWallet/Clients/CloudBackupClient/CloudBackupClient+Live.swift
```diff
@@ -152,8 +152,6 @@ extension CloudBackupClient {
 					}
 
 					let savedRecord = try await uploadProfileSnapshotToICloud(profileSnapshot, id: profile.id, existingRecord: backedUpRecord)
-					// Migration completed, deleting old copy
```
GhenadieVP, line 155: @kugel3 Before merging this, I want to actually understand the reasoning behind deleting the profile from the secure storage.

IMO, the Wallet should be able to locate the profile in secure storage even after the "backup migration". To me, it is wrong to assume that backup migration means uploading the profile from the Keychain to the cloud container and then deleting it from the Keychain. 

The SecureStorage(KeyChain) is the actual local storage of the Profile, and the profile should be suddenly gone. A more correct interpretation of migration maybe is:
- Upload the profile from keychain to cloud.
- Disable keychain sync for the said profile.

Probably I am missing something as you have worked a lot more on this feature.

CyonAlexRDX (approved):
LGTM

RadixWallet/Clients/CacheClient/CacheClient+Live.swift
```diff
@@ -61,7 +61,7 @@ extension CacheClient: DependencyKey {
 			@Dependency(\.diskPersistenceClient) var diskPersistenceClient
 
 			do {
-				try diskPersistenceClient.removeAll()
+				try diskPersistenceClient.remove(Entry.root)
```
matiasbzurovski, line 64: Should we delete the `removeAll` from `DiskPersistenceClient`?

matiasbzurovski (approved):
Good catch!

RadixWallet/Clients/CacheClient/CacheClient+Live.swift
GhenadieVP: I did choose to still have it, in case we do really need to clear the whole `caches` directory.

RadixWallet/Clients/CloudBackupClient/CloudBackupClient+Live.swift
kugel3: I didn't see this comment before, but it was decided last week to not actually delete non-current profiles when they are "migrated". Matt still wants to delete the currently used profile from backup, but it just hit me that I am not sure we even can do that.

kugel3: Keep in mind that this code only applies to non-current profiles.

kugel3 (approved):

RadixWallet/Clients/CloudBackupClient/CloudBackupClient+Live.swift
GhenadieVP: Interesting that you mention non-current profiles. In the context of SelectBackup, aren't all profiles non-current? I don't see the migration function to be called from anywhere else than from SelectBackup.

I guess what we have to do to remove the backup from cloud, is to disable iCloud sync of the profile in SecureStorage, which we should do anyway.

PR #1143, opened 2024-05-24
Security center part 2

kugel3:
- Starts persisting the profile header as metadata, along with the serialized profile, so that we don't have to download the entire profile for every backed up profile
- Changes to using the default container for each target
- Makes alpha and pre-alpha testflights work
- Various minor fixes

## PR submission checklist


RadixWallet/Clients/CloudBackupClient/CloudBackupClient+Live.swift
```diff
@@ -8,81 +8,106 @@ extension CKRecord.RecordType {
...
-		@Sendable
-		func uploadProfileToICloud(_ profile: Profile, existingRecord: CKRecord?) async throws -> CKRecord {
-			try await uploadProfileSnapshotToICloud(profile.profileSnapshot(), id: profile.id, existingRecord: existingRecord)
+			return BackedUpProfile(profile: profile, containsLegacyP2PLinks: containsLegacyP2PLinks)
 		}
 
 		@discardableResult
 		@Sendable
-		func uploadProfileSnapshotToICloud(_ profileSnapshot: Data, id: ProfileID, existingRecord: CKRecord?) async throws -> CKRecord {
+		func uploadProfileToICloud(_ profile: Profile, existingRecord: CKRecord?) async throws -> CKRecord {
```
GhenadieVP, line 99: @kugel3 the reason for having previously to upload raw Data instead of Profile, is that decoding the snapshot from the Keychain during migration will delete P2PLinks information, as now Profile does not have it anymore.
So just by the effect of decoding and then encoding, P2PLinks will be lost.

kugel3: But if it's not part of the profile definition anymore, why would we want to keep that information?

GhenadieVP: so we present an alert to the users that Wallet now supports the new CE and  that they need to re link.

GhenadieVP: more precisely see `getProfile` in this file, that it checks for legacy P2PLinks

kugel3: Sure, but do we actually care about this for backups of old profiles, that are not your main profile?

GhenadieVP: In my understanding we should inform users that their previous P2PLinks are lost and they need to re-link whenever the Wallet detects that - be it on the currently profile, on the profile from the cloud backup or on the profile imported from a file.

Additionally, an upload of the Profile to cloud should ideally not loose any information. True, this is a very particular case when the decoded representation of the Profile does not match the underlying bytes, but we should let other parts of the codebase handle it, cloud backups should not make any assumptions on its behalf.

kugel3: > should ideally not loose any information
This should of course be true, but i would say it's very bad practice to rely on information being present in an object, that is no longer part of the definition of that object. Either we care about it, then it should be part of the public contract, or we don't, and then we should make no effort to keep it there, or rely on it being present.

>cloud backups should not make any assumptions on its behalf.
I don't agree with this. I think all parts of the code should be able to rely on the public schema of objects, data structures should not have hidden data.

If we care about this information, it should be kept around as a property in the profile, or alternatively stored in metadata. Those are the robust solutions.

Nevertheless, I refactored it a bit so that it does retain that information.

RadixWallet.xcodeproj/project.pbxproj
Sajjon, line 1: the code sign changes should be set via XCConfig file, so I think the per target changes relating to codesigning in pbxproj ought to be reverted?

RadixWallet/Features/SettingsFeature/DisplayMnemonics/Coordinator/DisplayMnemonics.swift
```diff
@@ -12,6 +12,8 @@ public struct DisplayMnemonics: Sendable, FeatureReducer {
 		public var destination: Destination.State? = nil
 
 		public var deviceFactorSources: IdentifiedArrayOf<DisplayEntitiesControlledByMnemonic.State> = []
+		fileprivate var problems: [SecurityProblem]?
+		fileprivate var entities: IdentifiedArrayOf<EntitiesControlledByFactorSource>?
```
Sajjon, line 16: I think `entities` is a misleading name. A collection of accounts, or a collection of personas or a mix of both those would be `entities`. This a collection of Tuples with `(FactorSource, Entities)`. 

`factorsWithEntitiesTheyControl` is more accurate and maybe not too long?

RadixWallet/RadixWalletDebug-Dev.entitlements
```diff
@@ -3,9 +3,7 @@
 <plist version="1.0">
 <dict>
 	<key>com.apple.developer.icloud-container-identifiers</key>
-	<array>
-		<string>iCloud.com.radixpublishing.radixwallet.ios.dev.cloudBackup</string>
-	</array>
+	<array/>
```
GhenadieVP, line 6: What is the inferred iCloud container identifier? given that this array is empty

RadixWallet.xcodeproj/project.pbxproj
kugel3: I can revert those and see if it still works, the whole thing seems a bit fragile.

RadixWallet/RadixWalletDebug-Dev.entitlements
kugel3: The default name, which is `iCloud.[bundle ID]`

RadixWallet/Features/ProfileBackupsFeature/RestoreProfileFromBackup/Children/SelectBackup/SelectBackup+Reducer.swift
```diff
@@ -215,38 +221,33 @@ public struct SelectBackup: Sendable, FeatureReducer {
...
-				do {
+			do {
+				if !userDefaults.getDidMigrateKeychainProfiles {
+					await send(.internal(.setStatus(.migrating)))
+
 					let profilesInKeychain = try secureStorageClient.loadProfileHeaderList()?.count ?? 0
 					if profilesInKeychain > 0 {
 						_ = try await cloudBackupClient.migrateProfilesFromKeychain()
-						userDefaults.setDidMigrateKeychainProfiles(true)
+						// userDefaults.setDidMigrateKeychainProfiles(true)
```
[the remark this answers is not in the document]
kugel3: yes

fastlane/Fastfile
```diff
@@ -120,6 +120,12 @@ platform :ios do
         code_signing(type: "development", readonly: false)
     end
 
+    desc "Generate new alpha certificates"
+    lane :generate_new_alpha_certificates do
+        Dotenv.load ".env.ios.alpha"
+        code_signing(type: "development", readonly: false)
```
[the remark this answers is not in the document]
kugel3: testflight, but that was before i realise that testflight used release. either way, this lane is not needed

RadixWallet.xcodeproj/project.pbxproj
GhenadieVP: @kugel3 I even think that no changes at all should be done to code signing, your PR should just add the required capabilities for CloudKit.

RadixWallet/RadixWalletDebug-Dev.entitlements
GhenadieVP: then why Pre-Alpha and Alpha have the iCloud container explicitly specified? why the difference?

RadixWallet.xcodeproj/project.pbxproj
kugel3: But isn't it an advantage to be able to run those targets locally?

RadixWallet/RadixWalletDebug-Dev.entitlements
kugel3: it makes no difference when running it, except that in some cases I had to add it explicitly for xcode  to pick up the changes

RadixWallet.xcodeproj/project.pbxproj
GhenadieVP: I don't get how is this related to the changes, the XCConfig files are used by Xcode also when running locally

RadixWallet/RadixWalletDebug-Dev.entitlements
GhenadieVP: I would opt to be consistent, either the container are explicitly specified or not, I more for explicitly specifying these.

RadixWallet/Features/SettingsFeature/DisplayMnemonics/Coordinator/DisplayMnemonics.swift
matiasbzurovski: then we should rename `EntitiesControlledByFactorSource` as well?

RadixWallet/RadixWalletDebug-Dev.entitlements
kugel3: sure, already did that

RadixWallet.xcodeproj/project.pbxproj
kugel3: I just presumed that things would not work when xcode was showing errors, but:

- Now when I manually remove these changes from the text file and launch xcode it doesn't show any errors
- It seems to show the same profile names anyway
- It doesn't seem to be able to deploy directly to device in either case

GhenadieVP (approved):
LGTM!

PR #1147, opened 2024-05-28
Update Seed Phrases to use Security Client

matiasbzurovski:
## Changelog
- Uses `SecurityCeneterClient` to determine problems inside `DisplayMnemonics`
- Update UI to look as Aftab requested
- Consider hidden accounts for problems.

## Screenshots

| - | - |
| Simulator Screenshot - iPhone 15 Pro - 2024-05-28 at 16 15 51 | problem3 after |
| problem9 before | problem9 after |

### DIFF
--- a/RadixWallet/Clients/DeviceFactorSourceClient/DeviceFactorSourceClient+Interface.swift
+++ b/RadixWallet/Clients/DeviceFactorSourceClient/DeviceFactorSourceClient+Interface.swift
@@ -13,7 +13,7 @@ public struct DeviceFactorSourceClient: Sendable {
 	public var controlledEntities: GetControlledEntities
 
 	/// The `Accounts` & `Personas` for which the wallet doesn't have its mnmemonic.
-	public var missingMnemonicEntities: MissingMnemonicEntities
+	public var mnemonicMissingEntities: MnemonicMissingEntities
 
 	/// The `Accounts` & `Personas` the user wouldn't be able to recover if they loose their phone,
 	/// since they haven't been backed up (seed phrase not written).
@@ -25,7 +25,7 @@ public struct DeviceFactorSourceClient: Sendable {
 		isAccountRecoveryNeeded: @escaping IsAccountRecoveryNeeded,
 		entitiesControlledByFactorSource: @escaping GetEntitiesControlledByFactorSource,
 		controlledEntities: @escaping GetControlledEntities,
-		missingMnemonicEntities: @escaping MissingMnemonicEntities,
+		mnemonicMissingEntities: @escaping MnemonicMissingEntities,
 		unrecoverableEntities: @escaping UnrecoverableEntities
 	) {
 		self.publicKeysFromOnDeviceHD = publicKeysFromOnDeviceHD
@@ -33,7 +33,7 @@ public struct DeviceFactorSourceClient: Sendable {
 		self.isAccountRecoveryNeeded = isAccountRecoveryNeeded
 		self.entitiesControlledByFactorSource = entitiesControlledByFactorSource
 		self.controlledEntities = controlledEntities
-		self.missingMnemonicEntities = missingMnemonicEntities
+		self.mnemonicMissingEntities = mnemonicMissingEntities
 		self.unrecoverableEntities = unrecoverableEntities
 	}
 }
@@ -46,7 +46,7 @@ extension DeviceFactorSourceClient {
 	public typealias PublicKeysFromOnDeviceHD = @Sendable (PublicKeysFromOnDeviceHDRequest) async throws -> [HierarchicalDeterministicPublicKey]
 	public typealias SignatureFromOnDeviceHD = @Sendable (SignatureFromOnDeviceHDRequest) async throws -> SignatureWithPublicKey
 	public typealias IsAccountRecoveryNeeded = @Sendable () async throws -> Bool
-	public typealias MissingMnemonicEntities = @Sendable () async throws -> (accounts: [AccountAddress], personas: [IdentityAddress])
+	public typealias MnemonicMissingEntities = @Sendable () async throws -> (accounts: [AccountAddress], personas: [IdentityAddress])
 	public typealias UnrecoverableEntities = @Sendable () async throws -> (accounts: [AccountAddress], personas: [IdentityAddress])
 }
 
--- a/RadixWallet/Clients/DeviceFactorSourceClient/DeviceFactorSourceClient+Interface.swift
+++ b/RadixWallet/Clients/DeviceFactorSourceClient/DeviceFactorSourceClient+Interface.swift
@@ -12,8 +12,8 @@ public struct DeviceFactorSourceClient: Sendable {
 	/// Fetched accounts and personas on current network that are controlled by a device factor source, for every factor source in current profile
 	public var controlledEntities: GetControlledEntities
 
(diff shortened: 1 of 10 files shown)


RadixWallet/Clients/DeviceFactorSourceClient/DeviceFactorSourceClient+Interface.swift
```diff
@@ -46,7 +46,7 @@ extension DeviceFactorSourceClient {
 	public typealias PublicKeysFromOnDeviceHD = @Sendable (PublicKeysFromOnDeviceHDRequest) async throws -> [HierarchicalDeterministicPublicKey]
 	public typealias SignatureFromOnDeviceHD = @Sendable (SignatureFromOnDeviceHDRequest) async throws -> SignatureWithPublicKey
 	public typealias IsAccountRecoveryNeeded = @Sendable () async throws -> Bool
-	public typealias IsSeedPhraseNeededToRecoverAccounts = @Sendable () async throws -> Bool
+	public typealias MissingMnemonicEntities = @Sendable () async throws -> (accounts: [AccountAddress], personas: [IdentityAddress])
```
kugel3, line 49: It might be clearer if you reverse the name, to `EntitiesMissingMnemonic`. The current name is ambiguous.

RadixWallet/Clients/DeviceFactorSourceClient/DeviceFactorSourceClient+Live.swift
```diff
@@ -89,7 +100,9 @@ extension DeviceFactorSourceClient: DependencyKey {
 			for entity in entities {
```
kugel3, line 100: Did you consider something more functional, like

```
			let entities = try await deviceFactorSources.asyncMap {
				try await entitiesControlledByFactorSource($0, nil)
			}
			.filter { !$0.isMnemonicMarkedAsBackedUp }

			let accounts = entities.flatMap { $0.accounts + $0.hiddenAccounts }.map(\.address)
			let accounts = entities.flatMap { $0.personas + $0.hiddenPersonas }.map(\.address)

			return (accounts, personas)
```

RadixWallet/Clients/DeviceFactorSourceClient/DeviceFactorSourceClient+Interface.swift
matiasbzurovski: we are using `adjectiveNoun` for the rest of the cases:
```
controlledEntities
unrecoverableEntities
```

I would keep it this way or change all.

RadixWallet/Features/DisplayEntitiesControlledByMnemonic/DisplayEntitiesControlledByMnemonic+Reducer.swift
```diff
@@ -102,3 +103,33 @@ public struct DisplayEntitiesControlledByMnemonic: Sendable, FeatureReducer {
...
+	func isMnemonicMarkedAsBackedUp(accounts: [Account]) -> Bool {
+		!contains(where: { item in
+			switch item {
+			case let .problem3(problematicAccounts, _):
+				guard let first = accounts.first else {
+					return false
+				}
+				return problematicAccounts.contains(where: { $0.id == first.address.id })
+			default:
+				return false
```
kugel3, line 117: I would probably make an extension on a single problem, and delegate to that. But this seems to only consider the first of the supplied accounts, is that on purpose? Also, why are we comparing ids, rather than addresses? I would expect something like this to be correct:
```
private extension [SecurityProblem] {
	func isMnemonicMarkedAsBackedUp(accounts: [Account]) -> Bool {
		contains { $0.isMnemonicMarkedAsBackedUp(accounts: accounts) }
	}
}

private extension SecurityProblem {
	func isMnemonicMarkedAsBackedUp(accounts: [Account]) -> Bool {
		switch self {
		case let .problem3(problematicAccounts, _):
			Set(problematicAccounts.map(\.address)).isDisjoint(with: accounts.map(\.address))
		default:
			return true
		}
	}
}
```

kugel3: By the way, why does it take multiple accounts? Is the idea that we call this with all the accounts in a profile?

kugel3 (approved):
Just some suggestions

RadixWallet/Clients/DeviceFactorSourceClient/DeviceFactorSourceClient+Interface.swift
kugel3: Well, that's because in English you usually put adjectives in front of nouns. But "missing mnemonics" is a present participle used as an attribute, and unlike adjective attributes they almost always go before the noun when the verb has an object. 

You wouldn't say "there are three eating ice cream children in the park", (present participle with object) whereas you would say "there are three small children in the park".

kugel3: But my main point was more that it sounds like this is an array of "mnemonic entities" that are missing...

matiasbzurovski: ok, what about `MnemonicMissingEntities`?

RadixWallet/Features/DisplayEntitiesControlledByMnemonic/DisplayEntitiesControlledByMnemonic+Reducer.swift
matiasbzurovski: The idea for using only the first account, is that it if at least ONE of the accounts associated to a Mnemonic has problem3, then ALL of the accounts associated to that Mnmemonic will have problem 3. We can go with a Set and disjoint method though, would be more clear for sure

matiasbzurovski: > By the way, why does it take multiple accounts? Is the idea that we call this with all the accounts in a profile?

No, you call this with all the account associated with a mnemonic

kugel3: OK, but the function should not rely on facts ("all the accounts have the same mnemonic") that are only known at the call site, so we should either pick out that first account at the call site, are check all the supplied accounts.

RadixWallet/Clients/DeviceFactorSourceClient/DeviceFactorSourceClient+Interface.swift
kugel3: OK sure. Could still mean "a mnemonic that is missing its entities" but lets use this, then you get your `-Entities` suffix like you wanted, without violating grammar!

RadixWallet/Features/DisplayEntitiesControlledByMnemonic/DisplayEntitiesControlledByMnemonic+Reducer.swift
kugel3: Great, and when you made the changes you can just merge it into my pr

kugel3 (approved):

PR #1158, opened 2024-06-03
Use `OverlayWindowClient` to display cap36 migration messages

danvleju-rdx:
## Description
Rather than relying on custom delays, a more effective approach is to display migration messages using `OverlayWindowClient`.


RadixWallet/Features/HomeFeature/Coordinator/Home.swift
```diff
@@ -488,6 +453,21 @@ public struct Home: Sendable, FeatureReducer {
...
+	private func showLinkConnectorIfNeeded() -> Effect<Action> {
+		let purpose: NewConnectionApproval.State.Purpose? = if userDefaults.showRelinkConnectorsAfterProfileRestore {
+			.approveRelinkAfterProfileRestore
+		} else if userDefaults.showRelinkConnectorsAfterUpdate {
+			.approveRelinkAfterUpdate
+		} else {
+			nil
+		}
+		if let purpose {
+			overlayWindowClient.scheduleFullScreenIgnoreAction(.init(root: .relinkConnector(.init(root: .connectionApproval(.init(purpose: purpose))))))
```
matiasbzurovski, line 466: I was reviewing your other PR when it got merged, but I would rename this to `scheduleFullScreen` since we don't have actions associated to it (as we do on scheduled alerts).

RadixWallet/Features/HomeFeature/Coordinator/Home+View.swift
```diff
@@ -208,15 +207,6 @@ private extension View {
 			content: { NPSSurvey.View(store: $0) }
 		)
 	}
-
-	private func relinkConnector(with destinationStore: PresentationStoreOf<Home.Destination>) -> some View {
-		sheet(
```
matiasbzurovski, line 213: This used to be a sheet, and now is a full screen (which is more invasive in my opinion). Maybe we can have `SheetOverlayCoordinator` as well?

RadixWallet/Features/AppFeature/Overlay/FullScreenOverlayCoordinator+Reducer.swift
```diff
@@ -49,10 +58,21 @@ public struct FullScreenOverlayCoordinator: Sendable, FeatureReducer {
 	public func reduce(into state: inout State, childAction: ChildAction) -> Effect<Action> {
 		switch childAction {
 		case .root(.claimWallet(.delegate)):
-			.send(.delegate(.dismiss))
+			return .send(.delegate(.dismiss))
+
+		case let .root(.relinkConnector(.delegate(.newConnection(connectedClient)))):
+			userDefaults.setShowRelinkConnectorsAfterProfileRestore(false)
+			userDefaults.setShowRelinkConnectorsAfterUpdate(false)
```
matiasbzurovski, line 65: I can see this easily growing and ending up with this coordinator (which should be business logic independent, and only focused on presenting something anywhere on the app), into a class that takes care of very specific logic (as the one added now) for each case that it could present.

The ideal solution would be to have the `FullScreenOverlayCoordinator` be independent of `.claim` or `.relinkConnector`, and only take care of presenting a `FeatureView`. However, I don't know if there is an easy way of doing this with TCA. Have you looked into such approach?

RadixWallet/Features/HomeFeature/Coordinator/Home+View.swift
danvleju-rdx: Maybe better to have a new `OverlayReducer.Destination.State` say `sheet` and maybe rename the `FullScreenOverlayCoordinator`, but yes it’s a nice improvement. I’ll investigate what it implies.

RadixWallet/Features/AppFeature/Overlay/FullScreenOverlayCoordinator+Reducer.swift
danvleju-rdx: You’re right, although I don’t see any good solution right now. I already separated this logic from the `OverlayReducer` and in theory we should not overuse `OverlayWindowClient`, but maybe it’s worth looking for a better solution.

PR #1160, opened 2024-06-04
[ABW-3389] Fix navigation freeze

kugel3:
Jira ticket: [ABW-3389](https://radixdlt.atlassian.net/browse/ABW-3389)

## Description
Changes the navigation for the ImportMnemonic screen used when restoring from backup. It was incorrectly set up, which probably caused the freezes in the first place.

## PR submission checklist

### DIFF
--- a/RadixWallet/Features/ProfileBackupsFeature/RestoreProfileFromBackup/Children/ImportSeedPhrasesFlow/ImportMnemonicControllingAccounts+View.swift
+++ b/RadixWallet/Features/ProfileBackupsFeature/RestoreProfileFromBackup/Children/ImportSeedPhrasesFlow/ImportMnemonicControllingAccounts+View.swift
@@ -107,28 +107,20 @@ private extension View {
 	}
 
 	private func importMnemonic(with destinationStore: PresentationStoreOf<ImportMnemonicControllingAccounts.Destination>) -> some View {
-		sheet(
-			store: destinationStore,
-			state: /ImportMnemonicControllingAccounts.Destination.State.importMnemonic,
-			action: ImportMnemonicControllingAccounts.Destination.Action.importMnemonic,
-			content: {
-				ImportMnemonic.View(store: $0)
-					.radixToolbar(title: L10n.EnterSeedPhrase.Header.title, alwaysVisible: false)
-					.inNavigationView
-			}
-		)
+		sheet(store: destinationStore.scope(state: \.importMnemonic, action: \.importMnemonic)) { store in
+			ImportMnemonic.View(store: store)
+				.radixToolbar(title: L10n.EnterSeedPhrase.Header.title, alwaysVisible: false) {
+					store.send(.view(.closeButtonTapped))
+				}
+				.inNavigationStack
+		}
 	}
 
 	private func confirmSkippingBDFS(with destinationStore: PresentationStoreOf<ImportMnemonicControllingAccounts.Destination>) -> some View {
-		sheet(
-			store: destinationStore,
-			state: /ImportMnemonicControllingAccounts.Destination.State.confirmSkippingBDFS,
-			action: ImportMnemonicControllingAccounts.Destination.Action.confirmSkippingBDFS,
-			content: {
-				ConfirmSkippingBDFS.View(store: $0)
-					.inNavigationStack
-			}
-		)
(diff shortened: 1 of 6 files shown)


RadixWallet/Features/ProfileBackupsFeature/RestoreProfileFromBackup/Children/ImportSeedPhrasesFlow/ImportMnemonicControllingAccounts+View.swift
```diff
@@ -107,28 +107,20 @@ private extension View {
...
-				ImportMnemonic.View(store: $0)
-					.radixToolbar(title: L10n.EnterSeedPhrase.Header.title, alwaysVisible: false)
-					.inNavigationView
-			}
-		)
+		sheet(store: destinationStore.scope(state: \.importMnemonic, action: \.importMnemonic)) { store in
+			ImportMnemonic.View(store: store)
+				.radixToolbar(title: L10n.EnterSeedPhrase.Header.title, alwaysVisible: false) {
+					store.send(.view(.closeButtonTapped))
+				}
```
matiasbzurovski, line 114: This is adding a duplicate `X` button. Also, not sure this should be handled in the client that presents the `ImportMnemonic.View` rather than in the view itself.

IMG_0041

matiasbzurovski (changes requested):

RadixWallet/Features/ProfileBackupsFeature/RestoreProfileFromBackup/Children/ImportSeedPhrasesFlow/ImportMnemonicControllingAccounts+View.swift
kugel3: Weird, I have sort of the opposite problem, if I don't add this, I can't even see the built in close button. will check if it's a darkmode issue.

kugel3: it should ideally be handled in the view itself, and i would say the navbar should also be handled there. Tried a bunch of different approaches last night, but building was very slow so it too a lot of time (turns out I was still building the Release version...). If you do see the inner close button, then my initial approach should work, will revisit. 

In the meantime, the important thing is that this PR fixes a TCA runtime error (there was no navigation view) and it should fix the freezing issue.

matiasbzurovski: > , if I don't add this, I can't even see the built in close button. will check if it's a darkmode issue.

Weird, since we are setting `showCloseButton: true` when [setting this destination](https://github.com/radixdlt/babylon-wallet-ios/pull/1160/files#diff-eb286eb5c98f6b429e20261e4d53612820f99355732f93bcee68961e16671c45R127-R129)

kugel3: I couldn't reproduce it either but this PR fixes a TCA runtime error, which was probably related

kugel3: >Weird, since we are setting showCloseButton: true

Sure, it is setting it to show (I am doing that here because that makes dismissing work correctly), but on my phone it is not visible. Will look into this later today.

kugel3: It could be that the TCA error and the close buttons are related. If you run the old version and check the log, I get a big red TCA runtime error that the view is trying to present a destination without having a navigation view, which could also cause it to not show the navbar properly. Perhaps something changed in latest iOS.

kugel3:
I included the architectural fix in https://github.com/radixdlt/babylon-wallet-ios/pull/1161 instead, since it doesn't seem to fix the navigational glitch, and closing this.

PR #1161, opened 2024-06-04
[ABW-3396] Claim Wallet

kugel3:
Jira ticket: ABW-3396

## Description
Adds claim checking and reclaiming/clearing for two situations:
- When the profile is changed and is about to be uploaded
- Periodic checks every 15 minutes, checking the backed up copy of the current profile

When a profile is imported, we now follow this order:
1. Claim the local scope copy of the profile we just imported
2. Claim it on iCloud, if it exists there
3. Save it in ProfileStore/keychain

The PR also stops turning off iCloud keychain sync for recently migrated (non-active) profiles, and introduces turning it off for the active profile.

Fixes the navigation runtime error from TCA.

Changes the behaviour for the Configuration backup screen, w.r.t the four green/yellow items. Instead of linking it to security problems, the yellow versions are shown exactly when we show the "Last backup" label, as per [this discussion](https://rdxworks.slack.com/archives/C06SLHEHJ3T/p1717602259117309).

Changes the way we compare what we backed up with what we have currently, used for determining if we are properly backed up. Instead of using the hash of the profile, which changes on each execution of the app, it now uses a combination of `lastModified` and `lastUsedByDevice.id`. This eliminates some spurious backing up, but more importantly it does not claim that the manually exported backup is outdated if you relaunch the app. See discussion [here](https://rdxworks.slack.com/archives/C06SLHEHJ3T/p1717603922893469?thread_ts=1717602259.117309&cid=C06SLHEHJ3T) (same thread).

## PR submission checklist


RadixWallet/Features/ClaimWallet/ClaimWallet+View.swift
```diff
@@ -38,6 +38,12 @@ extension ClaimWallet {
 					Spacer()
 
 					VStack {
+						if viewStore.failedToReclaim {
+							Text("Failed to transfer ownership")
+								.foregroundColor(.app.error)
+								.textStyle(.secondaryHeader)
+						}
```
kugel3, line 45: We have no design for this, but on the other hand it is probably impossible to get into the state where we show it - if ProfileStore fails to claim the profile.

RadixWallet/Clients/CloudBackupClient/CloudBackupClient+Live.swift
```diff
@@ -142,20 +143,55 @@ extension CloudBackupClient {
 			try? userDefaults.setLastCloudBackup(result, of: profile)
 		}
 
+		@Sendable
```
kugel3, line 146: These three methods are the core of it

RadixWallet/Clients/CloudBackupClient/CloudBackupClient+Live.swift
```diff
@@ -142,20 +143,55 @@ extension CloudBackupClient {
...
+			let lastBackupSucceeded = lastBackup?.result == .success && lastBackup?.saveHash == profile.saveHash
+
+			let shouldBackUp = needsBackUp && !lastBackupSucceeded
+			let shouldCheckClaim = shouldBackUp || timeToCheckIfClaimed
+
+			guard shouldBackUp || shouldCheckClaim else { return }
+
+			let existingRecord = try? await fetchProfileRecord(.init(recordName: profile.id.uuidString))
+
+			let backedUpID = { try? existingRecord.map(getProfileHeader)?.lastUsedOnDevice.id }
```
GhenadieVP, line 159: Seems that `backedUpID` is called only once, so I guess it does not have to be a function.

RadixWallet/Clients/CloudBackupClient/CloudBackupClient+Live.swift
```diff
@@ -142,20 +143,55 @@ extension CloudBackupClient {
...
+
+			let didTransferBackProfile: Bool
+			if shouldCheckClaim, let id = backedUpID(), await !profileStore.isThisDevice(deviceID: id) {
```
GhenadieVP, line 162: Why go to profileStore? why not check directly on the `profile`?

GhenadieVP, line 162: Can `backedUpID` be ever `nil`? or it is just semantics

RadixWallet/Clients/CloudBackupClient/CloudBackupClient+Live.swift
```diff
@@ -142,20 +143,55 @@ extension CloudBackupClient {
...
+				let action = await overlayWindowClient.scheduleFullScreen(.init(root: .claimWallet(.init())))
+				didTransferBackProfile = action == .claimWallet(.didTransferBack)
+			} else {
+				didTransferBackProfile = false
+			}
+
+			guard shouldBackUp || didTransferBackProfile else { return }
+
+			await backupProfileAndSaveResult(existingRecord: existingRecord)
```
GhenadieVP, line 171: Why isn't profile passed down to be backed up?

RadixWallet/Clients/CloudBackupClient/CloudBackupClient+Live.swift
```diff
@@ -142,20 +143,55 @@ extension CloudBackupClient {
...
+				for try await (profile, tick) in combineLatest(profiles, ticks) {
 					guard !Task.isCancelled else { return }
-					guard profile.appPreferences.security.isCloudProfileSyncEnabled else { continue }
-					guard profile.header.isNonEmpty else { continue }
-
-					let last = userDefaults.getLastCloudBackups[profile.id]
-					if let last, last.result == .success, last.profileHash == profile.hashValue { continue }
-
-					await backupProfileAndSaveResult(profile)
+					try secureStorageClient.updateIsCloudProfileSyncEnabled(profile.id, .disable)
```
GhenadieVP, line 185: Done on every tick?

RadixWallet/Clients/CloudBackupClient/CloudBackupClient+Live.swift
```diff
@@ -219,6 +251,9 @@ extension CloudBackupClient {
 				try await fetchAllProfileRecords(headerOnly: true)
 					.map(getProfileHeader)
 					.filter(\.isNonEmpty)
+			},
+			reclaimProfile: {
+				try await profileStore.claimOwnershipOfProfile()
```
GhenadieVP, line 256: Isn't this basically a noop? I mean with new backup system, there is no way that current profile can have a different `lastUsedOnDeviceId`. Am I missing something?

RadixWallet/Clients/ProfileStore/ProfileStore.swift
```diff
@@ -476,7 +480,7 @@ extension ProfileStore {
 		// We present an alert to user where they must choice if they wanna keep using Profile
 		// on this device or delete it. If they delete a new one will be created and we will
 		// onboard user...
-		let choiceByUser = await overlayWindowClient.scheduleAlertAwaitAction(.profileUsedOnAnotherDeviceAlert(
+		let choiceByUser = await overlayWindowClient.scheduleAlert(.profileUsedOnAnotherDeviceAlert(
```
GhenadieVP, line 483: Why is this still here? do we have the claiming in two places? If still needed, we should present the new claim full screen, not alert.

RadixWallet/Clients/ResetWalletClient/ResetWalletClient+Live.swift
```diff
@@ -4,16 +4,24 @@ extension ResetWalletClient: DependencyKey {
...
 		@Dependency(\.userDefaults) var userDefaults
 
 		return Self(
 			resetWallet: {
-				cacheClient.removeAll()
-				await radixConnectClient.disconnectAll()
-				userDefaults.removeAll()
-				walletDidResetSubject.send(())
+				do {
+					try await appPreferencesClient.deleteProfileAndFactorSources(true)
```
GhenadieVP, line 16: Why was this added?

GhenadieVP (changes requested):
Nice work! This profile claim logic is tricky.

Before approving added some comments that require clarification.

RadixWallet/Clients/CloudBackupClient/CloudBackupClient+Live.swift
kugel3: It doesn't have to be, it's just for improving the readability the if-statement below. And we only want to calculate it if `shouldCheckClaim` is true.

RadixWallet/Clients/CloudBackupClient/CloudBackupClient+Live.swift
kugel3: Semantically, I want to check the backed up `lastUsedDevice.id` against the `deviceID` of the current device, not really against the `lastUsedDevice.id` of the profile in this loop, even if they should hopefully be the same.

kugel3: `existingRecord` is the backed up profile, and it can definitely be `nil`. In theory `getProfileHeader` could also throw, if the CKRecord doesn't contain all the properties.

RadixWallet/Clients/ProfileStore/ProfileStore.swift
kugel3: This particular piece of code is definitely not needed, but I want to go through ProfileStore with Cyon before cleaning it up, there is some logic that might still be useful.

RadixWallet/Clients/ResetWalletClient/ResetWalletClient+Live.swift
kugel3: It was just moved. Previously, this was called in the Main reducer, after `walletDidResetSubject` fired. So this change is just about:
1. Collecting all the logic for resetting in the client
2. Perform that throwing task first, so we don't end up clearing cache and userDefaults, and failing to delete the Profile.

RadixWallet/Clients/CloudBackupClient/CloudBackupClient+Live.swift
kugel3: It's way overkill, I really only want to do it when the profileID changes. I added it here at the last minute, but I will find a better place. Actually might be best to put it right outside the loop, because this loop is restarted when you delete the wallet.

RadixWallet/Clients/CloudBackupClient/CloudBackupClient+Live.swift
GhenadieVP: ok, so in this case should go forward and upload the current profile as to repair the record in the cloud?

kugel3: Yes, that's what the `didTransferBackProfile` flag is for

RadixWallet/Clients/CloudBackupClient/CloudBackupClient+Live.swift
kugel3: No you are right. I incorrectly thought that the local profile wasn't claimed automatically when importing from a file, so this is not needed.

RadixWallet/Features/AppFeature/Overlay/FullScreenOverlayCoordinator+Reducer.swift
```diff
@@ -16,6 +17,7 @@ public struct FullScreenOverlayCoordinator: Sendable, FeatureReducer {
 	}
 
 	public enum DelegateAction: Sendable, Equatable {
+		case claimWallet(ClaimWallet.DelegateAction)
 		case dismiss
```
danvleju-rdx, line 21: Is this dismiss action still needed here?

RadixWallet/Clients/CloudBackupClient/CloudBackupClient+Live.swift
kugel3: ...now renamed to `shouldReclaim`

RadixWallet/Features/AppFeature/Overlay/Overlay+Reducer.swift
```diff
@@ -99,19 +103,19 @@ struct OverlayReducer: Sendable, FeatureReducer {
 	}
 
 	func reduceDismissedDestination(into state: inout State) -> Effect<Action> {
-		dismissAlert(state: &state, withAction: .dismissed)
+		if case let .alert(state) = state.itemsQueue.first {
+			overlayWindowClient.sendAlertAction(.dismissed, state.id)
```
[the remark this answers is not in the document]
kugel3: No, fullscreen can not be dismissed without sending a delegate action, and any delegate action will dismiss it.

RadixWallet/Features/AppFeature/Overlay/FullScreenOverlayCoordinator+Reducer.swift
kugel3: Yes, but for other reasons. It is used as the fallback if we can't find any action corresponding to the fullscreen ID in question, similar to how it works in `sendAlertAction`.

RadixWallet/Clients/CloudBackupClient/CloudBackupClient+Live.swift
kugel3: It's left over from a previous flow, as it is now there is no reason to get the profile again. Will change it back.

RadixWallet/Features/AppFeature/Overlay/Overlay+Reducer.swift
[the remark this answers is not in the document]
matiasbzurovski: mm that goes for the Claim Wallet flow, but we probably will want to allow dismissal of other full screen flows presented this way.
We can make it so that it doesn't dismiss inside the view and just send `.delegate(.dismiss)` though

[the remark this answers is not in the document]
kugel3: The thing with fullscreen presentation is that you can't dismiss it by dragging or tapping outside, so you have to do it in code. But you are right, someone may think of calling the `dismiss()` dependency from the reducer, without actually sending any delegate actions, that wouldn't be good.

By the way, the way I set it up you have to send exactly one delegate action, and then we dismiss the view. Another way could have been to return a stream, and keep accepting delegate actions until the stream is completed, but that felt like overkill. For now you will have to do your logic either inside the fullscreen, or by telling the presenting reducer which delegate action to do. You can't for example continously update something in the presenting view, but that also is seldom needed when presenting a fullscreen view.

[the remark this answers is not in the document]
kugel3: I tested it and it work well. If someone uses TCA dismiss() then this method will get called, so I added handling here. Without that handling, the presenter would get stuck waiting for a delegate action forever.

GhenadieVP (approved):
LGTM!
Will start testing, I have two devices with the same cloud id.

kugel3:
I have also been testing with the help of the console, by changing e.g. lastUsedDeviceID for backed up profile directly

RadixWallet/Features/AppFeature/Overlay/Overlay+Reducer.swift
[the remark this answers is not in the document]
matiasbzurovski: Noticed that we only _catch_ the first action, and I think that's fine. Still, although fullscreens cannot be dismissed by dragging or tapping outside, it could be that they have a `X` button on navigation bar or similar. In that case, there would need to be a `DelegateAction` that is `case canceled` and just be dismissed here, is that right?

RadixWallet/Clients/UserDefaults+Dependency+Extension/UserDefaults+Dependency+Extension.swift
```diff
@@ -244,3 +244,9 @@ public struct BackupResult: Codable, Sendable {
 		case failure
 	}
 }
+
+extension Profile {
+	public var saveHash: String {
+		"\(header.lastModified.timeIntervalSince1970)-\(header.lastUsedOnDevice.id.uuidString)"
```
CyonAlexRDX, line 250: this is not a hash, will this be hashed?

CyonAlexRDX (approved):
LGTM

RadixWallet/Clients/UserDefaults+Dependency+Extension/UserDefaults+Dependency+Extension.swift
kugel3: True, it is not a hash anymore, I will rename it

RadixWallet/Features/ClaimWallet/ClaimWallet+Reducer.swift
```diff
@@ -34,13 +30,13 @@ public struct ClaimWallet: Sendable, FeatureReducer {
...
 		case .clearWalletButtonTapped:
-			.run { send in
+			return .run { send in
 				await resetWalletClient.resetWallet()
 				await send(.delegate(.didClearWallet))
 			}
 		case .transferBackButtonTapped:
-			// TODO: transfer back
-			.send(.delegate(.didTransferBack))
+			state.isLoading = true
```
danvleju-rdx, line 38: Is a loading state necessary on this screen, considering that the actual profile claiming is handled by `CloudBackupClient` and tapping this button will immediately dismiss the screen?

RadixWallet/Clients/SecureStorageClient/SecureStorageClient+Live.swift
```diff
@@ -77,29 +77,20 @@ extension SecureStorageClient: DependencyKey {
...
-			key: KeychainClient.Key,
-			iCloudSyncEnabled: Bool
+			key: KeychainClient.Key
 		) throws {
 			try keychainClient.setDataWithoutAuth(
 				data,
 				forKey: key,
 				attributes: .init(
-					iCloudSyncEnabled: iCloudSyncEnabled,
+					iCloudSyncEnabled: false,
```
GhenadieVP, line 86: I think that profile header list should also not be synced with Cloud. In SelectBackup, the profile header list is read, and then the id is used to load from keychain. Since profile is not synced anymore with keychain cloud, it is possible that the header will contain the id of profile missing from the keychain. Based on CloudBackupClient code, the migration will fail with `ProfileMissingFromKeychainError`.

RadixWallet/Features/ClaimWallet/ClaimWallet+Reducer.swift
kugel3: No it's not needed anymore. I used to have more loading state handling but apparently i forgot to remove the last bits

kugel3: I actually kept it, but on the other button. In theory resetWallet could take time and we don't want them to be able to keep tapping

RadixWallet/Clients/SecureStorageClient/SecureStorageClient+Live.swift
kugel3: alright

kugel3: But at the same time, we don't turn off syncing for migrated profiles, only for the currently active profile.

In your scenario: 
- we turn off syncing for profile X, the active one on phone A
- stop syncing the header list on phone A
- phone B asks for the list, it will include X, but X is not in iCloud keychain

Depending on how that list is maintained, it seems like we should keep it in sync?

RadixWallet/Clients/SecurityCenterClient/SecurityCenterClient+Live.swift
```diff
@@ -43,73 +43,90 @@ extension SecurityCenterClient {
...
+					} else {
+						print("•• CLOUD old but success")
 					}
+				case let .started(date):
+					print("•• CLOUD STARTED \(Date.now.timeIntervalSince(date))")
+				case .failure:
+					print("•• CLOUD FAILURE")
+				case .none:
+					print("•• CLOUD nil")
+				}
```
GhenadieVP, line 81: Should these still be here?

kugel3: no, they're gone now

RadixWallet/Clients/CloudBackupClient/CloudBackupClient+Live.swift
```diff
@@ -123,59 +120,99 @@ extension CloudBackupClient {
...
+
+			guard shouldBackUp || timeToCheckIfClaimed else { return }
+
+			let shouldReclaim: Bool
+			if let backedUpID = backedUpHeader?.lastUsedOnDevice.id, await !profileStore.isThisDevice(deviceID: backedUpID) {
+				let action = await overlayWindowClient.scheduleFullScreen(.init(root: .claimWallet(.init())))
+				switch action {
+				case .claimWallet(.transferBack):
+					shouldReclaim = true
+				case .claimWallet(.didClearWallet), .dismiss:
```
matiasbzurovski, line 164: I would leave the `assertionFailure` for the `.dismiss` case. It isn't possible right now, but if at some point we add a `X` to every full screen or something, we better fail rather than leave the wallet on an unwanted case (wallet not claimed nor deleted).

RadixWallet/Clients/DeviceFactorSourceClient/DeviceFactorSourceClient+Live.swift
```diff
@@ -71,68 +73,90 @@ extension DeviceFactorSourceClient: DependencyKey {
 			)
 		}
 
-		let problematicEntities: @Sendable () async throws -> (mnemonicMissing: ProblematicAddresses, unrecoverable: ProblematicAddresses) = {
-			let factorSources = try await factorSourcesClient.getFactorSources(type: DeviceFactorSource.self)
-			let accounts = try await accountsClient.getAccountsOnCurrentNetwork().elements
-			let hiddenAccounts = try await accountsClient.getHiddenAccountsOnCurrentNetwork().elements
-			let personas = try await personasClient.getPersonas().elements
-			let hiddenPersonas = try await personasClient.getHiddenPersonasOnCurrentNetwork().elements
+		struct FactorSourceHasMnemonic: Sendable, Equatable {
```
Sajjon, line 76: The name `FactorSourceHasMnemonic` does not really make any sense.

I think it should be renamed to `struct KeychainPresenceOfMnemonic`, which is what it does.

(Almost all FactorSources are "mnemonic based". Language wise a FactorSource cannot "have a mnemonic")

RadixWallet/Clients/DeviceFactorSourceClient/DeviceFactorSourceClient+Live.swift
```diff
@@ -71,68 +73,90 @@ extension DeviceFactorSourceClient: DependencyKey {
...
-			)
+				func unrecoverable(_ persona: Persona) -> Bool {
+					switch persona.securityState {
+					case let .unsecured(value):
+						unrecoverableFactorSources.contains(value.transactionSigning.factorSourceId)
+					}
+				}
 
-			return (mnemonicMissing: mnemonicMissing, unrecoverable: unrecoverable)
+				let mnemonicMissing = ProblematicAddresses(
```
Sajjon, line 143: Hmm an address cannot be "problematic". Better `AddressesOfEntitiesInBadState` IMO.

Maybe `entitiesWithoutControl` is better than `mnemonicMissing`? Also suitable post-MFA.

RadixWallet/Clients/DeviceFactorSourceClient/DeviceFactorSourceClient+Live.swift
```diff
@@ -71,68 +73,90 @@ extension DeviceFactorSourceClient: DependencyKey {
...
+					}
+				}
 
-			let unrecoverable = ProblematicAddresses(
-				accounts: accounts.filter(unrecoverable(_:)).map(\.address),
-				hiddenAccounts: hiddenAccounts.filter(unrecoverable(_:)).map(\.address),
-				personas: personas.filter(unrecoverable(_:)).map(\.address),
-				hiddenPersonas: hiddenPersonas.filter(unrecoverable(_:)).map(\.address)
-			)
+				func unrecoverable(_ persona: Persona) -> Bool {
```
Sajjon, line 136: the two functions:

```swift
				func unrecoverable(_ account: Account) -> Bool {
					switch account.securityState {
					case let .unsecured(value):
						unrecoverableFactorSources.contains(value.transactionSigning.factorSourceId)
					}
				}
```

and  `unrecoverable(_ account: Account)` can be merged into one:

```
func isUnrecoverable(entity: some EntityProtocol) -> Bool {
    switch entity.securityState {
         case .unsecured(control): unrecoverableFactorSources.contains(control.transactionSigning.factorSourceId)
    }
}
```

Sajjon: Or by use of `AccountOrPersona`...

RadixWallet/Clients/KeychainClient/KeychainClient+Interface.swift
```diff
@@ -16,6 +16,7 @@ public struct KeychainClient: Sendable {
 	public var _removeDataForKey: RemoveDataForKey
 	public var _removeAllItems: RemoveAllItems
 	public var _getAllKeysMatchingAttributes: GetAllKeysMatchingAttributes
+	public var _keychainChanged: KeychainChanged
```
Sajjon, line 19: Maybe we ought write a note that this is a "best effort" publisher, it does not detect external changes, such as Keychain getting wiped when a passcode is disabled. So that we neither nor, nor any colleague in future assumes it always emits for all changes.

RadixWallet/Clients/CloudBackupClient/CloudBackupClient+Live.swift
kugel3: It does actually sometimes get sent, about once in 25 presentations, in addition to the actual action. This is because it's the fallback action if the actions sequence completes prematurely, which seems to happen on rare occasions. It's not visible to the user so we better postpone that investigation.

matiasbzurovski: but is the view actually dismissed?

RadixWallet/Clients/TransportProfileClient/TransportProfileClient+Interface.swift
```diff
@@ -0,0 +1,26 @@
...
+		didExportProfile: @escaping DidExportProfile
+	) {
+		self.importProfile = importProfile
+		self.profileForExport = profileForExport
+		self.didExportProfile = didExportProfile
+	}
+}
+
+extension TransportProfileClient {
+	public typealias ImportProfile = @Sendable (Profile, Set<FactorSourceIDFromHash>, Bool) async throws -> Void
```
Sajjon, line 23: if you write `, _ containsP2PLinks: Bool)` in the typealias one knows what the Bool is just by looking at the typealias, or write a comment? :)

RadixWallet/Clients/TransportProfileClient/TransportProfileClient+Interface.swift
```diff
@@ -0,0 +1,26 @@
+import Sargon
+
+public typealias DeviceID = UUID
+
+// MARK: - TransportProfileClient
+public struct TransportProfileClient: Sendable {
```
Sajjon, line 6: Or `VandelayClient` 😆

CyonAlexRDX (approved):
Looks good, mostly naming I would like to change

RadixWallet/Clients/DeviceFactorSourceClient/DeviceFactorSourceClient+Live.swift
```diff
@@ -71,68 +73,90 @@ extension DeviceFactorSourceClient: DependencyKey {
...
-				secureStorageClient.containsMnemonicIdentifiedByFactorSourceID($0.id)
-			}
+		let problematicEntities: @Sendable () async throws -> AnyAsyncSequence<(mnemonicMissing: ProblematicAddresses, unrecoverable: ProblematicAddresses)> = {
+			await combineLatest(factorSourcesMnemonicPresence(), userDefaults.factorSourceIDOfBackedUpMnemonics(), profileStore.values()).map { factorSources, backedUpFactorSources, profile in
 
-			let unrecoverableFactorSources = mnemonicPresentFactorSources.filter {
-				!userDefaults.getFactorSourceIDOfBackedUpMnemonics().contains($0.id)
-			}.map(\.id)
+				let mnemonicMissingFactorSources = factorSources
+					.filter(not(\.present))
```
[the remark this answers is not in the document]
matiasbzurovski: you mean rename `present` to `presenceOfMnemonics`? If so, I'd say it is redundant given the name of the struct is `KeychainPresenceOfMnemonic`

danvleju-rdx (approved):

RadixWallet/Clients/CloudBackupClient/CloudBackupClient+Live.swift
kugel3: Yes, there is no user-visible glitch

PR #1165, opened 2024-06-05
Fix bug where Security Problems where missing for hidden entities

matiasbzurovski:
## Description
To reproduce this bug, just follow these steps:
- Reset Wallet
- Create new Account
- Hide such account
- Open Settings 

| Before | After |
| -- | -- |
| IMG_0046 | IMG_0045 |


CyonAlexRDX (approved):
LGTM

PR #1166, opened 2024-06-06
[ABW-3420]Fix performance for EntitySecurityProblems

matiasbzurovski:
Jira ticket: [ABW-3420](https://radixdlt.atlassian.net/browse/ABW-3420)

## Description
When loading a big profile (30+ accounts), the app would show laggy or even freeze after hiding accounts and scrolling on Home. To tackle this, this PR includes 2 changes:

**1) Improve logic used to determine `problematicEntities`**
When checking which entities were problematic (mnemonic not imported or written down), we were using `DeviceFactorSourceClient.entitiesControlledByFactorSource` for every factor source available. This was expensive and quite inefficient, as we don't really need to get the entities grouped by factor source. The new logic implemented reduced the time at ~6x, as you can see on [this log](https://github.com/user-attachments/files/15656522/before_after.json) where I performed old and new methods to compare their performance (and validate the output was the same).

**2) Check Security Problems once for all Accounts**
The `EntitySecurityProblems` was a nice component, since it would be able to fetch the problems and then filter those that mattered for a given entity. However, if we had 30 account rows at Home, then it meant that we would check the problems 30 times and then filter 30 times as well. This is quite inefficient, as the problems will be the same the 30 times, so it makes more sense to check only once and have each row filter the problems that matter to it.

[ABW-3420]: https://radixdlt.atlassian.net/browse/ABW-3420?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


RadixWallet/Features/EntitySecurityProblems/EntitySecurityProblems.swift
```diff
@@ -8,9 +8,16 @@ public struct EntitySecurityProblems: Sendable, FeatureReducer {
 
 		let kind: Kind
 		var problems: [SecurityProblem] = []
+		fileprivate let loadProblems: Bool
 
-		public init(kind: Kind) {
+		public init(kind: Kind, problems: [SecurityProblem]? = nil) {
```
matiasbzurovski, line 13: Initially I wanted to inject the `[SecurityProblems]` every time, forcing the client of it to have them loaded. However, this is used by `PersonaFeature.View`, which is used in different contexts in which handling the load of the security problems will be troublesome. Therefore, decided to leave this component as one that could either receive the problems (as in Home rows) or load them by itself (Personas).

matiasbzurovski: Update: I believe it is worth going with the plain SwiftUI solution for this, and have the `problems` always injected. Will work on improvement tomorrow.

CyonAlexRDX (approved):
Good job!

CyonAlexRDX (approved):
We should upgrade this to Sargon post launch. To have Sargon analyse the problems 🥳

kugel3 (approved):

GhenadieVP (approved):

PR #1167, opened 2024-06-07
SecurityCenterClient improvements

matiasbzurovski:
Use `AnyAsyncSequence` inside `SecurityCenterClient` so that checks take place only when a triggering change has fired.


RadixWallet/Clients/SecurityCenterClient/SecurityCenterClient+Interface.swift
```diff
@@ -5,13 +5,15 @@ import os
 
 // MARK: - SecurityCenterClient
 public struct SecurityCenterClient: DependencyKey, Sendable {
+	public let startMonitoring: StartMonitoringProblems
```
kugel3, line 8: the type and the variable should have exactly the same name

kugel3 (approved):

PR #1168, opened 2024-06-07
Fix Security Problems issues

matiasbzurovski:
- Fix bug where Personas list wouldn't show security alerts
- Fix bug where security alerts were sometimes not loaded on app start


RadixWallet/Clients/DeviceFactorSourceClient/DeviceFactorSourceClient+Live.swift
```diff
@@ -80,7 +80,7 @@ extension DeviceFactorSourceClient: DependencyKey {
 
 		@Sendable
 		func factorSourcesMnemonicPresence() async -> AnyAsyncSequence<[FactorSourceHasMnemonic]> {
-			await combineLatest(profileStore.factorSourcesValues(), secureStorageClient.keychainChanged())
+			await combineLatest(profileStore.factorSourcesValues(), secureStorageClient.keychainChanged().prepend(()))
```
matiasbzurovski, line 83: Given on app start we don't have a keychainChanged event, we weren't getting this fired

PR #1172, opened 2024-06-12
Retroactively Update Backup Log

kugel3:
## Description
Instead of only updating the locally stored log of old backups when the user has backups turned on and we discover that no backup is present, this PR will make it so that we always update the log. This ensures that we don't show the `Out-of-date backup still present on iCloud` warning if the user manually deletes iCloud content from settings.

## How to test
- Turn of Automatic backups
- Note that `Out-of-date backup still present on iCloud` is visible
- Manually delete iCloud contents from settings
- Note that the warning goes away

## Video
The scenario from How to test:

https://github.com/radixdlt/babylon-wallet-ios/assets/123396602/24f82413-01ef-4c65-9b9a-8fc66618fc2e

What happens if you delete the backup from within the app

https://github.com/radixdlt/babylon-wallet-ios/assets/123396602/84417444-655f-46e3-82bf-688fd3faa542

## PR submission checklist


GhenadieVP (approved):
LGTM!

PR #1174, opened 2024-06-13
Fix old QR code error message

danvleju-rdx:
Fixes the error message that appears when scanning an old QR code.


GhenadieVP (approved):

CyonAlexRDX (approved):
lgtm

PR #1183, opened 2024-06-21
Refine Backup Behaviour

kugel3:
## Description
Pursuant to [this discussion](https://rdxworks.slack.com/archives/C03QFAWBRNX/p1718896121997059), the restore from backup behaviour is fixed and refined by this PR:

1. If you restore a profile from disk which has iCloud sync turned off, we will no longer try to claim it on iCloud (bugfix)
2. If you restore a profile that has iCloud sync turned on and claiming on iCloud fails, we will continue to import (bugfix) but Security Center will show a problem
3. Prior to this, when loading the list of iCloud backups (or migrating keychain backups ahead of loading the list), if something goes wrong we now distinguish between three cases:

- Not logged in to iCloud, show "Not logged in to iCloud" in the list
- No internet, show "Network unavailable" in the list
- Other error (very rare), show "Could not load backups" in the list and the error in an alert

The strings were added [here](https://github.com/radixdlt/apps-localization-src/pull/413)

## How to test
(Using Pre-alpha 1.6.1 (1) for now, until this is merged)

1. Try to restore from a file that has iCloud sync disabled
2. Try to restore from a file while internet is off
3. Try to load the list of backups with internet off
4. Try to load the list of backups while logged out of iCloud
5. Try to load the list of backups with iCloud for Radix Wallet disabled

image

### DIFF
--- a/RadixWallet/Clients/CloudBackupClient/CloudBackupClient+Live.swift
+++ b/RadixWallet/Clients/CloudBackupClient/CloudBackupClient+Live.swift
@@ -167,13 +167,11 @@ extension CloudBackupClient {
 				try? userDefaults.removeLastCloudBackup(for: profile.id)
 			}
 
-			let isSyncEnabled = profile.appPreferences.security.isCloudProfileSyncEnabled
-			let needsBackingUp = isSyncEnabled && profile.header.isNonEmpty
+			guard profile.appPreferences.security.isCloudProfileSyncEnabled else { return }
 			let isBackedUp = backedUpHeader?.saveIdentifier == profile.header.saveIdentifier
-			let shouldBackUp = needsBackingUp && !isBackedUp
-			let shouldCheckIfClaimed = isSyncEnabled && timeToCheckIfClaimed
+			let shouldBackUp = profile.header.isNonEmpty && !isBackedUp
 
-			guard shouldBackUp || shouldCheckIfClaimed else { return }
+			guard shouldBackUp || timeToCheckIfClaimed else { return }
 
 			let shouldReclaim: Bool
 			if let backedUpID = backedUpHeader?.lastUsedOnDevice.id, await !profileStore.isThisDevice(deviceID: backedUpID) {
--- a/RadixWallet/Clients/CloudBackupClient/CloudBackupClient+Live.swift
+++ b/RadixWallet/Clients/CloudBackupClient/CloudBackupClient+Live.swift
@@ -43,6 +43,12 @@ extension [CKRecord.FieldKey] {
 }
 
 extension CloudBackupClient {
+	public func deleteProfileBackup() async throws {
+		try await deleteProfileBackup(nil)
+	}
+}
+
+extension CloudBackupClient {
 	struct IncorrectRecordTypeError: Error {}
 	struct NoProfileInRecordError: Error {}
 	struct MissingMetadataError: Error {}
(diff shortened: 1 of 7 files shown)


RadixWallet/Clients/CloudBackupClient/CloudBackupClient+Live.swift
```diff
@@ -156,11 +156,13 @@ extension CloudBackupClient {
 				try? userDefaults.removeLastCloudBackup(for: profile.id)
 			}
 
-			let needsBackUp = profile.appPreferences.security.isCloudProfileSyncEnabled && profile.header.isNonEmpty
+			let isSynced = profile.appPreferences.security.isCloudProfileSyncEnabled
```
CyonAlexRDX, line 159: I think `isSynced` is a misleading variable name, sounds like a sync status is *known to be in status 'synced'*, I'd call it isCloudSyncEnabled` or similar.

RadixWallet/Clients/CloudBackupClient/CloudBackupClient+Live.swift
```diff
@@ -156,11 +156,13 @@ extension CloudBackupClient {
...
+			let needsBackUp = isSynced && profile.header.isNonEmpty
```
CyonAlexRDX, line 160: `needsBackingUp` ? or `needsToBeBackedUp`?

RadixWallet/Clients/TransportProfileClient/TransportProfileClient+Live.swift
```diff
@@ -14,7 +14,9 @@ extension TransportProfileClient: DependencyKey {
 				do {
 					var profile = profile
 					await profileStore.claimOwnership(of: &profile)
-					try await cloudBackupClient.claimProfileOnICloud(profile)
+					if profile.appPreferences.security.isCloudProfileSyncEnabled {
+						try? await cloudBackupClient.claimProfileOnICloud(profile)
```
CyonAlexRDX, line 18: makes sense! Perhaps use `do catch` and log the error?

CyonAlexRDX (approved):
LGTM, suggested some name changes

RadixWallet/Clients/CloudBackupClient/CloudBackupClient+Live.swift
kugel3: Sure

RadixWallet/Clients/CloudBackupClient/CloudBackupClient+Live.swift
kugel3: Done

RadixWallet/Clients/TransportProfileClient/TransportProfileClient+Live.swift
kugel3: Sure, logging inside the function instead

RadixWallet/Clients/CloudBackupClient/CloudBackupClient+Live.swift
matiasbzurovski, line 159: We could also abandon early if iCloud sync is disabled 
```Swift
guard profile.appPreferences.security.isCloudProfileSyncEnabled else { return }
```

kugel3: Currently we will still check if the profile has been claimed on another phone, but you could argue that we don't care about that if they have disabled iCloud sync. That would simplify things.

kugel3: True, that was recently changed, will simplify this now

matiasbzurovski (approved):

PR #1186, opened 2024-06-25
[ABW-3261] Display all resource metadata fields

matiasbzurovski:
Jira ticket: [ABW-3261](https://radixdlt.atlassian.net/browse/ABW-3261)
Jira ticket: [ABW-2863](https://radixdlt.atlassian.net/browse/ABW-2863)

## Description
- Loads and display every metadata field available on a Resource.
- Shows which arbitrary metadata is locked
- Updates UI to follow new designs.

## How to test
You can create a new test token under `Dev preferences`, which will create a `STAR` token with 37 metadata fields.
If you want to test pagination with a resource that has more than 1 page of metadata, let me know and I can send you an NFT with that.

## Screenshots

[ABW-3261]: https://radixdlt.atlassian.net/browse/ABW-3261?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[ABW-2863]: https://radixdlt.atlassian.net/browse/ABW-2863?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


RadixWallet/Clients/GatewayAPI/GatewayAPI+Utils.swift
```diff
@@ -61,12 +65,12 @@ public enum EntityMetadataKey: String, CaseIterable {
 	case ownerBadge = "owner_badge"
 
 	// The GW limits the number of metadata keys we can ask for
-	static let maxAllowedKeys = 10
+	static let maxAllowedKeys = 11
```
GhenadieVP, line 68: Does the GW now allow for more than 10 metadata keys?

RadixWallet/Clients/OnLedgerEntitiesClient/OnLedgerEntitiesClient+Live.swift
```diff
@@ -251,12 +251,19 @@ extension OnLedgerEntitiesClient {
 					ledgerState: response.ledgerState
 				)
 
+				var allMetadataItems = Set(item.metadata.items)
+				if let nextCursor = item.metadata.nextCursor {
+					// Only fetch metadata if there is more after the first page returned.
+					let remaining = try await gatewayAPIClient.fetchEntityMetadata(item.address, ledgerState: response.ledgerState, nextCursor: nextCursor)
+					allMetadataItems.append(contentsOf: remaining)
+				}
```
GhenadieVP, line 259: Not 100% sure about this, this will increase the overall loading times of the Wallet. Ideally the metadata should fetched only when going in the Resource Details.

RadixWallet/Features/AssetsFeature/Components/HelperViews/ArbitraryDataField/ArbitraryDataField+Helpers.swift
```diff
@@ -0,0 +1,236 @@
+import Foundation
+
+extension [GatewayAPI.EntityMetadataItem] {
+	var asDataFields: [ArbitraryDataField] {
```
GhenadieVP, line 4: Nice 👍

RadixWallet/Clients/OnLedgerEntitiesClient/OnLedgerEntitiesClient+Live.swift
matiasbzurovski: Actually it wasn't increasing the overall loading times, since we only fetch the actual resource when we try to open its details. Before that, we are actually requesting `state/entity/details` for the accounts (not the resources), and in such case we don't have any `item.metadata.nextCursor` (as it only fetches the explicit metadata of the resources inside the accounts).

However, I pushed a new commit that explicitly requests the metadata only when requested. It adds some noise, so let me know if you prefer which version you prefer.

GhenadieVP: Almost true. We are fetching the resource details for the ones that are either part of staking or pool units, see [fetchPoolAndStakeUnitsDetails](https://github.com/radixdlt/babylon-wallet-ios/blob/5a676d23184651706492223f4080a9450738103b/RadixWallet/Clients/AccountPortfoliosClient/AccountPortfoliosClient%2BLive.swift#L127) or when categorising resources to pool units, stake units, nft claims

GhenadieVP (approved):

PR #1187, opened 2024-06-28
[ABW-3439] Card Carousel

kugel3:
Jira ticket: [ABW-3439](https://radixdlt.atlassian.net/browse/ABW-3439)
Jira ticket: [ABW-3443](https://radixdlt.atlassian.net/browse/ABW-3443)

Depends on https://github.com/radixdlt/sargon/pull/180

## Description
A carousel showing various action cards at the top of the home screen.

[ABW-3439]: https://radixdlt.atlassian.net/browse/ABW-3439?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[ABW-3443]: https://radixdlt.atlassian.net/browse/ABW-3443?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


RadixWallet/Clients/CardCarouselClient/CardCarouselClient+Interface.swift
```diff
@@ -0,0 +1,32 @@
...
+		closeCard: @escaping CloseCard
+	) {
+		self.cards = cards
+		self.closeCard = closeCard
+	}
+}
+
+// MARK: - CarouselCard
+public enum CarouselCard: Hashable, Sendable {
+	case threeSixtyDegrees
```
CyonAlexRDX, line 17: Do we plan on letting the Cards be an enum? Is it because we need to know what to do when the card is pressed, and we dont wanna represent that with a closure? Especially not if this is to be moved into Sargon?

kugel3: It looks like the repertoire of cards is very limited, and they are completely static, so it seems like a good model. The action will be known at compile time, not sure it makes sense to store it on the model, it can just be handled in the reducer, or possibly the client.

RadixWallet/Clients/CardCarouselClient/CardCarouselClient+Interface.swift
```diff
@@ -0,0 +1,48 @@
...
+	public let id: CardID
+	public let action: Action
+
+	public enum CardID: Sendable {
+		case rejoinRadQuest
+		case discoverRadix
+		case continueOnDapp
+		case useDappsOnDesktop
+		case threeSixtyDegrees
+	}
```
matiasbzurovski, line 29: Note: this should will be in Sargon

RadixWallet/Core/Resources/Resources/Assets.xcassets/CardCarousel/carousel_icon_continueOnDapp.imageset/Contents.json
```diff
@@ -0,0 +1,23 @@
+{
+  "images" : [
+    {
+      "filename" : "carousel_icon_continueOnDapp.png",
```
matiasbzurovski, line 4: This should be removed, since the dApp icon url will come from Sargon after retrieving metadata.

RadixWallet/Core/Resources/Resources/Assets.xcassets/CardCarousel/carousel_full_background_360.imageset/Contents.json
```diff
@@ -0,0 +1,12 @@
+{
+  "images" : [
+    {
+      "filename" : "card_background_TEMPORARY.pdf",
```
matiasbzurovski, line 4: Won't be used so can be removed.

RadixWallet/Features/HomeFeature/Coordinator/Home+View.swift
```diff
@@ -99,23 +99,19 @@ extension Home {
...
-				VStack(alignment: .leading, spacing: .small2) {
+				VStack(spacing: .small2) {
 					Text(L10n.HomePage.title)
 						.foregroundColor(.app.gray1)
 						.textStyle(.sheetTitle)
+						.flushedLeft
+						.padding(.horizontal, .medium1)
 
-					HStack {
-						Text(L10n.HomePage.subtitle)
```
matiasbzurovski, line 110: Confirm if we want to remove subtitle.

RadixWallet/Features/CardCarousel/CardCarousel+View.swift
```diff
@@ -0,0 +1,211 @@
...
+			}
+
+			return result
+		}
+	}
+}
+
+// MARK: - CarouselCardView
+public struct CarouselCardView: View {
+	private let trailingPadding: CGFloat = 115
```
matiasbzurovski, line 88: Should be 96

RadixWallet/Features/HomeFeature/Coordinator/Home+View.swift
kugel3: It's explicitly mentioned in the ticket. 

The ticket also says "After the carousel is in place, the Start using Radix Wallet CTA card from home screen can be moved up in the carousel. It is not yet clear if it will have the same design, but can be used as an example." so that should not be overlooked, when the design is finalised.

RadixWallet/Features/CardCarousel/CardCarousel+View.swift
```diff
@@ -0,0 +1,211 @@
...
+		private let spacing: CGFloat = .small1
+		@ScaledMetric private var height: CGFloat = 105
+
+		public var body: some SwiftUI.View {
+			WithPerceptionTracking {
+				GeometryReader { proxy in
+					coreView
+						.backgroundPreferenceValue(PositionsPreferenceKey.self) { positions in
+							dummyCards(positions, in: proxy.frame(in: .named(Self.coordSpace)))
+						}
```
matiasbzurovski, line 21: I am not really sure about the purpose of these dummy cards in the background, since the height is already set to 105 (dynamically changing based on device font size configuration).

As these videos show, adding the dummy views presents a glitch that would be avoided if removed:
| With dummy views | Without dummy views |
| -- | -- |
| <video src=https://github.com/radixdlt/babylon-wallet-ios/assets/164921079/35b3ac6e-5a67-4f54-bf5b-b6443fbe3381> | <video src=https://github.com/radixdlt/babylon-wallet-ios/assets/164921079/cba6c06a-8664-424a-9153-72accd45af55> |
| <video src=https://github.com/radixdlt/babylon-wallet-ios/assets/164921079/c7280444-3c96-4f4d-8cef-5fca79c42380> | <video src=https://github.com/radixdlt/babylon-wallet-ios/assets/164921079/ecd53fb4-e999-4bea-a157-f03015357974> |

RadixWallet/Clients/HomeCardsClient/HomeCardsClient+Live.swift
```diff
@@ -0,0 +1,63 @@
...
+		)
+	}()
+}
+
+// MARK: - HomeCardsManager + Sendable
+extension HomeCardsManager: @unchecked Sendable {}
+
+// MARK: - HomeCardsStorage
+private final class HomeCardsStorage: Sargon.HomeCardsStorage {
+	@Dependency(\.userDefaults) var userDefaults
```
GhenadieVP, line 45: just making sure, will this be using `UserDefaults.Dependency.radix`? You can check the actual values are displayed in Debug Settings

RadixWallet/Features/CardCarousel/CardCarousel+View.swift
```diff
@@ -0,0 +1,194 @@
...
+
+		private let margin: CGFloat = .medium1
+		private let spacing: CGFloat = .small1 * 0.5
+		@ScaledMetric private var height: CGFloat = 105
+		@SwiftUI.State private var selectedCardIndex = 0
+
+		public var body: some SwiftUI.View {
+			WithPerceptionTracking {
+				VStack(spacing: 8) {
+					GeometryReader { _ in
```
GhenadieVP, line 17: Is this GeometryReader still needed?

RadixWallet/Features/CardCarousel/CardCarousel+View.swift
```diff
@@ -0,0 +1,194 @@
...
+						coreView
+					}
+					.frame(height: store.cards.isEmpty ? 0 : height)
```
GhenadieVP, line 20: could we rather not add the view if there are no cards?

RadixWallet/Features/CardCarousel/CardCarousel+View.swift
```diff
@@ -0,0 +1,194 @@
...
+			}
+			.onAppear {
+				store.send(.view(.didAppear))
+			}
+		}
+
+		@MainActor
+		private var coreView: some SwiftUI.View {
+			TabView(selection: $selectedCardIndex) {
+				ForEach(Array(store.cards.enumerated()), id: \.1) { index, card in
```
GhenadieVP, line 33: could we use `\.element` instead of `\.1`?

RadixWallet/Features/CardCarousel/CardCarousel+View.swift
```diff
@@ -0,0 +1,194 @@
...
+					.transition(.scale(scale: 0.8).combined(with: .opacity))
+				}
+			}
+			.tabViewStyle(.page(indexDisplayMode: .never))
+			.animation(.default, value: store.cards)
+		}
+
+		private var positionIndicator: some SwiftUI.View {
+			HStack(spacing: spacing) {
+				if store.cards.count > 1 {
```
GhenadieVP, line 50: Minor: maybe it should be outside of `positionIndicator`, just to not add an empty HStack I guess

RadixWallet/Features/CardCarousel/CardCarousel+View.swift
```diff
@@ -0,0 +1,194 @@
...
+						.lineLimit(nil)
+						.minimumScaleFactor(0.8)
+						.textStyle(.body2Regular)
+				}
+				.multilineTextAlignment(.leading)
+				.foregroundStyle(.app.gray1)
+				.padding([.top, .leading], .medium2)
+				.padding(.trailing, trailingPadding)
+				.padding(.bottom, .small1)
+				.frame(maxWidth: .infinity, maxHeight: .infinity, alignment: .topLeading)
```
GhenadieVP, line 95: This is like a lot of padding + frame, is it really needed?

RadixWallet/Features/CardCarousel/CardCarousel+View.swift
```diff
@@ -0,0 +1,194 @@
...
+				.cornerRadius(.small1)
+			}
+
+			Button(action: closeAction) {
+				Image(asset: AssetResource.close)
+					.resizable()
+					.frame(width: .medium3, height: .medium3)
+					.tint(.app.gray1)
+					.padding(.small2)
+			}
```
GhenadieVP, line 109: We have already close button, surprised that this is needed

RadixWallet/Features/CardCarousel/CardCarousel+View.swift
```diff
@@ -0,0 +1,194 @@
...
+			.frame(.small, alignment: .topTrailing)
+		}
+	}
+
+	private var trailingPadding: CGFloat {
```
GhenadieVP, line 114: Not quite getting what this is for, I guess I need to play with it.

RadixWallet/Features/CardCarousel/CardCarousel+View.swift
matiasbzurovski: The close button is intended to be used in a toolbar so it has a bigger padding, which makes the `X` overlap with the background/dapp image. That's why I changed it to this manual implementation.

RadixWallet/Features/CardCarousel/CardCarousel+View.swift
matiasbzurovski: basically how much space we must leave empty from the trailing side of the view. Depending on its content, we have more or less space that we can use without the text overlapping with an image.

Original implementation had a fixed value, but that would leave very little space for some views on smaller devices (as iPhone SE) and content would look trimmed.

RadixWallet/Features/CardCarousel/CardCarousel+View.swift
matiasbzurovski: given the logic to determine if there are (or not) cards is inside the view's reducer, we can't

RadixWallet/Clients/CardCarouselClient/CardCarouselClient+Live.swift
```diff
@@ -0,0 +1,23 @@
...
+
+extension CardCarouselClient: DependencyKey {
+	public static let liveValue: Self = {
+		let cardSubject = AsyncCurrentValueSubject<[CarouselCard]>([.threeSixtyDegrees, .connect, .somethingElse])
+
+		@Dependency(\.errorQueue) var errorQueue
+		@Dependency(\.appPreferencesClient) var appPreferencesClient
+		@Dependency(\.cacheClient) var cacheClient
+		@Dependency(\.radixConnectClient) var radixConnectClient
+		@Dependency(\.userDefaults) var userDefaults
```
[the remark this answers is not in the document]
matiasbzurovski: Ignoring since updated version of client no longer has it.

RadixWallet/Features/CardCarousel/CardCarousel+View.swift
matiasbzurovski: Yes 🫤

RadixWallet/Clients/HomeCardsClient/HomeCardsClient+Live.swift
matiasbzurovski: all good. It wasn't showing any data but I've updated it so that it either shows `Data available` or `No Data available` for home cards key.

RadixWallet/Features/CardCarousel/CardCarousel+View.swift
kugel3: This makes it so that the frame grows as much as it can, instead of hugging the content. It's very similar to using say an HSTack with Spacer's.

RadixWallet/Features/CardCarousel/CardCarousel+View.swift
kugel3: The reason it was fixed is that the image we want to avoid covering is fixed size, because the view has a fixed height. That in turn is because tabview can't handle flexible heights very well.

kugel3:
I saw a question about the dummy views, though I can't find it anymore. Their purpose is to show the teased cards on the sides. The tabview itself does not reach the end of the screen, it is only as wide as the card, plus half the inter card spacing on each side.

RadixWallet/Features/CardCarousel/CardCarousel+View.swift
kugel3: Without the dummy cards, and hence the geometry reader. the carousel will not look like Aftab's design, it will not have teaser cards on the sides. Did he change his mind about that?

kugel3: Zeplin [still shows](https://app.zeplin.io/project/639c7e7ab391bb15ac5d25fa/screen/666b6f35261692048fbc7929) teaser cards.

RadixWallet/Features/CardCarousel/CardCarousel+View.swift
kugel3: What is the purpose of enumerating it and then use `element` as `id`, instead of just doing `\id:.self`? And stepping back, what was wrong with having the card be `identifiable`?

RadixWallet/Features/CardCarousel/CardCarousel+View.swift
GhenadieVP: We persuaded Aftab to remove these, as they were redundant in the context of showing the position indicator.

RadixWallet/Features/CardCarousel/CardCarousel+View.swift
GhenadieVP: CloseButton does not seem to have any padding, but anyway would be nice to have a single definition of CloseButton if possible.

RadixWallet/Features/CardCarousel/CardCarousel+View.swift
matiasbzurovski: We use the index to tag each card, which is necessary for the positionIndicator defined below. We can extend the card model (now defined in Sargon), and make it `Identifiable` by defining an id for each. Same result, just need some extra code.

RadixWallet/Features/CardCarousel/CardCarousel+View.swift
matiasbzurovski: `CloseButton` defines a size of 44x44, meaning the asset (24x24) will have a 10 padding on each side.

In this case, we want the asset to be of 16x16 and to have a padding of 8 to each side.

I can extend the CloseButton to adapt for each type.

RadixWallet/Clients/AppEventsClient/AppEventsClient+Live.swift
```diff
@@ -0,0 +1,18 @@
+extension AppEventsClient: DependencyKey {
+	public static let liveValue: AppEventsClient = {
+		@Dependency(\.homeCardsClient) var homeCardsClient
+
+		return .init(
+			handleEvent: { event in
+				switch event {
```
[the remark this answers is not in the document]
matiasbzurovski: I liked this idea a lot, so gave it a try [here](https://github.com/radixdlt/babylon-wallet-ios/pull/1187/commits/e1cf6f9e96b267ce9864e2c247e2c7dafad06119). However, there seems to be a race condition error since the `.appStarted` event is never received by the listener. I even added [this](https://github.com/radixdlt/babylon-wallet-ios/pull/1187/commits/e1cf6f9e96b267ce9864e2c247e2c7dafad06119#diff-4c2af50d3b79c7cb4b1f6a5f61abfbbfa369508a1c44ac6a574379002aced77aR29) handy method to be called by the `AppEventsClient` on its initialization, to make sure the `HomeCardsClient` subscribes to events before the first one is dispatched, but problem was still present.

We can work on some delays but even having the `AppsEventsClient` _starting_ the `HomeCardsClient` (and whatever other listener client we would have in the future) goes against the whole reasoning behind this change.

RadixWallet/Clients/AppsFlyerClient/AppsFlyerClient+Live.swift
```diff
@@ -37,16 +37,13 @@ extension AppsFlyerClient: DependencyKey {
...
 				loggerGlobal.info("did resolve deep link. Is deferred: \(deepLink.isDeferred). Click events: \(deepLink.clickEvent)")
-				if deepLink.isDeferred {
-					let message = if let deepLinkValue = deepLink.clickEvent["deep_link_value"] as? String {
-						"Resolved deferred DL with value \(deepLinkValue)"
-					} else {
-						"Resolved deferred DL without value"
-					}
-					AppsFlyerLib.shared().logEvent(message, withValues: deepLink.clickEvent)
+				if deepLink.isDeferred, let value = deepLink.clickEvent["deep_link_value"] as? String {
+					appEventsClient.handleEvent(.deepLinkReceived(value))
```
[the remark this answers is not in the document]
matiasbzurovski: mm I am not sure I prefer it this way, since `DeepLinkHandlerClient` is about non deferred deep links. The logic for deferred vs non deferred is totally different, so not sure if it wouldn't add more confusion to put it there just for the sake of being a deep link.

RadixWallet/Clients/AppEventsClient/AppEventsClient+Live.swift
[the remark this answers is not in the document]
matiasbzurovski: resolved with `AsyncReplaySubject`

RadixWallet/Clients/HomeCardsClient/HomeCardsClient+Interface.swift
```diff
@@ -0,0 +1,37 @@
+// MARK: - HomeCardsClient
+public struct HomeCardsClient: Sendable {
+	public var cards: Cards
+	public var removeCard: RemoveCard
+	public var walletStarted: WalletStarted
+	public var walletCreated: WalletCreated
+	public var deepLinkReceived: DeepLinkReceived
```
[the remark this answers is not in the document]
matiasbzurovski: ignored as per changes to AppEventClient

RadixWallet/Features/CardCarousel/CardCarousel+View.swift
kugel3: The problem with that is that swiftui will not be able to handle insertion and removal animations properly if the the index is the id. For example if you have four cards and remove the third card, with id and index 2, the collection will still have a card with id 2, but it will be the one that used to have id 3. In short, using index as id with a dynamic collection is not supported in SwiftUI. That is also why I called my helper type that wraps up the functionality you use here, "ForEachStatic".

RadixWallet/Features/CardCarousel/CardCarousel+View.swift
kugel3: What a shame, I spent at least 75% of the time on getting teaser cards to work properly. Plus I think it looks much better an indicator, that feels pretty stale nowadays.

RadixWallet/Features/CardCarousel/CardCarousel+View.swift
matiasbzurovski: We aren't using the `index` as the `id`, we are using the `\.element` to identify them.

We are using the `index` to tag them, and that works correctly on your example since after removing card with `index: 2`, the index (and tag) is properly updated with the new set of elements.

RadixWallet/Features/CardCarousel/CardCarousel+View.swift
matiasbzurovski: I also liked the teaser approach better, but it was showing a glitch on first load. Also, solution adds a lot of noise to the code to achieve something not natively supported.

RadixWallet/Features/CardCarousel/CardCarousel+View.swift
kugel3: Ah, alright, so you use element for id, and index only for the tag, for the benefit of the tabview. 

I would still consider using the same tag and id, and map it in the reducer, since in there you have access to both the elements and their offsets.

RadixWallet/Features/CardCarousel/CardCarousel+View.swift
kugel3: I never saw any glitch, was that present in my branch? That should be trivial to fix. I can wrap it all up in a TeasedPagerView when I get back.

RadixWallet/Features/CreateAccount/Coordinator/CreateAccountCoordinator+Reducer.swift
```diff
@@ -191,6 +192,9 @@ extension CreateAccountCoordinator {
 			return .none
 
 		case let .handleAccountCreated(.success(account)):
+			if state.config.shouldSendWalletCreatedEvent {
+				appEventsClient.handleEvent(.walletCreated)
+			}
```
GhenadieVP, line 197: This event should be added when the onboarding is completed in App+Reducer, as by definition the onboarding happens when user creates a new Wallet.

matiasbzurovski: But in such case, it could be because the user restored a wallet

GhenadieVP: Correct, then it would be better to have this code in OnboardingCoordinator when the CreateAccount completes.

GhenadieVP (approved):

PR #1190, opened 2024-06-28
Proof that SargonOS does not introduce non-trivial breaking changes

CyonAlexRDX:
> [!IMPORTANT]
> Cannot be merged until SargonOS PR https://github.com/radixdlt/sargon/pull/176 has been merged.
> Merge SargonOS, then update this, **which uses LOCAL Sargon**, and use new version.

It fixes the breaking changes introduce by it - non are hard, easy changes.

The biggest change is really that DeviceFactorSource creation method takes a `DeviceInfo` to populate the `DeviceFactorSourceHint`. *This is not important*, and I dont thin we even display this to the user. 

This PR could/should be clean up! And can be cleaned up in several ways, but I have no time to do it. 

One idea is to create a new temporary method on `SecureStorageClient` `loadDeviceInfoOrFallback` or similar, which is a *non*-throwing function which returns a non-optional `DeviceInfo` with description "iPhone"/.

Once again, this info is not at all crucial. Feel free to use whatever workaround suits you!

Or a bit more elegant, is to change declare an initialiser in *Swift Sargon* which takes an *optional* `DeviceInfo`...

I gave it a quick spin and it seems to work fine! (I got some iCloud error, but it is most likely completely unrelated)

https://github.com/radixdlt/babylon-wallet-ios/assets/116169792/cd32f5b0-cce0-4c0f-8511-0f47d193d2aa


RadixWallet/Features/DebugInspectProfile/InspectProfile+View.swift
```diff
@@ -181,6 +181,7 @@ extension DebugInspectFactorSourceView {
 				DeviceFactorSouceView(deviceFactorSource: deviceFactorSource)
 			case let .ledger(ledgerFactorSource):
 				LedgerFactorSourceView(ledgerFactorSource: ledgerFactorSource)
+			default: fatalError("DISCREPANCY: Found non .device | .ledger factor source. A real world user cannot possible have this.")
```
Sajjon, line 184: This is safe for our users, it is not possible for our users to create any other factor source than `DeviceFactorSource` or `LedgerFactorSource`.

RadixWallet/Features/DerivePublicKeys/DerivePublicKeys.swift
```diff
@@ -89,6 +89,8 @@ public struct DerivePublicKeys: Sendable, FeatureReducer {
 						.device
 					case let .ledger(ledger):
 						.ledger(ledger)
+					default:
+						fatalError("DISCREPANCY: Found non .device | .ledger factor source. A real world user cannot possible have this.")
```
Sajjon, line 93: This is safe for our users, it is not possible for our users to create any other factor source than `DeviceFactorSource` or `LedgerFactorSource`.

RadixWallet/Features/DerivePublicKeys/DerivePublicKeys.swift
```diff
@@ -210,6 +212,7 @@ public struct DerivePublicKeys: Sendable, FeatureReducer {
 					)
 				case let .ledger(ledgerFactorSource):
 					return deriveWith(ledgerFactorSource: ledgerFactorSource, state)
+				default: fatalError("DISCREPANCY: Found non .device | .ledger factor source. A real world user cannot possible have this.")
```
Sajjon, line 215: This is safe for our users, it is not possible for our users to create any other factor source than `DeviceFactorSource` or `LedgerFactorSource`.

RadixWallet/Features/Signing/Coordinator/Signing.swift
```diff
@@ -195,8 +195,7 @@ public struct Signing: Sendable, FeatureReducer {
 				signingFactors: nextFactors,
 				signingPurposeWithPayload: signingPurposeWithPayload
 			)
-		case .offDeviceMnemonic, .securityQuestions, .trustedContact:
-			fatalError("Implement me")
+		default: fatalError("DISCREPANCY: Found non .device | .ledger factor source. A real world user cannot possible have this.")
```
Sajjon, line 198: This is safe for our users, it is not possible for our users to create any other factor source than `DeviceFactorSource` or `LedgerFactorSource`.

RadixWallet/WalletApp.swift
```diff
@@ -6,15 +6,6 @@ import SwiftUI
 struct WalletApp: SwiftUI.App {
 	@UIApplicationDelegateAdaptor var delegate: AppDelegate
 
-	init() {
-		#if DEBUG
-		// MUST NOT be called twice (panics...)
-		// MUST NOT be enabled in PROD (not ensured that secrets are omitted)
-		// This is VERY VERY verbose, so mosly useful for debugging.
-		enableLoggingFromRust()
```
Sajjon, line 14: We should use SargonOS loggin systems instead. So I removed that freestanding enable log function in Sargon.

GhenadieVP (approved):

PR #1194, opened 2024-07-02
[ABW-3254] Account resources background refresh

danvleju-rdx:
Jira ticket: [ABW-3254](https://radixdlt.atlassian.net/browse/ABW-3254)

## Description
This PR implements the logic for background refresh of account resources on the Home screen and Account Details screen.

The following reload use cases were added:
- When a user navigates to a specific account, that account's balances are immediately reloaded.
- When a user stays on an account screen with the wallet open, that account's balances are reloaded every _1_ minute.
- When a user returns to the Home screen and keeps the wallet open, all account balances are reloaded every _5_ minutes.

## How to test
The refresh interval can be decreased by adjusting the values of `accountPortfoliosRefreshIntervalInSeconds` and `accountPortfolioRefreshIntervalInSeconds`.

[ABW-3254]: https://radixdlt.atlassian.net/browse/ABW-3254?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


RadixWallet/Features/AccountDetailsFeature/Coordinator/AccountDetails+Reducer.swift
```diff
@@ -277,4 +289,18 @@ public struct AccountDetails: Sendable, FeatureReducer {
 			}
 		}
 	}
+
+	private func scheduleFetchAccountPortfolioTimer(_ state: State) -> Effect<Action> {
+		.run { [address = state.account.address] _ in
+			await withTaskCancellation(id: CancellableId.fetchAccountPortfolio, cancelInFlight: true) {
```
matiasbzurovski, line 295: Is there any difference between this `withTaskCancellation` and adding `.cancellable(id: CancellableId.fetchAccountPortfolio, cancelInFlight: true)` to the `.run` effect?

RadixWallet/Features/AssetsFeature/Components/FungibleAssetList/FungibleAssetList+Reducer.swift
```diff
@@ -5,6 +5,20 @@ import SwiftUI
 public struct FungibleAssetList: Sendable, FeatureReducer {
 	public struct State: Sendable, Hashable {
 		public var sections: IdentifiedArrayOf<FungibleAssetList.Section.State> = []
+
+		public mutating func update(
+			token: OnLedgerEntity.OwnedFungibleResource,
+			for sectionID: Section.State.ID
+		) -> FungibleAssetList.Section.Row.State? {
+			guard
+				let xrdSection = sections.first(where: { $0.id == sectionID }),
```
matiasbzurovski, line 14: this should probably be named `section` since it doesn't need to be for XRD

matiasbzurovski (approved):
Looking great! one small comment only on the code.

One thing I had discussed with @kugel3 in the past is that we have many reducers that could probably be removed. For example, each of the children displayed on the `AssetsView` (`FungibleAssetList`, `NonFungibleAssetList`, `StakeUnitList` & `PoolUnitsList`) have its own reducer and state, and in 3 of these cases they are just about displaying data and calling delegate when selected.

The reason I am bringing this up is because now that we want to perform updates on the parent (`AssetsView`), it makes even less sense to have the children with their state allowing a mutating func from given class. It would make more sense in my opinion to just handle the updates in one place.

danvleju-rdx:
@matiasbzurovski Yes, we could remove those reducers, but since we plan to add UITableView here, it’s probably not worth refactoring right now. However, I did some refactoring and split the `AssetsView` code into separate files for better readability. Additionally, I moved the update functions to `AssetsView+Update`.

RadixWallet/Features/AccountDetailsFeature/Coordinator/AccountDetails+Reducer.swift
danvleju-rdx: It does the same thing, and yes, `.cancellable` looks cleaner. Updated

RadixWallet/Features/AssetsFeature/AssetsView+Update.swift
```diff
@@ -0,0 +1,178 @@
...
+		self.refresh(from: portfolio, valueChangeMap: { old, new in
+			var old = old
+			old.refresh(from: new)
+			return old
+		})
+	}
+}
+
+extension FungibleAssetList.State {
+	public mutating func update(
```
GhenadieVP, line 143: These `update` functions should rather be called `updated`.

Additionally, the functions which update a row, need to be named `updatedRow`

RadixWallet/Features/HomeFeature/Coordinator/Home.swift
```diff
@@ -420,6 +427,25 @@ public struct Home: Sendable, FeatureReducer {
...
+	public func fetchAccountPortfolios(_ state: State) -> Effect<Action> {
+		let accountAddresses = state.accounts.map(\.address)
+		return .run { _ in
+			_ = try await accountPortfoliosClient.fetchAccountPortfolios(accountAddresses, true)
+		} catch: { error, _ in
+			errorQueue.schedule(error)
+		}
+	}
+
+	public func scheduleFetchAccountPortfoliosTimer(_ state: State) -> Effect<Action> {
```
GhenadieVP, line 440: hmm, need to test it, but this seems that it will start the automatic refresh for the accounts that are in the state at the moment of the task being fired.
What about the case when new accounts are added/hidden? Gateway switch?

Do check the `accountsLoadedResult` internal action.

danvleju-rdx: Good point. Starting the timer on the `accountsLoadedResult` action (and also on the `task` action to ensure it resumes when coming back) will fix the problem.

GhenadieVP (approved):
LGTM!
nice and much awaited improvement.

PR #1199, opened 2024-07-04
Home Cards logic and UI updates

matiasbzurovski:
Depends on https://github.com/radixdlt/sargon/pull/180

## Description
This PR uses Sargon to determine which cards to show on Home carousel.

## Videos

https://github.com/radixdlt/babylon-wallet-ios/assets/164921079/9ac8b8a9-fc88-4136-b693-5048ffd471d0

https://github.com/radixdlt/babylon-wallet-ios/assets/164921079/2775a196-8107-4b62-a7e1-76cb7a28ff8e


RadixWallet/Clients/AppEventsClient/AppEventsClient+Live.swift
```diff
@@ -0,0 +1,18 @@
...
+
+		return .init(
+			handleEvent: { event in
+				switch event {
+				case .appStarted:
+					homeCardsClient.walletStarted()
+				case .walletCreated:
+					homeCardsClient.walletCreated()
+				case let .deepLinkReceived(value):
+					homeCardsClient.deepLinkReceived(value)
```
matiasbzurovski, line 13: This client is just a middle man, so we don't have high level components (such as `AppReducer`), calling a client specific to a subview: `homeCardsClient`.  Also, it could become handy if in the future we need to react to particular app events (for example, configure other clients, log an event, etc). If the approach is accepted, we could even delete the `resetWalletClient` and instead handle a new `AppEvent.resetWallet` here.

Otherwise, we can easily remove it by just reverting this [commit](https://github.com/radixdlt/babylon-wallet-ios/pull/1199/commits/51e16a46fb5a4ed65e3c2cdbde5e9dfc6a245583)

PR #1200, opened 2024-07-05
[ABW-3259] NPS survey upload using body params

danvleju-rdx:
Jira ticket: [ABW-3259](https://radixdlt.atlassian.net/browse/ABW-3259)

## Description
This PR changes the NPS survey upload from using query parameters to body parameters. Refiner requires this to ensure the NPS score is treated as a number instead of a string.

[ABW-3259]: https://radixdlt.atlassian.net/browse/ABW-3259?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/RadixWallet/Clients/GatewayAPI/GatewayAPIClient/GatewayAPIClient+Live.swift
+++ b/RadixWallet/Clients/GatewayAPI/GatewayAPIClient/GatewayAPIClient+Live.swift
@@ -70,7 +70,7 @@ extension GatewayAPIClient {
 				urlRequest.timeoutInterval = timeoutInterval
 			}
 
-			let data = try await httpClient.executeRequest(urlRequest)
+			let data = try await httpClient.executeRequest(urlRequest, nil)
 
 			do {
 				return try jsonDecoder.decode(Response.self, from: data)
--- a/RadixWallet/Clients/GatewayAPI/GatewayAPIClient/GatewayAPIClient+Live.swift
+++ b/RadixWallet/Clients/GatewayAPI/GatewayAPIClient/GatewayAPIClient+Live.swift
@@ -70,7 +70,7 @@ extension GatewayAPIClient {
 				urlRequest.timeoutInterval = timeoutInterval
 			}
 
-			let data = try await httpClient.executeRequest(urlRequest, nil)
+			let data = try await httpClient.executeRequest(urlRequest)
 
 			do {
 				return try jsonDecoder.decode(Response.self, from: data)
--- a/RadixWallet/Clients/HTTPClient/HTTPClient+Interface.swift
+++ b/RadixWallet/Clients/HTTPClient/HTTPClient+Interface.swift
@@ -5,5 +5,11 @@ public struct HTTPClient: Sendable, DependencyKey {
 
 // MARK: HTTPClient.ExecuteRequest
 extension HTTPClient {
-	public typealias ExecuteRequest = @Sendable (_ request: URLRequest, _ isStatusCodeValid: ((Int) -> Bool)?) async throws -> Data
+	public typealias ExecuteRequest = @Sendable (_ request: URLRequest, _ acceptedStatusCodes: [HTTPStatusCode]) async throws -> Data
+}
+
+extension HTTPClient {
+	func executeRequest(_ request: URLRequest) async throws -> Data {
+		try await executeRequest(request, [.ok])
+	}
 }
--- a/RadixWallet/Clients/HTTPClient/HTTPClient+Interface.swift
+++ b/RadixWallet/Clients/HTTPClient/HTTPClient+Interface.swift
@@ -5,5 +5,5 @@ public struct HTTPClient: Sendable, DependencyKey {
 
 // MARK: HTTPClient.ExecuteRequest
 extension HTTPClient {
-	public typealias ExecuteRequest = @Sendable (URLRequest) async throws -> Data
+	public typealias ExecuteRequest = @Sendable (_ request: URLRequest, _ isStatusCodeValid: ((Int) -> Bool)?) async throws -> Data
 }
--- a/RadixWallet/Clients/HTTPClient/HTTPClient+Live.swift
+++ b/RadixWallet/Clients/HTTPClient/HTTPClient+Live.swift
@@ -3,15 +3,14 @@ extension HTTPClient {
 		let session = URLSession.shared
 
 		return .init(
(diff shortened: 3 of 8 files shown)


RadixWallet/Clients/NPSSurveyClient/NPSSurveyClient+Live.swift
```diff
@@ -101,51 +103,37 @@ extension NPSSurveyClient {
...
+		}
+		if let reason = feedback?.reason {
+			feedbackParams[BodyParam.feedbackReason.rawValue] = reason
+		}
+		urlRequest.httpBody = try? JSONSerialization.data(withJSONObject: feedbackParams)
 
 		do {
-			_ = try await httpClient.executeRequest(urlRequest)
+			_ = try await httpClient.executeRequest(urlRequest) { statusCode in
+				[200, 202].contains(statusCode)
```
danvleju-rdx, line 133: This request returns a 202 response code. For other requests, we want to be more restrictive and accept only a 200 response code.

RadixWallet/Clients/HTTPClient/HTTPClient+Live.swift
```diff
@@ -3,14 +3,15 @@ extension HTTPClient {
...
-			executeRequest: { request in
+			executeRequest: { request, isStatusCodeValid in
 				let (data, urlResponse) = try await session.data(for: request)
 
 				guard let httpURLResponse = urlResponse as? HTTPURLResponse else {
 					throw ExpectedHTTPURLResponse()
 				}
 
-				guard httpURLResponse.statusCode == BadHTTPResponseCode.expected else {
+				let isSuccessStatusCode = isStatusCodeValid?(httpURLResponse.statusCode) ?? (httpURLResponse.statusCode == BadHTTPResponseCode.expected)
```
matiasbzurovski, line 13: not part of this PR, but wondering why are we checking the success code from something like `BadHTTPResponseCode` 😅

RadixWallet/Clients/HTTPClient/HTTPClient+Interface.swift
```diff
@@ -5,5 +5,5 @@ public struct HTTPClient: Sendable, DependencyKey {
 
 // MARK: HTTPClient.ExecuteRequest
 extension HTTPClient {
-	public typealias ExecuteRequest = @Sendable (URLRequest) async throws -> Data
+	public typealias ExecuteRequest = @Sendable (_ request: URLRequest, _ isStatusCodeValid: ((Int) -> Bool)?) async throws -> Data
```
matiasbzurovski, line 8: instead of a closure to determine whether a status code is considered valid, I would have it receive an array of accepted status codes. We could use just `Int` or define an enum with the predefined values ([example](https://gist.github.com/ollieatkinson/322338df8a5220d649ac01ff11e7de12)).

RadixWallet/Clients/GatewayAPI/GatewayAPIClient/GatewayAPIClient+Live.swift
```diff
@@ -70,7 +70,7 @@ extension GatewayAPIClient {
 				urlRequest.timeoutInterval = timeoutInterval
 			}
 
-			let data = try await httpClient.executeRequest(urlRequest)
+			let data = try await httpClient.executeRequest(urlRequest, nil)
```
matiasbzurovski, line 73: We can define a helper method to avoid sending nil most of the times
```Swift
extension HTTPClient {
	func executeRequest(_ request: URLRequest) async throws -> Data {
		try await executeRequest(request, nil)
	}
}
```

or if you follow my previous comment:
```Swift
extension HTTPClient {
	func executeRequest(_ request: URLRequest) async throws -> Data {
		try await executeRequest(request, [.ok])
	}
}
```

matiasbzurovski (commented):
Looking good, some small suggestions only.

(Assume test worked fine on NPS dashboard, to which I don't have access)

RadixWallet/Clients/HTTPClient/HTTPClient+Interface.swift
GhenadieVP: Either what @matiasbzurovski suggested, or return the status code along with the Data and let the consumer decide if it is valid or not.

danvleju-rdx: @matiasbzurovski not sure if we need such an ample enum, but an array of `Int` sounds good

RadixWallet/Clients/HTTPClient/HTTPClient+Live.swift
danvleju-rdx: Good question :) 
@GhenadieVP do you think we can move the expected success code elsewhere?

RadixWallet/Clients/HTTPClient/HTTPClient+Interface.swift
matiasbzurovski: we may not want to add all the cases (since we are not mapping the int value to the corresponding case anyway), but if define at least the expected ones (`200` & `202` for now) we could refer it directly (`HttpStatusCode.ok` vs `BadHTTPResponseCode.expected`)

danvleju-rdx: agree

RadixWallet/Clients/HTTPClient/HTTPClient+Live.swift
danvleju-rdx: I would say we can completely remove `expected` and `butExpected` from `BadHTTPResponseCode`

RadixWallet/Core/SharedModels/HTTPURLSessionResponse/HTTPURLSessionResponse.swift
```diff
@@ -33,3 +31,18 @@ public struct ResponseDecodingError: Swift.Error {
 		self.error = error
 	}
 }
+
+// MARK: - HTTPStatusCode
+public enum HTTPStatusCode: Int, Error {
```
GhenadieVP, line 36: 👍

GhenadieVP (approved):

matiasbzurovski (approved):

PR #1201, opened 2024-07-05
Address Details + NFT Name UI fixes

matiasbzurovski:
Jira ticket: [ABW-3484](https://radixdlt.atlassian.net/browse/ABW-3484)
Jira ticket: [ABW-3487](https://radixdlt.atlassian.net/browse/ABW-3487)

## Description
This PR solves some small UI bugs reported.

### Notes
Regarding bug [3484](https://radixdlt.atlassian.net/browse/ABW-3484), the solution isn't perfect. This is, if user drags the modal to the top (and hence extend its height), the QR code will still resize while dragging. This is because we want the QR code to take all the available size, and if you increase the modal size then so does the image size. The only way to have it not doing this would be to either:
- prevent the user from dragging the modal => worse UX
- make the QR image size static => we would need to either make the view scrollable, or the QR code smaller enough to fit in every screen regardless of the address length

I think the current situation is good enough and better than before this change, but will leave some videos for comparisson.

| Before | After |
| -- | -- |
| <video src=https://github.com/radixdlt/babylon-wallet-ios/assets/164921079/56715141-9b66-4f0f-ab60-858746f45ceb> | <video src=https://github.com/radixdlt/babylon-wallet-ios/assets/164921079/cfeb82ea-e6e9-433c-a788-4df45d3ec495>
 |

[ABW-3484]: https://radixdlt.atlassian.net/browse/ABW-3484?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[ABW-3487]: https://radixdlt.atlassian.net/browse/ABW-3487?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/RadixWallet/Core/FeaturePrelude/AddressView/AddressDetails+View.swift
+++ b/RadixWallet/Core/FeaturePrelude/AddressView/AddressDetails+View.swift
@@ -74,10 +74,14 @@ public extension AddressDetails {
 			Group {
 				switch store.qrImage {
 				case let .success(value):
-					Image(decorative: value, scale: 1)
-						.resizable()
-						.aspectRatio(1, contentMode: .fit)
-						.transition(.scale(scale: 0.95).combined(with: .opacity))
+					GeometryReader { proxy in
+						Image(decorative: value, scale: 1)
+							.resizable()
+							.frame(width: proxy.size.height, height: proxy.size.height)
+							.frame(maxWidth: .infinity)
+							.transition(.scale(scale: 0.95).combined(with: .opacity))
+					}
+
 				case .failure:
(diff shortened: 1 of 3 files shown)


danvleju-rdx (approved):
LGTM, doesn't look like a bug anymore!

GhenadieVP (approved):

PR #1217, opened 2024-07-17
[ABW-3522] Create Account flow updates

matiasbzurovski:
Jira ticket: [ABW-3522](https://radixdlt.atlassian.net/browse/ABW-3522)

## Description
This PR makes the following updates:
- Create Account flow is now presented from Onboarding, rather than replacing the root view. This allows the user to exit the flow.
- Small UI updates on `NameAccount` & `Completion` views.
- Updates `AppTextField` to only add grey border when focused (see comments in code).
- Allow `TooggleView` title to have more than one line (so that the Create with Ledger toggle doesn't look trimmed as first video shows). 

## Video
| Before | After |
| - | - |
| <video src=https://github.com/user-attachments/assets/569aa81f-6cc6-47f2-abe2-fd14b2e84e72> | <video src=https://github.com/user-attachments/assets/2d2b82ac-77c9-450a-ad3d-6601cca0f838> |

[ABW-3522]: https://radixdlt.atlassian.net/browse/ABW-3522?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


RadixWallet/Core/DesignSystem/Components/AppTextField.swift
```diff
@@ -44,6 +44,8 @@ public struct AppTextField<FocusValue: Hashable, Accessory: View, InnerAccessory
 	let accessory: Accessory
 	let innerAccesory: InnerAccessory
 
+	@State private var isFocused: Bool = false
```
matiasbzurovski, line 47: My first approach was to use `@Environment(\.isFocused)`, but it didn't work. Therefore, I went with this custom solution that we manually updated the flag with the `onEditingChanged` callback of the `TextField`. 

I think we could be smarter and use our `Focus` and some callback when it changes, but given such property is optional, it would be tricky handling it when `focus = nil`. Open to suggestions if someone comes up with a better idea for this.

RadixWallet/Core/DesignSystem/Components/AppTextField.swift
```diff
@@ -169,7 +173,7 @@ public struct AppTextField<FocusValue: Hashable, Accessory: View, InnerAccessory
 				.cornerRadius(.small2)
 				.overlay(
 					RoundedRectangle(cornerRadius: .small2)
-						.stroke(accentColor(border: !isEnabled), lineWidth: 1)
+						.stroke(accentColor(border: !isFocused), lineWidth: 1)
```
danvleju-rdx, line 176: I think we can reverse the condition for the `border` as we display a darker one when the field is focused: 
```Swift
.stroke(accentColor(border: isFocused), lineWidth: 1)
```
```Swift
	private func accentColor(border: Bool) -> Color {
		switch hint?.viewState.kind {
		case .none:
                     border ? .app.gray1 : .app.gray4
		case .info:
			.app.gray1
		case .error:
			.app.red1
		case .warning:
			.app.alert
		}
	}
```

danvleju-rdx (approved):

GhenadieVP (approved):

PR #1221, opened 2024-07-17
[ABW-3539] UI Adjustments - Account permission

danvleju-rdx:
Jira ticket: [ABW-3539](https://radixdlt.atlassian.net/browse/ABW-3539)

## Description
Updates padding for Account Permission screen.
Account cards updates for `AccountPermissionChooseAccounts` are handled [here](https://github.com/radixdlt/babylon-wallet-ios/pull/1216).

## Screenshot

| - | - |
| Simulator Screenshot - iPhone 15 Pro - 2024-07-17 at 17 45 33 | Simulator Screenshot - iPhone 15 Pro - 2024-07-17 at 17 39 31 |

[ABW-3539]: https://radixdlt.atlassian.net/browse/ABW-3539?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/RadixWallet/Features/DappInteractionFeature/Children/AccountPermission/AccountPermission+View.swift
+++ b/RadixWallet/Features/DappInteractionFeature/Children/AccountPermission/AccountPermission+View.swift
@@ -44,7 +44,7 @@ extension AccountPermission {
 			) { viewStore in
 				GeometryReader { geometry in
 					ScrollView {
-						VStack(spacing: .large1) {
+						VStack(spacing: .zero) {
 							DappHeader(
 								thumbnail: viewStore.thumbnail,
 								title: viewStore.title,
@@ -58,13 +58,14 @@ extension AccountPermission {
 									.padding(.medium1)
 							}
 							.padding(.horizontal, .medium2)
+							.padding(.top, .large1)
 
 							Text(L10n.DAppRequest.AccountPermission.updateInSettingsExplanation)
 								.foregroundColor(.app.gray2)
 								.textStyle(.body1Regular)
 								.multilineTextAlignment(.center)
 								.padding(.horizontal, .medium2)
-								.padding(.top, .large1)
+								.padding(.top, .medium1)
 
 							Spacer()
 						}
--- a/RadixWallet/Features/DappInteractionFeature/Children/AccountPermission/AccountPermission+View.swift
+++ b/RadixWallet/Features/DappInteractionFeature/Children/AccountPermission/AccountPermission+View.swift
@@ -44,7 +44,7 @@ extension AccountPermission {
 			) { viewStore in
 				GeometryReader { geometry in
 					ScrollView {
-						VStack(spacing: .medium2) {
+						VStack(spacing: .large1) {
 							DappHeader(
 								thumbnail: viewStore.thumbnail,
 								title: viewStore.title,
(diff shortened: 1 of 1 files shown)


RadixWallet/Features/DappInteractionFeature/Children/AccountPermission/AccountPermission+View.swift
```diff
@@ -44,7 +44,7 @@ extension AccountPermission {
 			) { viewStore in
 				GeometryReader { geometry in
 					ScrollView {
-						VStack(spacing: .medium2) {
+						VStack(spacing: .large1) {
```
matiasbzurovski, line 47: Not sure if this value is correct since Zeplin is indicating a padding to a wrong part of the view
Screenshot 2024-07-17 at 16 58 17

RadixWallet/Features/DappInteractionFeature/Children/AccountPermission/AccountPermission+View.swift
```diff
@@ -59,13 +59,12 @@ extension AccountPermission {
...
 							.padding(.horizontal, .medium2)
 
-							Spacer()
-
 							Text(L10n.DAppRequest.AccountPermission.updateInSettingsExplanation)
 								.foregroundColor(.app.gray2)
 								.textStyle(.body1Regular)
 								.multilineTextAlignment(.center)
 								.padding(.horizontal, .medium2)
+								.padding(.top, .large1)
```
matiasbzurovski, line 67: Zeplin indicates the distance between the two elements should be 76, would be `.large1` + `36` (which we don't have any `unit * 9`). Still, this distance seems quite arbitrary, so I would confirm with Aftab that is a fixed distance and not that we want it centered (as it currently shows).

matiasbzurovski (commented):
Looks to me we need Zeplin updates before implementing this.

RadixWallet/Features/DappInteractionFeature/Children/AccountPermission/AccountPermission+View.swift
danvleju-rdx: `.large1` from VStack spacing + this `.large1` will be 80. I recall Aftab agreed that we should round to the nearest multiple of 4.
Regarding fixed distance vs centered, I will confirm with Aftab and the Android team.

matiasbzurovski: but 76 is also multiple of 4, that could be the expected distance as well

RadixWallet/Features/DappInteractionFeature/Children/AccountPermission/AccountPermission+View.swift
matiasbzurovski: but are we sure that 69 represents 3 lines? and not just a random box that happens to look like it represents 3 lines from the Text we need to calculate the distance?

I believe we should push for having updated and consistent designs before making these changes. Not only it would be easier for us, but also state the source of truth if implementation differs among platforms.

danvleju-rdx: Fair enough, asked Aftab

RadixWallet/Features/DappInteractionFeature/Children/AccountPermission/AccountPermission+View.swift
danvleju-rdx: Discussed with Aftab and he adjusted the spacing to `.medium1`

GhenadieVP (approved):

PR #1229, opened 2024-07-19
[ABW-3634] Fix bug where Account Details wouldn't show Verify Address on Ledger

matiasbzurovski:
Jira ticket: [ABW-3634](https://radixdlt.atlassian.net/browse/ABW-3634)

## Description
Manually check whether an `AccountForDisplay` is controller by a Ledger device, instead of assuming it isn't.

## Before vs After
| - | - |
| <video src=https://github.com/user-attachments/assets/0f48e878-8b2e-4005-a7be-e31c4ff9994a> | <video src=https://github.com/user-attachments/assets/71c634e9-f926-4b80-b96c-ae6e43116598> |

### DIFF
--- a/RadixWallet/Core/SharedModels/LedgerIdentifiable.swift
+++ b/RadixWallet/Core/SharedModels/LedgerIdentifiable.swift
@@ -33,7 +33,9 @@ public enum LedgerIdentifiable: Sendable {
 // MARK: LedgerIdentifiable.Address
 extension LedgerIdentifiable {
 	public enum Address: Hashable, Sendable, Identifiable {
-		case account(AccountAddress, isLedgerHWAccount: Bool = false)
+		/// `isLedgerHWAccount` indicates if the account is controlled by a Ledger device.
+		/// if the value is nil, it means we don't know.
+		case account(AccountAddress, isLedgerHWAccount: Bool?)
 		case package(PackageAddress)
 		case resource(ResourceAddress)
 		case resourcePool(PoolAddress)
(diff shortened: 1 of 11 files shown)


RadixWallet/Core/SharedModels/LedgerIdentifiable.swift
```diff
@@ -33,7 +33,11 @@ public enum LedgerIdentifiable: Sendable {
 // MARK: LedgerIdentifiable.Address
 extension LedgerIdentifiable {
 	public enum Address: Hashable, Sendable, Identifiable {
-		case account(AccountAddress, isLedgerHWAccount: Bool = false)
+		/// `isLedgerHWAccount` indicates if the account is controlled by a Ledger device.
+		/// - `true`: we know the account is controlled by a Ledger device
+		/// - `false`: either the account isn't controller by a Ledger device or it is an external account and we don't care.
+		/// - `nil`: we don't know if the account is controller by a Ledger device, we should check if needed.
+		case account(AccountAddress, isLedgerHWAccount: Bool?)
```
GhenadieVP, line 40: `isLedgerHWAccount` complicates things more than it helps, and by having it passed from outside of AddressDetails just saves a very quick check in the Profile(even if it is async), additionally very likely that at some point somebody will miss-use this. For the sake of consistency and encapsulation, I would completely remove this `isLedgerHWAccount`, unless there is a use case which I am missing.

GhenadieVP (changes requested):

RadixWallet/Core/SharedModels/LedgerIdentifiable.swift
matiasbzurovski: definitely, I had left it with the optional case as I thought it was used somewhere else, but seems the only usage of it is for this button. Agree 100% that's better to do the quick async check every time.

GhenadieVP (approved):

PR #1230, opened 2024-07-19
[ABW-3630] Fix NFTs collection background refresh

danvleju-rdx:
Jira ticket: [ABW-3630](https://radixdlt.atlassian.net/browse/ABW-3630)

## Description
Fixes an issue where an expanded collection is not updated if it changes after a direct transfer or background refresh.
The solution is to update the collection when the portfolio is updated, but only if the collection is expanded and the `nonFungibleIdsCount` has changed. This ensures the update occurs only when necessary to avoid resetting pagination.

## Video

https://github.com/user-attachments/assets/55351d48-f065-4140-95df-cef99af813c1

[ABW-3630]: https://radixdlt.atlassian.net/browse/ABW-3630?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/RadixWallet/Features/AssetsFeature/Components/NonFungibleAssetList/Components/Row/NonFungbileAssetRow+Reducer.swift
+++ b/RadixWallet/Features/AssetsFeature/Components/NonFungibleAssetList/Components/Row/NonFungbileAssetRow+Reducer.swift
@@ -59,6 +59,7 @@ extension NonFungibleAssetList {
 			}
 
 			case tokensLoaded(TaskResult<TokensLoadResult>)
+			case refreshResources
 		}
 
 		@Dependency(\.onLedgerEntitiesClient) var onLedgerEntitiesClient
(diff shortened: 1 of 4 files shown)


matiasbzurovski (approved):
LGTM

RadixWallet/Features/AssetsFeature/Components/NonFungibleAssetList/Components/Row/NonFungbileAssetRow+Reducer.swift
```diff
@@ -131,6 +132,9 @@ extension NonFungibleAssetList {
 					state.isLoadingResources = false
 				}
 				return .none
+
+			case .refreshResources:
+				return loadResources(&state, pageIndex: 0)
```
GhenadieVP, line 137: I think that `state.nextPageCursor` needs to also be reset to nil if pageIndex is set to zero. And more, shouldn't the `tokens` be reset?

I guess that this update works properly for a single page NFT collection.

GhenadieVP (approved):

PR #1234, opened 2024-07-23
Sargon os v1

GhenadieVP:
SargonOS V1 implementation, recommended for reading to understand the big picture:
- Ticket - https://radixdlt.atlassian.net/browse/ABW-3591
- SargonOS whimsical - https://whimsical.com/sargon-os-integration-J8FN4ScciWkMAWqk9fSWFc.

With this PR SargonOS is fully integrated in the Wallet replacing the ProfileStore. 
Now Profile is fully managed by SargonOS, still all Profile mutations happen on the Wallet side, and then the Wallet tells Sargon to replace the old Profile. This is pretty much what ProfileStore was doing previously.


RadixWallet/Clients/FactorSourcesClient/FactorSourcesClient+Interface.swift
```diff
@@ -265,14 +265,21 @@ extension FactorSourcesClient {
...
+		@Dependency(\.secureStorageClient) var secureStorageClient
+
+		let deviceInfo = secureStorageClient.loadDeviceInfoOrFallback()
 		let factorSource = switch onDeviceMnemonicKind {
 		case let .babylon(isMain):
 			DeviceFactorSource.babylon(
 				mnemonicWithPassphrase: mnemonicWithPassphrase,
-				isMain: isMain
+				isMain: isMain,
+				deviceInfo: deviceInfo
```
CyonAlexRDX, line 276: Was this not changed to be named to `hostInfo`.

I think a merge main into Sargon, and update iOS host code is needed? Would make review easier :)

RadixWallet/Clients/CloudBackupClient/CloudBackupClient+Live.swift
```diff
@@ -234,14 +228,13 @@ extension CloudBackupClient {
...
 			},
 			migrateProfilesFromKeychain: {
-				let activeProfile = await profileStore.profile.id
 				guard let headerList = try secureStorageClient.loadProfileHeaderList() else { return [] }
 
 				let previouslyMigrated = userDefaults.getMigratedKeychainProfiles
 
 				let migratable = try headerList.compactMap { header -> (Data, Profile.Header)? in
 					let id = header.id
-					guard !previouslyMigrated.contains(id), header.id != activeProfile else { return nil }
```
GhenadieVP, line 244: This was put in place to avoid uploading empty profiles which where created right when the Wallet was opened. Not needed anymore, as now the new Profile is created and saved only when user goes to the New Wallet user flow.

RadixWallet/Clients/TransportProfileClient/TransportProfileClient+Live.swift
```diff
@@ -10,14 +10,12 @@ extension TransportProfileClient: DependencyKey {
 		@Dependency(\.cloudBackupClient) var cloudBackupClient
 
 		return Self(
-			importProfile: { profile, factorSourceIDs, containsP2PLinks in
+			importProfile: { profile, factorSourceIDs, skippedMainBdfs, containsP2PLinks in
 				do {
-					var profile = profile
-					await profileStore.claimOwnership(of: &profile)
```
GhenadieVP, line 16: ownership is claimed in Sargon on import.

RadixWallet/Features/AccountRecoveryScan/Coordinator/AccountRecoveryScanCoordinator.swift
```diff
@@ -158,12 +158,10 @@ public struct AccountRecoveryScanCoordinator: Sendable, FeatureReducer {
 		case let .createProfile(privateHD):
 			let recoveredAccountAndBDFS = AccountsRecoveredFromScanningUsingMnemonic(
 				accounts: sortedAccounts,
-				deviceFactorSource: privateHD.factorSource
+				factorSource: privateHD
 			)
 			return .run { send in
 				let result = await TaskResult<EqVoid> {
-					try secureStorageClient.saveMnemonicForFactorSource(privateHD)
```
GhenadieVP, line 165: Mnemonic will be forwarded to Sargon, and then Sargon will save it to secure storage.

RadixWallet/Features/ProfileBackupsFeature/RestoreProfileFromBackup/Children/ImportSeedPhrasesFlow/ImportMnemonicControllingAccounts.swift
```diff
@@ -173,19 +173,8 @@ public struct ImportMnemonicControllingAccounts: Sendable, FeatureReducer {
 			return .none
 
 		case .confirmSkippingBDFS(.delegate(.confirmed)):
-			loggerGlobal.notice("Skipping BDFS! Generating a new one and hiding affected accounts/personas.")
 			state.destination = nil
-			return .run { [entitiesControlledByFactorSource = state.entitiesControlledByFactorSource] send in
-				loggerGlobal.info("Generating mnemonic for new main BDFS")
-				let newMainBDFS = try await factorSourcesClient.createNewMainBDFS()
```
GhenadieVP, line 180: BDFS will be created by Sargon when the profile is imported and the skippedBDFS flag is true

RadixWallet/Features/ProfileBackupsFeature/RestoreProfileFromBackup/Children/ImportSeedPhrasesFlow/ImportMnemonicsFlowCoordinator.swift
```diff
@@ -255,17 +255,11 @@ public struct ImportMnemonicsFlowCoordinator: Sendable, FeatureReducer {
 			return .none
 		} else {
 			state.destination = nil
-			return .run { [skipped = state.skipped, imported = state.imported, newMainBDFS = state.newMainBDFS, context = state.context] send in
-				if let newMainBDFS, !context.isFromOnboarding {
-					try await factorSourcesClient.saveNewMainBDFS(newMainBDFS)
```
GhenadieVP, line 260: new BDFS will be [saved by Sargon](https://github.com/radixdlt/sargon/blob/6512769d46a954b1c16cf6d06bf6e921e567098b/crates/sargon/src/system/sargon_os/sargon_os.rs#L142)

RadixWallet/Clients/ProfileStore/ProfileStore.swift
```diff
@@ -61,6 +61,16 @@ extension ProfileStore {
 	}
 }
 
+extension ProfileStore {
+	public func isThisDevice(deviceID: DeviceID) async -> Bool {
```
CyonAlexRDX, line 65: minor but I'd there is a possibility of "false negative" here, and thus might be more correct to call it `isKnownToBeThisDevice` or return `Bool?` and return `nil`  inside the guard. But yeah really minor...

CyonAlexRDX: not your code...

RadixWallet/Clients/ProfileStore/ProfileStore.swift
```diff
@@ -1,235 +1,99 @@
...
-/// an empty account list.
-///
-/// And then a lot of sugar/convenience AsyncSequences using `values` but mapping to other
-/// values inside of `Profile`, e.g.:
-///
-/// 	func accountValues() async -> AnyAsyncSequence<Accounts>
-///
-/// And similar async sequences.
-///
 public final actor ProfileStore {
```
CyonAlexRDX, line 4: maybe a word about future directions? In the future this will be deleted right?

RadixWallet/Features/ProfileBackupsFeature/RestoreProfileFromBackup/Children/RecoverWalletWithoutProfile/Coordinator/RecoverWalletWithoutProfileCoordinator.swift
```diff
@@ -131,21 +135,20 @@ public struct RecoverWalletWithoutProfileCoordinator: Sendable, FeatureReducer {
...
-					isMain: true,
-					hostInfo: .current()
-				)
-
-				let privateHD = PrivateHierarchicalDeterministicFactorSource(
-					mnemonicWithPassphrase: mnemonicWithPassphrase,
-					factorSource: mainBDFS
-				)
+				return .run { send in
+					let mainBDFS = await DeviceFactorSource.babylon(
```
CyonAlexRDX, line 139: @GhenadieVP hmm it looks like Sargon does not validate / assert the length of the mnemonic for `DeviceFactorSource.babylon`? I briefly looked through the call chain and looks like no validation happens? we should probably add it, making Sargon return `Result<Self> here: https://github.com/radixdlt/sargon/blob/main/crates/sargon/src/profile/v100/factors/factor_sources/device_factor_source/device_factor_source.rs#L78-L91

CyonAlexRDX (approved):
Nice LOC delta! Good job!

RadixWallet/Features/SplashFeature/Splash.swift
```diff
@@ -195,6 +210,22 @@ public struct Splash: Sendable, FeatureReducer {
 		}
 	}
 
+	private func boot_sargon_os() -> Effect<Action> {
```
matiasbzurovski, line 213:
```suggestion
	private func bootSargonOS() -> Effect<Action> {
```

RadixWallet/Features/SplashFeature/Splash.swift
```diff
@@ -82,8 +82,25 @@ public struct Splash: Sendable, FeatureReducer {
 	public func reduce(into state: inout State, viewAction: ViewAction) -> Effect<Action> {
 		switch viewAction {
 		case .appeared:
-			return .run { send in
-				let isAdvancedLockEnabled = await onboardingClient.loadProfile().appPreferences.security.isAdvancedLockEnabled
+			switch state.context {
+			case .appStarted:
+				return boot_sargon_os().concatenate(with: loadAdvancedLockState())
```
matiasbzurovski, line 87: Shouldn't the boot take place only on `.task`? In the end it is the same result since we ignore the `SargonOSAlreadyBooted` error, but would be cleaner to only boot it once and, if we ever get this error, at least log it.

matiasbzurovski (commented):
working smooth so far! Will continue doing manual tests

RadixWallet/Clients/ProfileStore/ProfileStore.swift
GhenadieVP: Added.

> In the future this will be deleted right?

Likely, we would potentially call Sargon directly in the respective clients, instead of going through the ProfileStore.

RadixWallet/Features/ProfileBackupsFeature/RestoreProfileFromBackup/Children/RecoverWalletWithoutProfile/Coordinator/RecoverWalletWithoutProfileCoordinator.swift
GhenadieVP: Good catch!
Did we ever have this kind of validation?

RadixWallet/Clients/CloudBackupClient/CloudBackupClient+Live.swift
```diff
@@ -188,13 +188,7 @@ extension CloudBackupClient {
...
 			guard shouldBackUp || shouldReclaim else { return }
 
-			var profileToUpload = profile
-			if shouldReclaim {
-				// The profile will already be locally claimed, but we want to update the lastUsedOnDevice date
-				await profileStore.claimOwnership(of: &profileToUpload)
-			}
-
-			try? await backupProfileAndSaveResult(profileToUpload, existingRecord: existingRecord)
+			try? await backupProfileAndSaveResult(profile, existingRecord: existingRecord)
```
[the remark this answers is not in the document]
GhenadieVP: good catch forgot to stage the claiming code.
The claim for this case does not happen in Sargon, since local Profile by definition claimed. It is needed to update the profile from the cloud to point to current device.

RadixWallet/Features/SplashFeature/Splash.swift
GhenadieVP: You mean to use task and not have the check of the `context`?

matiasbzurovski: yes

matiasbzurovski (changes requested):
Found a failing flow, following these steps:
- Run fresh install
- Restore Wallet from backup but skip seed phrase entry
- Try to create a new account

It should allow to create an account with new mnemonics, while it is failing to do so. 

image

One thing I also noticed, is that when restoring such profile it loads the wallet with stokenet  selected as current gateway, while on main version with same profile it loads on mainnet.

RadixWallet/Features/SplashFeature/Splash.swift
GhenadieVP: Splash feature can be instantiated more than once, at app start and then on each foregrounding, this was added as part of Advanced Lock feature.

GhenadieVP:
@matiasbzurovski I did identify the issue as well with my last testing, and fixed it.

matiasbzurovski (changes requested):
Issue regarding creating accounts after restoring profile without entering seed phrase is now solved ✅ 

Just so we don't loose track of it, pasting an issue reported by Umair on Slack:
> the app crashes even when tapping a resource to bring up the assetdetails screen

https://github.com/user-attachments/assets/82752f13-c27a-40b3-b6b2-001eb3b498c0

PR #1235, opened 2024-07-23
[ABW-3650] Security state incorrectly updating

matiasbzurovski:
Jira ticket: [ABW-3650](https://radixdlt.atlassian.net/browse/ABW-3650)

## Description
This PR solves an issue where a security problem would up for an instant when toggling the balances visibility. 

### Notes
This issue was introduced when we fixed another similar on https://github.com/radixdlt/babylon-wallet-ios/pull/1207. Basically, we were considering `.started` backups as failures, which helped solved the previous situation. 
However, we shouldn't consider `.started` backups as failures when the last backup was a `.success`: they should only be considered failures when the previous one was a `.failure`. Therefore, the solution is to just ignore `.started` backups for this check.

### Before vs After
| Before | After |
| -- | -- |
| <video src=https://github.com/user-attachments/assets/12eaf8f7-953f-4d6f-9ad3-37969571862e> | <video src=https://github.com/user-attachments/assets/46fbdb0a-7813-48bb-a34f-fac5edb4f4db> |

[ABW-3650]: https://radixdlt.atlassian.net/browse/ABW-3650?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/RadixWallet/Clients/SecurityCenterClient/SecurityCenterClient+Live.swift
+++ b/RadixWallet/Clients/SecurityCenterClient/SecurityCenterClient+Live.swift
@@ -38,6 +38,10 @@ extension SecurityCenterClient {
 		func cloudBackups() async -> AnyAsyncSequence<BackupStatus?> {
 			let profileID = await profileStore.profile.id
 			let backups = userDefaults.lastCloudBackupValues(for: profileID)
+				.filter { backup in
+					backup?.isFinal ?? true
+				}
+				.eraseToAnyAsyncSequence()
 			return await statusValues(results: backups)
 		}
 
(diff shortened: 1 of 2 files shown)


danvleju-rdx (approved):

GhenadieVP:
To understand those issues better, 
it seems that [this removal](https://github.com/radixdlt/babylon-wallet-ios/blob/f3767b7521e4ceea9f0e8d7167dc39d7169df344/RadixWallet/Clients/CloudBackupClient/CloudBackupClient%2BLive.swift#L167) is root cause that triggers those inconsistencies. It feels a bit odd to end up in the state of no backup when it is removed right before another backup is started, shouldn't the old backup info be rather replaced than removed and then added again 🤔?

matiasbzurovski:
@GhenadieVP not really, the removal is only done if the latest backup failed. For the case this PR fixes, no backup has failed and we still had this issue because a backup was set as `.started` and we considered that an error.

GhenadieVP (approved):

PR #1236, opened 2024-07-23
[ABW-3651] Fix RUID NFTs local id abbreviation

danvleju-rdx:
Jira ticket: [ABW-3651](https://radixdlt.atlassian.net/browse/ABW-3651)

## Description
This PR fixes the abbreviation for NFT's localID of an RUID type.
Formatting is handled in [Sargon](https://github.com/radixdlt/sargon/blob/b0f6c8948eb17237414f1bd34ca776fec71a4858/src/profile/v100/address/non_fungible_local_id.rs#L61).

[ABW-3651]: https://radixdlt.atlassian.net/browse/ABW-3651?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/RadixWallet/Features/AssetsFeature/Components/NonFungibleAssetList/Components/Row/NonFungibleAssetRow+View.swift
+++ b/RadixWallet/Features/AssetsFeature/Components/NonFungibleAssetList/Components/Row/NonFungibleAssetRow+View.swift
@@ -100,7 +100,7 @@ extension NonFungibleAssetList.Row.View {
 
 				HStack {
 					NFTIDView(
-						id: asset.id.nonFungibleLocalId.toUserFacingString(),
+						id: asset.id.nonFungibleLocalId.formatted(.default),
 						name: asset.data?.name,
 						thumbnail: asset.data?.keyImageURL
 					)
--- a/RadixWallet/Features/AssetsFeature/Components/NonFungibleAssetList/Components/Row/NonFungibleAssetRow+View.swift
+++ b/RadixWallet/Features/AssetsFeature/Components/NonFungibleAssetList/Components/Row/NonFungibleAssetRow+View.swift
@@ -100,7 +100,7 @@ extension NonFungibleAssetList.Row.View {
 
 				HStack {
 					NFTIDView(
-						id: asset.id.nonFungibleLocalId.formatted(.default),
+						id: asset.id.nonFungibleLocalId.formatted(),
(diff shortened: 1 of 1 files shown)


RadixWallet/Features/AssetsFeature/Components/NonFungibleAssetList/Components/Row/NonFungibleAssetRow+View.swift
```diff
@@ -100,7 +100,7 @@ extension NonFungibleAssetList.Row.View {
 
 				HStack {
 					NFTIDView(
-						id: asset.id.nonFungibleLocalId.toUserFacingString(),
+						id: asset.id.nonFungibleLocalId.formatted(.default),
```
GhenadieVP, line 103: Is this the only place? I believe we should properly abreviate everywhere?

danvleju-rdx: In `AddressView` it is already formatted correctly. Do we display `nonFungibleLocalId` in any other places?

matiasbzurovski, line 103: minor: we can just do `formatted()` since it defaults to `.default`

matiasbzurovski (approved):

RadixWallet/Features/AssetsFeature/Components/NonFungibleAssetList/Components/Row/NonFungibleAssetRow+View.swift
matiasbzurovski: This is the only display usage of `toUserFacingString()`, so I believe we don't need to update it anywhere else.

So if we have any other incorrect abbreviation, it is because we are using an incorrect `AddrressFormat`, which I don't recall of any case where this is happening.

GhenadieVP: yeah looks good, just tested.

GhenadieVP (approved):

PR #1238, opened 2024-07-24
[ABW-3649] Improve the address details actions to support increased font size

danvleju-rdx:
Jira ticket: [ABW-3649](https://radixdlt.atlassian.net/browse/ABW-3649)

## Description
- Removed unnecessary horizontal padding for actions
- Added multiline stack for cases when font size is too big
- Fixed icons

### Notes
As shown in screenshot 3, since the screen is not scrollable, the QR code becomes smaller as the font size increases.

## Screenshot

| 1 | 2 | 3 |
| - | - | - |
| IMG_0032 | IMG_0030 | IMG_0031 |

[ABW-3649]: https://radixdlt.atlassian.net/browse/ABW-3649?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


RadixWallet/Core/FeaturePrelude/AddressView/AddressDetails+View.swift
```diff
@@ -94,22 +94,35 @@ public extension AddressDetails {
 			.animation(.easeInOut, value: store.qrImage.isSuccess)
 		}
 
+		@ViewBuilder
 		private var actions: some SwiftUI.View {
-			HStack(spacing: .large3) {
-				Button(L10n.AddressDetails.copy, image: .copy) {
+			FlowLayout(multilineAlignment: .center, spacing: .large3) {
+				actionButton(L10n.AddressDetails.copy, image: .copyMedium) {
```
matiasbzurovski, line 100: If we are already resizing the image, why not use `copy-big` (that is 24x24`) and avoid adding a new asset?

RadixWallet/Core/FeaturePrelude/AddressView/AddressDetails+View.swift
```diff
@@ -94,22 +94,35 @@ public extension AddressDetails {
 			.animation(.easeInOut, value: store.qrImage.isSuccess)
 		}
 
+		@ViewBuilder
 		private var actions: some SwiftUI.View {
-			HStack(spacing: .large3) {
-				Button(L10n.AddressDetails.copy, image: .copy) {
+			FlowLayout(multilineAlignment: .center, spacing: .large3) {
```
matiasbzurovski, line 99: I believe we should decrease this spacing as well.

matiasbzurovski (changes requested):
Not part of this, but given we are fixing the view for larger font types I believe we should also make sure that the address is fully displayed on bigger fonts.
This is, edit the `fullAddress` on line `154` to be as this:
```Swift
Text(colorisedText)
    .fixedSize(horizontal: false, vertical: true)
```

This way it look like the image on the right, and not like the one on the left

RadixWallet/Core/FeaturePrelude/AddressView/AddressDetails+View.swift
danvleju-rdx: Hmm, not sure how much. Reducing this space will deviate from the design.

GhenadieVP (approved):

matiasbzurovski (approved):

PR #1249, opened 2024-07-26
Stacked cards effect

matiasbzurovski:
## Description
This PR shows an attempt to have the stacked cards effect on the NFT cards when they are not expanded.

### Notes
If we decide to move forward with this approach, we should use proper assets (and not a screenshot without the background removed 😅).

## Video
[Video](https://drive.google.com/file/d/1YbCZZpr64Nggu_mHrNbXtZArjyyBhnNw/view)


kugel3:
If the idea is to show a picture of stacked cards I don't think that's a good idea. Why not just put dummy cards there in that case?

kugel3:
The same way you put an image there... This puts a photo behind the card, you can also just put some empty cards behind the card. But ideally I would just animate the layout, and use actual cards. That's how this used to work.

matiasbzurovski:
I couldn't find a way of doing what you say here (similar to what we do on New Account completion view), given in this case we are putting elements inside a `List`. Please give it a try if you want, I believe what you had in mind used to work when the elements where inside a `VStack` but I may be missing something.

RadixWallet/Features/AssetsFeature/Components/NonFungibleAssetList/Components/Row/NonFungibleAssetRow+View.swift
```diff
@@ -70,14 +71,38 @@ extension NonFungibleAssetList.Row.View {
...
+
+	var rowBackground: some View {
+		Group {
+			switch stackedCards {
+			case 0:
+				Color.white
+			case 1:
+				Image(.listBackgroundSimple).resizable()
+			default:
+				Image(.listBackgroundMultiple).resizable()
```
kugel3, line 100: This could just be a view containing two empty cards, rather than an image. But as I said, we should do it the proper way. Probably this list should be a UICollectionView anyway.

RadixWallet/Features/AssetsFeature/Components/NonFungibleAssetList/Components/Row/NonFungibleAssetRow+View.swift
```diff
@@ -70,14 +71,38 @@ extension NonFungibleAssetList.Row.View {
...
+			0
+		} else if self.resource.nonFungibleIdsCount == 1 {
+			1
+		} else {
+			2
+		}
+	}
+
+	var rowBackground: some View {
+		Group {
```
kugel3, line 93: Is there a particular reason to not just annotate this with `@ViewBuilder` and skip the Group?

RadixWallet/Features/AssetsFeature/Components/NonFungibleAssetList/Components/Row/NonFungibleAssetRow+View.swift
matiasbzurovski: Yes, the idea is to move this to a UIKit collection.

I tried making this view two empty cards, it didn't work as it wouldn't "exceed" the original card bounds.

RadixWallet/Features/AssetsFeature/Components/NonFungibleAssetList/Components/Row/NonFungibleAssetRow+View.swift
matiasbzurovski: no, same result

PR #1253, opened 2024-07-30
[ABW-3584] UI adjustments - Gateways list

danvleju-rdx:
Jira ticket: [ABW-3584](https://radixdlt.atlassian.net/browse/ABW-3584)

## Description
This PR updates the UI for Gateways screen

## Screenshot

| Screen Before | Screen After |
| - | - |
| IMG_0046 | IMG_0043 |
| IMG_0047 | IMG_0044 |

[ABW-3584]: https://radixdlt.atlassian.net/browse/ABW-3584?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


RadixWallet/Features/GatewaySettingsFeature/Children/AddNewGateway/AddNewGateway+View.swift
```diff
@@ -26,12 +26,23 @@ extension AddNewGateway {
...
 		public init(store: StoreOf<AddNewGateway>) {
 			self.store = store
 		}
 
 		public var body: some SwiftUI.View {
+			content
+				.withNavigationBar {
+					dismiss()
+				}
+				.presentationDetents([.fraction(0.55)])
```
kugel3, line 40: Let's make `0.55` at least a private var

RadixWallet/Features/GatewaySettingsFeature/Children/AddNewGateway/AddNewGateway+View.swift
```diff
@@ -63,11 +74,14 @@ extension AddNewGateway {
...
 						)
+						.padding(.top, .small3)
 						.textInputAutocapitalization(.never)
 						.keyboardType(.URL)
 						.autocorrectionDisabled()
 					}
-					.padding([.bottom, .horizontal], .medium1)
+					.padding(.top, .medium3)
+					.padding(.bottom, .medium1)
+					.padding(.horizontal, .large2)
```
kugel3, line 84: Very tiny style comment: we usually put them in this order: top, horizontal (or leading and trailing), bottom

RadixWallet/Features/GatewaySettingsFeature/Components/GatewayList/GatewayList+View.swift
```diff
@@ -14,22 +12,32 @@ extension GatewayList {
...
+						content: { store in
+							VStack(spacing: .zero) {
+								GatewayRow.View(store: store)
+
+								if store.gateway.id != viewStore.gateways.last?.gateway.id {
+									Separator()
+										.padding(.horizontal, .medium3)
+								} else {
+									Separator()
+								}
```
kugel3, line 31: I would do something like:
```
let padSeparator = store.gateway.id != viewStore.gateways.last?.gateway.id
Separator()
    .padding(.horizontal, padSeparator ? .medium3 : 0)
```
or even make `padSeparator` a computed property on the store

RadixWallet/Features/GatewaySettingsFeature/Components/GatewayList/GatewayList+View.swift
```diff
@@ -14,22 +12,32 @@ extension GatewayList {
...
+							}
+							.background(Color.app.white)
```
kugel3, line 33: You can probably get rid of `Color` here, even if you needed it while code completing.

RadixWallet/Features/GatewaySettingsFeature/Components/GatewayRow/GatewayRow+Reducer.swift
```diff
@@ -3,11 +3,14 @@ import SwiftUI
 
 // MARK: - GatewayRow
 public struct GatewayRow: Sendable, FeatureReducer {
+	@ObservableState
```
kugel3, line 6: Why make this Observable if we still use the old ViewStore paradigm? We should ideally migrate to observable btw.

RadixWallet/Features/GatewaySettingsFeature/Components/GatewayRow/GatewayRow+View.swift
```diff
@@ -4,11 +4,12 @@ import SwiftUI
 extension Gateway {
 	var displayName: String {
 		if isWellknown {
-			if network.id == .mainnet {
-				"Mainnet Gateway"
-			} else if network.id == .stokenet {
+			switch network.id {
+			case .mainnet:
+				"Radix Mainnet Gateway"
+			case .stokenet:
```
kugel3, line 10: Much better

RadixWallet/Features/GatewaySettingsFeature/Components/GatewayRow/GatewayRow+Reducer.swift
kugel3: Aha, it's a different reducer

RadixWallet/Features/GatewaySettingsFeature/GatewaySettings+Reducer.swift
```diff
@@ -38,13 +38,15 @@ public struct GatewaySettings: Sendable, FeatureReducer {
 	}
```
kugel3, line 39: Might as well go all the way and add the `@Reducer` macro, so we stay up to date with what TCA adds in the future. Note the things you need to make explicit for that to work. [See here](https://rdxworks.slack.com/archives/C03QFAWBRNX/p1722331252960899).

kugel3 (approved):

RadixWallet/Features/GatewaySettingsFeature/Children/AddNewGateway/AddNewGateway+Reducer.swift
```diff
@@ -102,7 +96,11 @@ public struct AddNewGateway: Sendable, FeatureReducer {
 			return handle(error, state: &state)
 
 		case .addGatewayResult(.success):
-			return .send(.delegate(.dismiss))
+			return .run { _ in
+				if isPresented {
```
matiasbzurovski, line 100: do we need the `isPresented` check?

RadixWallet/Features/GatewaySettingsFeature/Components/GatewayRow/GatewayRow+View.swift
```diff
@@ -46,33 +29,33 @@ extension GatewayRow {
...
+			WithPerceptionTracking {
 				Button(
 					action: {
-						viewStore.send(.didSelect)
+						store.send(.view(.didSelect))
 					}, label: {
-						HStack {
-							Image(asset: AssetResource.checkmarkBig)
+						HStack(spacing: .zero) {
+							Image(asset: AssetResource.check)
```
matiasbzurovski, line 38: We should be building a custom `HStack` here, but instead using our common `PlainListRow`. This way we can be sure that we have the same UI implementation everywhere we use these kind of rows. 

For this case, you probably want to use a `PlainListRowCore` with `title` and `detail` set, and have deletion button as the `Accesory` while the check image as the `Icon`.

matiasbzurovski (changes requested):

RadixWallet/Features/GatewaySettingsFeature/Components/GatewayRow/GatewayRow+View.swift
matiasbzurovski: shouldn't we localize these values?

RadixWallet/Features/GatewaySettingsFeature/Children/AddNewGateway/AddNewGateway+Reducer.swift
danvleju-rdx: Without this check, the `test_whenNewAddGatewayButtonIsTapped_thenDelegateIsCalled` test fails:
```
A reducer requested dismissal at "Radix_Wallet_Dev/AddNewGateway+Reducer.swift:110", but couldn't be dismissed. …

This is generally considered an application logic error, and can happen when a reducer assumes it runs in a presentation context. If a reducer can run at both the root level of an application, as well as in a presentation destination, use @Dependency(\.isPresented) to determine if the reducer is being presented before calling @Dependency(\.dismiss).
```

matiasbzurovski (approved):

PR #1262, opened 2024-08-02
[ABW-3458] NFT page request updates

matiasbzurovski:
Jira ticket: [ABW-3458](https://radixdlt.atlassian.net/browse/ABW-3458)

## Description
This PR updates paging mechanism for NFTs. Our current logic was based on the assumptions that the GW would return pages with the max amount of items, and this could change in the future. In other words, if there are 100 elements to load and we were requesting pages of 29 elements, the app assumed the GW would return 4 pages with `[29, 29, 29, 13]` elements each. However, GW will change and the amount of pages/elements per page won't be fixed.

## How to test
Play with different NFT collections and sizes to verify the loading happens correctly every time.

[ABW-3458]: https://radixdlt.atlassian.net/browse/ABW-3458?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/RadixWallet/Clients/OnLedgerEntitiesClient/OnLedgerEntitiesClient+Live.swift
+++ b/RadixWallet/Clients/OnLedgerEntitiesClient/OnLedgerEntitiesClient+Live.swift
@@ -2,7 +2,7 @@ import Sargon
 
 // MARK: - OnLedgerEntitiesClient + DependencyKey
 extension OnLedgerEntitiesClient: DependencyKey {
-	public static let maximumNFTIDChunkSize = 29
+	public static let maximumNFTIDChunkSize = 100
 
 	enum Error: Swift.Error {
 		case emptyResponse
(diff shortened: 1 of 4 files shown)


RadixWallet/Clients/OnLedgerEntitiesClient/OnLedgerEntitiesClient+Live.swift
```diff
@@ -2,7 +2,7 @@ import Sargon
 
 // MARK: - OnLedgerEntitiesClient + DependencyKey
 extension OnLedgerEntitiesClient: DependencyKey {
-	public static let maximumNFTIDChunkSize = 29
+	public static let maximumNFTIDChunkSize = 100
```
danvleju-rdx, line 5: So you are saying that GW might not return pages with the max number of items, even if there are more pages available? Didn't quite understand why this value should change.

matiasbzurovski: There are two things here:
- This number changes to 100 because such is the value that each page can return now (and not 29 as it previously did).
- The GW will change to potentially return less items than what it fit on one page, even if there are items to return. For example, if there are 250 items, it may return `[100, 80, 70]` and not `[100, 100, 50]` (which is what the app, without my change, currently expects and why it wouldn't work if different amount of items were returned).

danvleju-rdx: Got it, thanks for the clarification

danvleju-rdx (approved):

GhenadieVP (approved):

PR #1264, opened 2024-08-05
[ABW-3062] Display nil fiat values without fractional digits

danvleju-rdx:
Jira ticket: [ABW-3062](https://radixdlt.atlassian.net/browse/ABW-3062)

## Description
Display nil fiat values without fractional digits

## Screenshot

| - | - |
| Simulator Screenshot - iPhone 15 Pro Max - 2024-08-05 at 10 59 39| Simulator Screenshot - iPhone 15 Pro Max - 2024-08-05 at 10 57 59 |

[ABW-3062]: https://radixdlt.atlassian.net/browse/ABW-3062?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/RadixWallet/Core/SharedModels/Assets/FiatWorth.swift
+++ b/RadixWallet/Core/SharedModels/Assets/FiatWorth.swift
@@ -99,16 +99,18 @@ extension FiatWorth {
 	private static let hiddenValue = "• • • •"
 	private static let unknownValue = "—"
 
-	func currencyFormatted(applyCustomFont: Bool = false) -> AttributedString {
+	private static let currencyFormatter: NumberFormatter = {
 		let formatter = NumberFormatter()
 		formatter.numberStyle = .currency
-		formatter.currencyCode = currency.currencyCode
+		return formatter
+	}()
 
+	func currencyFormatted(applyCustomFont: Bool = false) -> AttributedString {
 		let value = worth.value ?? .zero // Zero for the unknown case, just to do to the base formatting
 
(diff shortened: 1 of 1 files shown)


CyonAlexRDX (approved):
LGTM

matiasbzurovski (commented):
I wasn't aware that we were creating a new instance of `NumberFormatter` every time we want to format a FiatWort. I understand that, as any formatter, its creation could be quite expensive. Maybe it is a good time to have a shared formatter, and just set the `maximumFractionDigits` to 0/2 depending on the value being `.zero`?

kugel3 (commented):
Isn't this something that should be handled in Sargon, since we do the other number formatting there? As mentioned, if we do use a native formatter it should not be recreated all the time, for some reason that is quite expensive.

In this case though, you should be able to use the `myNumber.formatted(...)` method instead, no?

danvleju-rdx:
Agree, we should avoid recreating the formatter each time. Using `myNumber.formatted(…)` is not ideal, as we need below the formatter for currencySymbol and decimalSeparator.
Perhaps we should have this logic in Sargon, but I view it as a separate migration task. The Android team will also need to make adjustments on their side.

GhenadieVP (approved):

matiasbzurovski (approved):

PR #1272, opened 2024-08-07
[ABW-2700] Show divisibility on details view

matiasbzurovski:
Jira ticket: [ABW-2700](https://radixdlt.atlassian.net/browse/ABW-2700)

## Description
Shows the divisibility on fungible token details.

## Screenshots
  

[ABW-2700]: https://radixdlt.atlassian.net/browse/ABW-2700?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/RadixWallet/Features/AssetsFeature/Components/HelperViews/AssetResourceDetails.swift
+++ b/RadixWallet/Features/AssetsFeature/Components/HelperViews/AssetResourceDetails.swift
@@ -62,7 +62,7 @@ struct AssetResourceDetailsSection: View {
 				if let divisibility = viewState.divisibility {
 					loadable(divisibility) { value in
 						if let value {
-							KeyValueView(key: "Divisibility", value: "\(value)")
+							KeyValueView(key: L10n.AssetDetails.divisibility, value: "\(value)")
 						}
 					}
 				}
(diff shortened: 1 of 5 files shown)


RadixWallet/Features/AssetsFeature/Components/HelperViews/AssetResourceDetails.swift
```diff
@@ -13,6 +13,7 @@ struct AssetResourceDetailsSection: View {
 		let validatorAddress: ValidatorAddress?
 		let resourceName: Loadable<String?>?
 		let currentSupply: Loadable<String?>
+		let divisibility: Loadable<UInt8?>?
```
kugel3, line 16: Using optional Loadables seems a bit much, no? IF the inner optional is nil, that means that e.g. divisibility is not present, but what does it mean if the entire property is nil? That divisibility is not relevant for this resource? If so, that makes some sense, but it makes less sense for resourceName.

matiasbzurovski: yes, for example NFTs would have the `divisibility` set to nil. However, if the `Lodable` is present, we still have to make its inner value optional since it may not be present for a given fungible resource.

giannis-rdx: >, if the Lodable is present, we still have to make its inner value optional since it may not be present for a given fungible resource.

@matiasbzurovski if the value is not present what do you show in the resource details view? 0? or just empty?

matiasbzurovski: empty

PR #1274, opened 2024-08-08
[ABW-3601] Privacy protection overlay

danvleju-rdx:
Jira ticket: [ABW-3601](https://radixdlt.atlassian.net/browse/ABW-3601)

## Description
Adds a security overlay that hides the content of the app when it is moved to background / App Switcher.

## Video

https://github.com/user-attachments/assets/3a04f7fb-4ccc-4b1b-8bcc-68377678b7d1

https://github.com/user-attachments/assets/52f0a370-b8dd-48f6-897f-584b2a4fe4b0

[ABW-3601]: https://radixdlt.atlassian.net/browse/ABW-3601?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


CyonAlexRDX (approved):
LGTM.

Are there any corner case we might need to handle? Any important gotchas? 

Perhaps not..? But eg trying Lock Screen. Phone calls. Timers. Deep links. In different states, force quitted or suspended.

danvleju-rdx:
@CyonAlexRDX I've tested all the scenarios you mentioned, except for phone calls, and it works as expected. `sceneWillResignActive` and `sceneDidBecomeActive` seem to do their job.

RadixWallet/Features/SplashFeature/Splash+View.swift
```diff
@@ -101,3 +77,36 @@ extension Splash.State {
 	public static let previewValue = Self()
 }
 #endif
+
+// MARK: - SplashView
+public struct SplashView: View {
+	var biometricsCheckFailed: Bool = false
+
+	public var body: some View {
+		ForceFullScreen {
```
kugel3, line 86: ForceFullScreen is an old hack that should not be needed or used. In fact it would be nice to get rid of it altogether, it's only used here an in one other spot (where it is even more dodgy). Check if it's not enough to do something like `.frame(maxHeight: .infinity)` instead. And for the other use site, maybe just replacing it with `Color.clear` will be enough.

RadixWallet/Features/SplashFeature/Splash+View.swift
```diff
@@ -101,3 +77,36 @@ extension Splash.State {
...
+						.frame(.small)
+					Text(L10n.Splash.tapAnywhereToUnlock)
+						.textStyle(.body1HighImportance)
+				}
+			}
+			.padding(.bottom, .medium1)
+			.foregroundColor(.app.white)
+			.frame(
+				minWidth: 0,
+				maxWidth: .infinity,
```
kugel3, line 101: Alright, we already have `maxHeight`, is that not enough? And does `minWidth` actually do anything here?

RadixWallet/Features/SplashFeature/Splash+View.swift
danvleju-rdx: removed ForceFullScreen

RadixWallet/Features/SplashFeature/Splash+View.swift
danvleju-rdx: Only `minWidth ` isn't enough as the image has portrait ratio. Setting the `maxWidth` to `.infinity` will do the trick, but it would be safer to have it:
```
.frame(maxWidth: .infinity, maxHeight: .infinity)
```

And yes, `minWidth` and `minHeight` don't make sense here.

kugel3: Sorry, was a bit unclear. What I meant was if maxWidth + maxHeight was enough, that they obviated the need for ForceFullScreen. Which seems to be the case, so that's good!

kugel3 (approved):
Good riddance of ForceFullScreen, if you also checked that Color.clear was enough we should be good to go.

PR #1286, opened 2024-08-14
[ABW-3692] Option to require biometrics check

danvleju-rdx:
Jira ticket: [ABW-3692](https://radixdlt.atlassian.net/browse/ABW-3692)

## Description
- Adds an option to require a biometric/PIN check when the app starts or returns to the foreground.
- Adds a toggle in App Preferences to determine whether a biometric/PIN check is required (only if iOS 18 is not available).
- Handles iOS versions as follows:
   - **iOS 18 available**: Users will lock the app at the system level. In this case, we agreed to disable the in-app biometric check feature.
   - **iOS 18 not available**: Users will use the in-app biometric check feature.

## Videos
<details>
  <summary>iOS 18 not available</summary>

https://github.com/user-attachments/assets/eb12938d-ef36-48c1-8f4f-3b1bd92ba9be 

</details>

<details>
  <summary>iOS 18 available</summary>

https://github.com/user-attachments/assets/9d63102d-c352-4952-bb01-ba1d42f0a127

</details>

[ABW-3692]: https://radixdlt.atlassian.net/browse/ABW-3692?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


RadixWallet/Core/DesignSystem/Components/PlainListRow.swift
```diff
@@ -192,9 +192,7 @@ struct PlainListRowCore: View {
 private extension PlainListRowCore.ViewState {
 	var titleTextStyle: TextStyle {
 		switch context {
-		case .toggle:
-			.secondaryHeader
-		case .settings, .dappAndPersona:
+		case .toggle, .settings, .dappAndPersona:
 			.body1Header
```
danvleju-rdx, line 196: The title font for the `toggle` context didn't match the design.

RadixWallet/Core/DesignSystem/ToggleView.swift
```diff
@@ -30,10 +33,27 @@ public struct ToggleView: SwiftUI.View {
...
+		case toggle
+	}
+}
+
+extension ToggleView.Context {
+	var plainListRowContext: PlainListRowCore.ViewState.Context {
+		switch self {
+		case .settings: .settings
+		case .toggle: .toggle
+		}
```
matiasbzurovski, line 57: why do we need this ?

`PlainListRowCore.ViewState.Context` sent to `PlainListRow` should always be `.toggle`.

CyonAlexRDX (approved):
Nice, lgtm!

RadixWallet/Features/SplashFeature/Splash.swift
```diff
@@ -23,40 +32,43 @@ public struct Splash: Sendable, FeatureReducer {
...
+		@CasePathable
 		public enum State: Sendable, Hashable {
 			case passcodeCheckFailed(AlertState<Action.PasscodeCheckFailedAlert>)
+			case errorAlert(AlertState<Action.ErrorAlert>)
 		}
 
+		@CasePathable
 		public enum Action: Sendable, Equatable {
 			case passcodeCheckFailed(PasscodeCheckFailedAlert)
+			case errorAlert(ErrorAlert)
```
matiasbzurovski, line 54: I find it a bit confusing that we have:
- `state.biometricsCheckFailed`: boolean used to show something or not on view
- `state.destination.passcodeCheckFailed`: a destination alert
- `state.destination.errorAlert`: a destination alert

We can probably merge the last two in one single destination (`errorAlert`), that depending on the situation that triggers it we populate `AlertState` with different title, message and buttons.

matiasbzurovski (commented):
Code is looking good, will do some testing with different conifgurations later

RadixWallet/Core/DesignSystem/ToggleView.swift
danvleju-rdx: >PlainListRowCore.ViewState.Context sent to PlainListRow should always be .toggle.

For toggle views in Preferences, we need to use `.settings` since they have a different design in that context.


I explicitly defined `ToggleView.Context` (instead of directly passing `PlainListRowCore.ViewState.Context`) to make it clear which contexts are supported by `ToggleView`. This prevents the mistaken use of unsupported contexts, such as `.dappAndPersona`, under the assumption that they are valid states and would be handled.

RadixWallet/Features/SplashFeature/Splash.swift
danvleju-rdx: I thought it would be clearer to keep them separate, but yes, I agree we can merge them.

RadixWallet/Core/DesignSystem/ToggleView.swift
matiasbzurovski: oh I see now, thanks for explanation

matiasbzurovski (approved):

RadixWallet/Features/AppFeature/SceneDelegate.swift
```diff
@@ -58,6 +72,35 @@ public final class SceneDelegate: NSObject, UIWindowSceneDelegate, ObservableObj
...
+		biometricsSplashWindow?.rootViewController = UIHostingController(rootView: splashView)
+		biometricsSplashWindow?.windowLevel = .normal + 2
+		biometricsSplashWindow?.makeKeyAndVisible()
+
+		@Dependency(\.localAuthenticationClient) var localAuthenticationClient
+		Task { @MainActor in
+			for try await _ in localAuthenticationClient.authenticatedSuccessfully() {
+				hideBiometricsSplashWindow()
+			}
+		}
```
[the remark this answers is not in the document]
danvleju-rdx: Right. Moved in `willConnectTo session` method

RadixWallet/Features/AppFeature/SceneDelegate.swift
```diff
@@ -58,6 +79,28 @@ public final class SceneDelegate: NSObject, UIWindowSceneDelegate, ObservableObj
 		self.overlayWindow = overlayWindow
 	}
 
+	// MARK: Biometrics
+
+	private var biometricsSplashWindow: UIWindow?
+
+	private func showBiometricsSplashWindow(in scene: UIWindowScene) {
+		let splashView = Splash.View(
```
GhenadieVP, line 90: Why not reuse the `privacyProtectionWindow`?

RadixWallet/Features/SettingsFeature/Preferences/Preferences.swift
```diff
@@ -104,6 +105,13 @@ public struct Preferences: Sendable, FeatureReducer {
 				try await appPreferencesClient.updatePreferences(preferences)
 			}
 
+		case let .advancedLockToogled(isEnabled):
+			state.appPreferences?.security.isAdvancedLockEnabled = isEnabled
```
GhenadieVP, line 109: hmm, not sure what how it is done in other apps - should ask users to authenticate at least when they are disabling the Advanced Lock?

RadixWallet/Features/SettingsFeature/Preferences/Preferences.swift
```diff
@@ -104,6 +105,13 @@ public struct Preferences: Sendable, FeatureReducer {
...
+			guard let preferences = state.appPreferences else { return .none }
+			return .run { _ in
+				try await appPreferencesClient.updatePreferences(preferences)
+			}
```
GhenadieVP, line 113: Should probably handle the error. True that this can actually fail when the Profile failed to be saved, which would be a bigger issue, but still good practice to catch the error.

RadixWallet/Features/SplashFeature/Splash.swift
```diff
@@ -72,7 +80,21 @@ public struct Splash: Sendable, FeatureReducer {
...
+					#if targetEnvironment(simulator)
+					let action: InternalAction = _XCTIsTesting ? .checkBiometrics : .loadingDataCompleted
+					#else
+					let action: InternalAction = .checkBiometrics
+					#endif
+					await send(.internal(action))
+				} else {
+					await send(.internal(.loadingDataCompleted))
+				}
+			}
```
GhenadieVP, line 97: I would suggest to introduce a single internal action, such as `advancedLockStateLoaded(isEnabled: Bool)`, which sent from this run block.
Then you could remove the other two actions `checkBiometrics` and `loadingDataCompleted`, in favour of having:
```swift
case .let advancedLockStateLoaded(isEnabled):
   return isEnabled ? verifyPasscode() : delegateCompleted()
```

RadixWallet/Features/AppFeature/SceneDelegate.swift
danvleju-rdx: I see the `privacyProtectionWindow` as a completely separate layer. It has a different windowLevel and context, and its purpose is to cover all other windows when the app is in App Switcher mode, including the biometricsSplashWindow if it's displayed at that time.

RadixWallet/Features/SettingsFeature/Preferences/Preferences.swift
danvleju-rdx: Indeed, it will be an improvement. I believe it's sufficient to request biometrics only when users are disabling Advanced Lock, as there's no practical reason to require biometrics when enabling it. 
Here's a video showing the result:

https://github.com/user-attachments/assets/0d626bb3-6f20-4102-a82c-4c7baf937fb2

GhenadieVP (approved):

PR #1290, opened 2024-08-20
[ABW-3723] Resource hiding/unhiding

matiasbzurovski:
Jira ticket: [ABW-3723](https://radixdlt.atlassian.net/browse/ABW-3723)

## Description
This PR allows to hide/unhide resources (fungible, non-fungible & pool units)

## How to test
Play with your user as the demo video shows.

## Video
[Video](https://drive.google.com/file/d/1YbCZZpr64Nggu_mHrNbXtZArjyyBhnNw/view?usp=sharing)

## PR submission checklist

[ABW-3723]: https://radixdlt.atlassian.net/browse/ABW-3723?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


RadixWallet/Features/AssetsFeature/Components/NonFungibleAssetList/Components/Details/NonFungibleTokenDetails+View.swift
```diff
@@ -122,6 +122,11 @@ extension NonFungibleTokenDetails {
 						}
 						.padding(.vertical, .medium1)
 						.background(.app.gray5, ignoresSafeAreaEdges: .bottom)
+
+						IfLetStore(store.scope(state: \.hideAsset, action: \.child.hideAsset)) { store in
```
CyonAlexRDX, line 126: hmm I thought one did not use `IfLetStore` from TCA anymore? [See migration guide 1.7](https://pointfreeco.github.io/swift-composable-architecture/main/documentation/composablearchitecture/migratingto1.7/#Replacing-IfLetStore-with-if-let)

Cant we:
```swift
if let childStore = store.scope(state: \.hideAsset, action: \.child.hideAsset) {
    HideAsset.View(store: store).padding(...)
}
```
?

RadixWallet/Features/AssetsFeature/Components/NonFungibleAssetList/Components/Row/NonFungbileAssetRow+Reducer.swift
```diff
@@ -146,3 +149,13 @@ extension NonFungibleAssetList {
 		}
 	}
 }
+
+private extension NonFungibleAssetList.Row.InternalAction.TokensLoadResult {
+	init(tokens: [OnLedgerEntity.NonFungibleToken], hiddenAssets: [AssetAddress], nextPageCursor: String?, previousTokenIndex: Int) {
```
CyonAlexRDX, line 154: split lines please (is it CTRL + M ?) :)

RadixWallet/Features/SettingsFeature/Preferences/HiddenAssets/HiddenAssets.swift
```diff
@@ -0,0 +1,225 @@
...
+			}
+			.compactMap { $0 }
+			await send(.internal(.setPoolUnit(poolUnitDetails)))
+		}
+	}
+}
+
+// MARK: - Helpers
+
+private extension [AssetAddress] {
```
CyonAlexRDX, line 189: personal preferences, I dislike `extension [Element]` syntax - since it makes it impossible to search for all Array extensions and all Dictionary extensions individually. But I think I'm the only one to dislike it and we are using it elsewhere.

CyonAlexRDX (approved):
Excellent!

RadixWallet/Features/SettingsFeature/Preferences/HiddenAssets/HiddenAssets+View.swift
```diff
@@ -0,0 +1,139 @@
...
+
+		private func header(_ value: String) -> some SwiftUI.View {
+			Text(value)
+				.textStyle(.secondaryHeader)
+				.foregroundColor(.app.gray2)
+		}
+
+		@ViewBuilder
+		private var fungibles: some SwiftUI.View {
+			if store.fungible.isEmpty {
```
CyonAlexRDX, line 44: annoying that ResultBuilders does not support guard or otherwise return early :/ 
[SE-0289 states](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0289-result-builders.md)

> `guard` is provisionally ill-formed when it appears within a transformed function. Situations in which this statement may appear may be supported in the future, such as when the statement does not produce a partial result.

RadixWallet/Features/AssetsFeature/Components/NonFungibleAssetList/Components/Details/NonFungibleTokenDetails+View.swift
matiasbzurovski: mm I don't think so, the [method](https://pointfreeco.github.io/swift-composable-architecture/main/documentation/composablearchitecture/store/scope(state:action:)-90255/) referred to in that guide returns a non-optional `Store<ChildState, ChildAction>`. 

Therefore, in this case it returns:
```Swift
Store<HideAsset.State?, HideAsset.Action>
```

while what we would need is:
```Swift
Store<HideAsset.State, HideAsset.Action>?
```

RadixWallet/Features/AssetsFeature/Components/NonFungibleAssetList/Components/Row/NonFungbileAssetRow+Reducer.swift
matiasbzurovski: TIL about that shortcut, thanks!

RadixWallet/Features/SettingsFeature/Preferences/HiddenAssets/HiddenAssets+View.swift
matiasbzurovski: right, or ternary operators so that we could do something like this on the `body`:
```Swift
store.fungibles.isEmpty ? empty : fungibles
```

and remove the conditional check inside `var fungibles`

RadixWallet/Features/SettingsFeature/Preferences/HiddenAssets/HiddenAssets.swift
matiasbzurovski: I actually used to add extensions like this as `extension Array where Element == XYZ` before, would you prefer it that way?

RadixWallet/Features/AssetsFeature/Components/NonFungibleAssetList/Components/Details/NonFungibleTokenDetails+View.swift
matiasbzurovski: We would be able to use this [method](https://pointfreeco.github.io/swift-composable-architecture/main/documentation/composablearchitecture/store/scope(state:action:fileid:filepath:line:column:)-3yvuf) if we had this reducer updated to use `ObservableState`

CyonAlexRDX: Right, ObservableState required

RadixWallet/Features/SettingsFeature/Preferences/HiddenAssets/HiddenAssets.swift
CyonAlexRDX: Yes I prefer that, but then we have fragmentation in the code base. I think this is the most common syntax we use, so maybe best stick to it

RadixWallet/Clients/AccountPortfoliosClient/AccountPortfoliosClient+Live.swift
```diff
@@ -25,12 +25,21 @@ extension AccountPortfoliosClient: DependencyKey {
 			}
 		}
 
+		/// Update when hidden assets change
+		Task {
+			for try await _ in await appPreferencesClient.appPreferenceUpdates().map(\.assets.hiddenAssets) {
```
kugel3, line 30: Doesn't this actually get called every time _anything_ changes in appPreferences? Not a big deal probably, but you might as well throw on a `removeDuplicates()`.

RadixWallet/Clients/AccountPortfoliosClient/AccountPortfoliosClient+State.swift
```diff
@@ -12,6 +13,43 @@ extension AccountPortfoliosClient {
...
+			modified.fungibleResources.nonXrdResources.removeAll(where: { resource in
+				hiddenAssets.contains(.fungible(resource.resourceAddress))
+			})
+
+			// Remove every hidden pool unit
+			modified.poolUnitResources.poolUnits.removeAll(where: { poolUnit in
+				hiddenAssets.contains(.poolUnit(poolUnit.resourcePoolAddress))
+			})
+
+			// We cannot remove the NFTs, since at this moment we haven't loaded them yet.
```
kugel3, line 30: In what sense are they not loaded yet? Also, this whole section confuses me, because I thought we only wanted to hide entire (non-fungible) resources, but perhaps this is older code that hasn't been removed yet?

RadixWallet/Clients/AppPreferencesClient/AppPreferencesClient+Interface.swift
```diff
@@ -70,6 +70,14 @@ extension AppPreferencesClient {
 			display.isCurrencyAmountVisible.toggle()
 		}
 	}
+
+	public func getHiddenAssets() async -> [AssetAddress] {
+		await getPreferences().assets.hiddenAssets
+	}
+
+	public func isAssetHidden(asset: AssetAddress) async -> Bool {
```
kugel3, line 78: Minor style question: I don't think you need to repeat "asset" in both the function name and the argument, especially when the type is also called Asset

RadixWallet/Clients/EntitiesVisibilityClient/EntitiesVisibilityClient+Interface.swift
```diff
@@ -1,46 +1,32 @@
...
+	public typealias UnhideAccount = @Sendable (Account.ID) async throws -> Void
+	public typealias UnhidePersona = @Sendable (Persona.ID) async throws -> Void
+	public typealias GetHiddenEntities = @Sendable () async throws -> HiddenEntities
 }
 
 extension EntitiesVisibilityClient {
-	public func hideAccounts(ids: some Collection<Account.ID>) async throws {
-		try await hideAccounts(Set(ids))
+	func hide(account: Account) async throws {
+		try await hideAccount(account.id)
```
kugel3, line 26: Why do we need both `hide(account: Account)` and `hideAccount`? Isn't the latter enough? Come to that, isn't it enough to have `hideEntity`?

RadixWallet/Clients/AccountPortfoliosClient/AccountPortfoliosClient+State.swift
GhenadieVP: correct, we should hide the entire non fungible resource, not the specific non fungible ids.

RadixWallet/Features/AssetsFeature/Components/HelperViews/ResourceBalance/ResourceBalanceButton.swift
```diff
@@ -5,30 +5,42 @@ public struct ResourceBalanceButton: View {
...
-					.contentShape(Rectangle())
-					.background(background)
+
+				if let warning {
+					WarningErrorView(text: warning, type: .warning, useNarrowSpacing: true)
+				}
 			}
+			.padding(.top, topPadding)
+			.padding(.bottom, bottomPadding)
+			.padding(.horizontal, horizontalSpacing)
```
kugel3, line 41: Minor: we usually put horizontal between top and bottm

RadixWallet/Features/AssetsFeature/Components/HideAsset/HideAsset.swift
```diff
@@ -0,0 +1,120 @@
...
+		case task
+		case buttonTapped
+	}
+
+	public enum DelegateAction: Sendable, Equatable {
+		case didHideAsset
+	}
+
+	// MARK: - Destination
+	public struct Destination: DestinationReducer {
```
kugel3, line 33: Did you try using the new style, where you just define something like:
```
extension ContactsFeature {
  @Reducer(state: .hashable, action: .equatable)
  enum Destination {
    case confirmation(ConfirmationAction)
  }
}
```

RadixWallet/Features/AssetsFeature/Components/HideAsset/HideAsset.swift
```diff
@@ -0,0 +1,120 @@
+// MARK: - HideAsset
+@Reducer
```
kugel3, line 2: If the macro is causing issues when combined with FeatureReducer, feel free to skip the latter, and implement the needed actions etc manually.

RadixWallet/Features/AssetsFeature/Components/HideAsset/HideAsset.swift
```diff
@@ -0,0 +1,120 @@
...
+			return hideAssetEffect(state: state)
+		case .confirmation(.cancel):
+			state.destination = nil
+			return .none
+		}
+	}
+}
+
+private extension HideAsset {
+	func shouldShowEffect(state: State) -> Effect<Action> {
```
kugel3, line 93: Good!

RadixWallet/Features/AssetsFeature/Components/NonFungibleAssetList/Components/Details/NonFungibleTokenDetails+Reducer.swift
```diff
@@ -11,6 +11,7 @@ public struct NonFungibleTokenDetails: Sendable, FeatureReducer {
 		public let ledgerState: AtLedgerState
 		public let stakeClaim: OnLedgerEntitiesClient.StakeClaim?
 		public let isClaimStakeEnabled: Bool
+		var hideAsset: HideAsset.State?
```
kugel3, line 14: I suppose this should also be public?

RadixWallet/Features/SettingsFeature/Preferences/HiddenAssets/HiddenAssets+View.swift
kugel3: You could put the if-else at the parent level, and let the fungibles var only be called when not empty. That would be pretty clean. I think it should be called `emptyState` though, not just `empty`. And I would make it a nested view, so you would use it with `EmptyState()` instead, it doesn't seem to use any variables.

RadixWallet/Features/SettingsFeature/Preferences/HiddenEntities/HiddenEntities+View.swift
```diff
@@ -0,0 +1,99 @@
...
+						}
+					}
+				}
+			}
+		}
+
+		@ViewBuilder
+		private var accounts: some SwiftUI.View {
+			if store.accounts.isEmpty {
+				empty
```
kugel3, line 61: Same here, since the empty state is not specific to accounts, I don't think it needs to be part of `var accounts`

kugel3 (approved):

RadixWallet/Features/AssetsFeature/Components/HideAsset/HideAsset.swift
```diff
@@ -0,0 +1,120 @@
...
+			case let .fungible(resource):
+				let networkId = await gatewaysClient.getCurrentNetworkID()
+				isXrd = resource.isXRD(on: networkId)
+			case .nonFungible, .poolUnit:
+				isXrd = false
+			}
+			if isXrd {
+				await send(.internal(.setShouldShow(false)))
+			} else {
+				let isAlreadyHidden = await appPreferencesClient.isAssetHidden(asset: state.asset)
```
GhenadieVP, line 106: I suppose you added this for the scenario when viewing the details of an asset from TX Review, but should we allows users to hide an asset from TX Review in the first place?

RadixWallet/Clients/AccountPortfoliosClient/AccountPortfoliosClient+State.swift
matiasbzurovski: Yes, these are the changes that I am working on since the change on the designs after Friday's demo

RadixWallet/Clients/EntitiesVisibilityClient/EntitiesVisibilityClient+Interface.swift
matiasbzurovski: `hideEntity` doesn't work since one uses `Account.ID` and the other `Persona.ID`. The second method is just an inherited helper method used to call from the Account object instead than from its id. Can definitely be removed.

RadixWallet/Features/AssetsFeature/Components/HideAsset/HideAsset+View.swift
```diff
@@ -0,0 +1,51 @@
...
+
+@MainActor
+private extension View {
+	func destination(store: StoreOf<HideAsset>) -> some View {
+		let destinationStore = store.destination
+		return confirmation(with: destinationStore, store: store)
+	}
+
+	private func confirmation(with destinationStore: PresentationStoreOf<HideAsset.Destination>, store: StoreOf<HideAsset>) -> some View {
+		sheet(store: destinationStore.scope(state: \.confirmation, action: \.confirmation)) { _ in
```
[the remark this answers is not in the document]
matiasbzurovski: I scope the `destinationStore` to tell when this `ConfirmationView` alert should be presented. Given the alert doesn't have a reducer (it is a plain SwiftUI view that have a closure for each button that can be tapped: confirm and cancel), we just need to handle the button tap on this reducer.

RadixWallet/Features/AssetsFeature/Components/NonFungibleAssetList/Components/Details/NonFungibleTokenDetails+Reducer.swift
matiasbzurovski: nothing needs to be public actually :)

kugel3: Sure, but since the struct and all the other properties are public, this one should be as well. There is no reason for them to have different access levels, it only causes confusion.

RadixWallet/Clients/AccountPortfoliosClient/AccountPortfoliosClient+Live.swift
```diff
@@ -25,12 +25,21 @@ extension AccountPortfoliosClient: DependencyKey {
...
+				guard !Task.isCancelled else { return }
+				let accountAddresses = state.portfoliosSubject.value.wrappedValue.map { $0.map(\.key) } ?? []
+				_ = try await fetchAccountPortfolios(accountAddresses, forceRefreshEntities: false, forceRefreshPrices: true)
+			}
+		}
+
 		/// Fetches the pool and stake units details for a given account; Will update the portfolio accordingly
 		@Sendable
-		func fetchPoolAndStakeUnitsDetails(_ account: OnLedgerEntity.OnLedgerAccount, cachingStrategy: OnLedgerEntitiesClient.CachingStrategy) async {
+		func fetchPoolAndStakeUnitsDetails(_ account: OnLedgerEntity.OnLedgerAccount, hiddenAssets: [AssetAddress], cachingStrategy: OnLedgerEntitiesClient.CachingStrategy) async {
```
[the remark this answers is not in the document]
matiasbzurovski: I thought about it, but I don't see any benefit in performance (and would have the filter logic in two places)

RadixWallet/Clients/OnLedgerEntitiesClient/Helpers/OnLedgerEntitiesClient+ComplexTypes.swift
```diff
@@ -64,11 +68,13 @@ extension OnLedgerEntitiesClient {
 				guarantee: guarantee
 			)
 		}
+
 		// Normal fungible resource
 		let isXRD = resourceAddress.isXRD(on: networkID)
+		let isHidden = hiddenAssets.contains(.fungible(resourceAddress))
```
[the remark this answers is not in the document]
matiasbzurovski: We already need to await for getting the `hiddenAssets` in this method, since they need to be sent to the `func poolUnit()`. Then, using the helper `func isAssetHidden()` here means adding an extra await when there is no need.

RadixWallet/Clients/OnLedgerEntitiesClient/Helpers/OnLedgerEntitiesClient+ComplexTypes.swift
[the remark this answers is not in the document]
matiasbzurovski: Yes, I did try 3 different solutions around this because none of them were ideal, and using the `ResourceBalance` ended up being the cleanest one (since it is where all the information is loaded).

RadixWallet/Features/AssetsFeature/Components/HideAsset/HideAsset.swift
matiasbzurovski: according to my demo on Friday, yes

RadixWallet/Features/AssetsFeature/Components/NonFungibleAssetList/Components/Details/NonFungibleTokenDetails+Reducer.swift
matiasbzurovski: IMO the confusion is caused by leaving the others public, and what we should do is actually remove the access modifier from thise.

RadixWallet/Clients/AccountPortfoliosClient/AccountPortfoliosClient+Live.swift
matiasbzurovski: great catch!

RadixWallet/Features/SettingsFeature/Preferences/HiddenAssets/HiddenAssets+View.swift
matiasbzurovski: well this is debatable. I prefer looking at the body and getting a clear and quick vision of what's going on. In my opinion it looks cleaner to see this:
```Swift
LazyVStack(alignment: .leading, spacing: .large3) {
	Text(L10n.HiddenAssets.text)
		.textStyle(.body1HighImportance)
		.foregroundColor(.app.gray2)

	header(L10n.HiddenAssets.fungibles)
	fungibles

	header(L10n.HiddenAssets.nonFungibles)
	nonFungibles

	header(L10n.HiddenAssets.poolUnits)
	poolUnits
}
```

than this:
```Swift
LazyVStack(alignment: .leading, spacing: .large3) {
	Text(L10n.HiddenAssets.text)
		.textStyle(.body1HighImportance)
		.foregroundColor(.app.gray2)

	header(L10n.HiddenAssets.fungibles)
	if store.fungibles.isEmpty {
		empty
	} else {
		fungibles
	}

	header(L10n.HiddenAssets.nonFungibles)
	if store.nonFungibles.isEmpty {
		empty
	} else {
		nonFungibles
	}

	header(L10n.HiddenAssets.poolUnits)
	if store.poolUnits.isEmpty {
		empty
	} else {
		poolUnits
	}
}
```

RadixWallet/Features/AssetsFeature/Components/HideAsset/HideAsset.swift
matiasbzurovski: how do you make the parent conform to `FeatureReducer` ?

RadixWallet/Features/SettingsFeature/Preferences/HiddenAssets/HiddenAssets+View.swift
kugel3: Sure, but the second option makes the variables cleaner. But my main point is that it is better to not hide the empty state, since that is in fact a top-level thing, I don't think it should be hidden.

RadixWallet/Clients/EntitiesVisibilityClient/EntitiesVisibilityClient+Interface.swift
kugel3: Hm, I thought we still had an Entity address that could be cast to either of those two, perhaps not.

RadixWallet/Features/AssetsFeature/Components/NonFungibleAssetList/Components/Details/NonFungibleTokenDetails+Reducer.swift
kugel3: Previously we had separate modules for every feature, so everything had to be public. Now there is no technical reason for it, but we should at the very least be consistent within a feature.

RadixWallet/Features/AssetsFeature/Components/HideAsset/HideAsset.swift
kugel3: You can do put `@Reducer(state: .hashable, action: .equatable)` on the Destination if that is what you have in mind? But it was a genuine question, it seems that it only will play nicely with FeatureReducer sometimes.

GhenadieVP (approved):

PR #1293, opened 2024-08-22
ABW-1671 Info Buttons and Sheet

kugel3:
Jira ticket: [ABW-1671](https://radixdlt.atlassian.net/browse/ABW-1671)

## Description
Introduces the long awaited info button, that pops a bottom sheet with a short description of the item in question. Tapping links to other items from within the sheet replaces the text (and scrolls to the top), instead of queueing up another sheet or dismissing and showing the new one (see clip).

## How to test
- Verify that all info buttons open and close as expected
- Verify that links work as expected
- Verify the places in Zeplin that have an info button, and that it leads to the desired info, especially the ones in [this list](https://docs.google.com/spreadsheets/d/1TLMhMdiLLACklRgoqrAUNcr2Yra8uRRratJV4xfrPuw/edit?gid=0#gid=0)

## Video

https://github.com/user-attachments/assets/9e31eef0-d936-42c3-a346-a499f6707ee1

## Note

I also added an infolink for Behaviors (though I haven't added the string yet, in case we don't actually want this:

## PR submission checklist

[ABW-1671]: https://radixdlt.atlassian.net/browse/ABW-1671?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


RadixWallet/Features/AssetsFeature/Components/HelperViews/EmptyAssetListView.swift
```diff
@@ -3,40 +3,60 @@ import SwiftUI
...
 
-extension EmptyAssetListView {
-	static var fungibleResources: Self {
-		.init(viewState: .init(imageAsset: AssetResource.fungibleTokens, description: L10n.AssetDetails.TokenDetails.noTokens))
-	}
+extension EmptyAssetListView.ViewState {
+	static let fungibleResources = Self(
+		imageAsset: AssetResource.fungibleTokens,
+		description: L10n.AssetDetails.TokenDetails.noTokens,
+		glossaryItem: .tokens,
```
CyonAlexRDX, line 38: It feels wrong to have to pass two values: `glossaryItem` **and** `glossaryLabel`. I think the "label" should be a computed property on the enum itself, just like you do with `OverlayWindowClient.GlossaryItem`'s computed property `string`. 

Maybe `GlossaryItem` should not be an enum but rather an "open enum"? i.e. a struct and have `static let tokens = Self(title: ..., image: ..., body: ...)` ?

kugel3: It's different though. The label depends on the context where the button is shown, it's the text shown on the button itself, but the `string` property is what is shown on the sheet, which corresponds directly to the glossaryItem. 

One reason to keep it an enum is that we have a fixed set of "glossaryAnchors" in the markdown.

kugel3: In this particular case, the label is analogous to the `description` property, and indeed they are already organized in the same L10n folder.

Sajjon: ok makes sense

RadixWallet/Clients/OverlayWindowClient/InfoLink.swift
```diff
@@ -0,0 +1,129 @@
+import Foundation
+
+// MARK: - OverlayWindowClient.GlossaryItem
+extension OverlayWindowClient {
+	public enum GlossaryItem: String, Sendable {
```
GhenadieVP, line 5: This enum should ideally be defined in Sargon so both platforms are in sync.

RadixWallet/Clients/OverlayWindowClient/OverlayWindowClient+Interface.swift
```diff
@@ -78,29 +85,17 @@ extension OverlayWindowClient {
 			case emailSupport(additionalInfo: String)
 		}
 
+		public struct SheetState: Sendable, Hashable, Identifiable {
+			public let id = UUID()
+			public let image: ImageAsset?
+			public let text: String
```
GhenadieVP, line 91: Very specific content for a generic sheet, I would opt implementing something to the FullScreenOverlayCoordinator.State.

RadixWallet/Clients/OverlayWindowClient/InfoLink.swift
kugel3: Yeah

kugel3: Adding that in a separate PR

RadixWallet/Clients/OverlayWindowClient/OverlayWindowClient+Interface.swift
```diff
@@ -113,10 +102,34 @@ extension OverlayWindowClient {
...
+			}
+		}
+
 		case hud(HUD)
 		case alert(AlertState)
+		case sheet(SheetOverlayCoordinator.Root.State, SheetBehavior)
 		case fullScreen(FullScreenOverlayCoordinator.State)
 	}
+
+	public enum SheetBehavior: Sendable {
```
GhenadieVP, line 129: Is this to be used to replace the old sheet content when tapping on the embeded info link? if yes, why isn't the replacement simply done in the InfoLinkSheet view?

GhenadieVP: sure, but in the context of info sheet, can't the replace happen internally? what is the reason for having the replace to happen in Overlay reducer? seems like logic specific to the info sheet

RadixWallet/Features/AppFeature/Overlay/FullScreenOverlayCoordinator+Reducer.swift
```diff
@@ -1,5 +1,6 @@
 import ComposableArchitecture
-import SwiftUI
+
+// import SwiftUI
```
matiasbzurovski, line 3: minor: remove

RadixWallet/Features/AppFeature/Overlay/InfoLinkSheet+Reducer.swift
```diff
@@ -0,0 +1,24 @@
+import Foundation
+
+// MARK: - InfoLinkSheet
+public struct InfoLinkSheet: FeatureReducer {
+	public struct State: Sendable, Hashable, Identifiable {
```
matiasbzurovski, line 5: Why do we need this to implement `Identifiable`?

RadixWallet/Features/AppFeature/Overlay/SheetOverlayCoordinator+Reducer.swift
```diff
@@ -0,0 +1,60 @@
+import Foundation
+
+// MARK: - SheetOverlayCoordinator
+public struct SheetOverlayCoordinator: Sendable, FeatureReducer {
+	public struct State: Sendable, Hashable, Identifiable {
```
matiasbzurovski, line 5: Why do we need this to implement `Identifiable`?

RadixWallet/Clients/OverlayWindowClient/OverlayWindowClient+Interface.swift
GhenadieVP: > Ah, you mean that we could do the replacing right inside the info sheet without even involving the OverlayReducer at all?

Yup, that would simplify things.

> One reason I made it like this is that I imagine other situations where we might want to replace the existing sheet rather than enqueueing, for example if we move the Completion sheets to the Overlay.

This enqueueing or replacing should probably be a parameter to the scheduling function itself, or higher level item, as it is not specific to the sheet.

RadixWallet/Features/CreatePersona/Children/Introduction/IntroductionToPersonas.swift
```diff
@@ -4,22 +4,13 @@ import SwiftUI
 @Reducer
 public struct IntroductionToPersonas: Sendable, FeatureReducer {
 	@ObservableState
-	public struct State: Sendable, Hashable {
-		@Presents
-		public var infoPanel: SlideUpPanel.State?
-		public init() {}
-	}
+	public struct State: Sendable, Hashable {}
```
matiasbzurovski, line 7: minor: I would move this into a vanilla SwiftUI view and remove the reducer just to inform delegate.

matiasbzurovski (approved):
works great!

RadixWallet/Features/AppFeature/Overlay/InfoLinkSheet+Reducer.swift
GhenadieVP: No, removed.

PR #1298, opened 2024-08-23
Simplify dependencies

kugel3:
## Description

Gets rid of
- swift-builders (essentially only used for a typealias)
- textbuilder (not used)
- BigInt (not used)
- overture (not used)
- NavigationTransitions (used in a 2-3 places but barely noticeable)
- IdentifiedCollections (imported anyway through TCA)

## Note
A rather specific test, named `test__GIVEN__saved_profile_P__WHEN__we_update_P_changing_its_identity__THEN__identity_is_checked`, had to be disabled


RadixWallet.xcodeproj/project.pbxproj
```diff
@@ -8508,8 +8456,8 @@
 			isa = XCRemoteSwiftPackageReference;
 			repositoryURL = "https://github.com/tgrapperon/swift-dependencies-additions";
 			requirement = {
-				kind = exactVersion;
-				version = 1.0.1;
+				kind = upToNextMajorVersion;
+				minimumVersion = 1.0.1;
```
CyonAlexRDX, line 8460: Unsure about this. We should use **exact** pinning no?

CyonAlexRDX (commented):
Many change to not use exact pinning. It didn't work with exact pinning?

RadixWallet.xcodeproj/project.pbxproj
kugel3: This one actually works with exact pinning now, it didn't last week. The other two I had just forgotten to change back after updating.

matiasbzurovski:
> PresentationBackground (used in a 2-3 places but barely noticeable)

It is actually used in 10 places and IMO the effect is quite noticeable, are we removing it just to reduce dependencies number or is there any conflict if we try to keep it?

kugel3:
Sorry, it's navigationTransition that is used in 2-3 places and is barely noticeable. Presentation background is noticeable, but I don't think it is worth keeping that at the cost of having to use introspection and swizzling. It is not the number of dependencies, it's specifically the introspection dependency that I really would like to get rid of.

matiasbzurovski:
Ok but we are still using Introspection for account header, no?

kugel3:
Yes, but we don't need to, and shouldn't

kugel3:
Alright, I am keeping the presentationBackground for now, but only so we can get this in quickly!

RadixWalletTests/Clients/ProfileStoreTests/ProfileStoreTests.swift
```diff
@@ -712,39 +713,39 @@ final class ProfileStoreExistingProfileTests: TestCase {
...
-				d.assertionFailure = AssertionFailureAction.init(action: { _, _, _ in
-					// THEN identity is checked
-					identityCheckFails.fulfill()
-				})
-			}
-
-			await self.nearFutureFulfillment(of: identityCheckFails)
-		}
-	}
+//	func test__GIVEN__saved_profile_P__WHEN__we_update_P_changing_its_identity__THEN__identity_is_checked() async throws {
```
matiasbzurovski, line 716: should we put this back? or at least comment why it had to be commented

kugel3: We can't put it back because it will not let us assign to `d.assertionFailure`, but we can add a comment

PR #1312, opened 2024-09-03
[ABW-3762] Account lock claim status

matiasbzurovski:
Jira ticket: [ABW-3762](https://radixdlt.atlassian.net/browse/ABW-3762)

## Description
This PR adds the logic to check for pending claims on account lockers, and displaying them on the account rows at Home or inside the Approved dApps view.

## Video
[Link](https://drive.google.com/file/d/156XipTSDNVqd5IJKUuHyaUILBlC2PT2e/view?usp=sharing)

This video shows two sessions. Simulator on the left shows a wallet with different dApps, while simulator on the right shows a regular _consumer_ wallet with two accounts.
- Consumer wallet starts and indicates there is a pending account locker claim for each account. These deposits correspond to `dApp1`, and are displayed for each account at Home.
- User hides deposit claims for `dApp1`, which hides the notifications from previous step.
- A new deposit is made from `dApp2` into `Deny all` account
- After a few seconds, the consumer app detects the deposit and displays it under `Deny all` account
- User reverts change from second step, and shows deposits from `dApp1` again
- App now shows 1 pending deposit for first account, and 2 pending deposits for second account 

[ABW-3762]: https://radixdlt.atlassian.net/browse/ABW-3762?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


RadixWallet/Clients/AccountLockersClient/AccountLockersClient+Live.swift
```diff
@@ -0,0 +1,210 @@
...
+	public static func live() -> AccountLockersClient {
+		@Dependency(\.authorizedDappsClient) var authorizedDappsClient
+		@Dependency(\.accountsClient) var accountsClient
+		@Dependency(\.onLedgerEntitiesClient) var onLedgerEntitiesClient
+		@Dependency(\.gatewayAPIClient) var gatewayAPIClient
+		@Dependency(\.cacheClient) var cacheClient
+
+		let claimsPerAccountSubject = AsyncCurrentValueSubject<ClaimsPerAccount>([:])
+
+		let timerInterval: DispatchTimeInterval = .seconds(600) // 10 minutes
```
CyonAlexRDX, line 17: I think there is a static func called `minutes`?

RadixWallet/Clients/AccountLockersClient/AccountLockersClient+Live.swift
```diff
@@ -0,0 +1,210 @@
...
+			return dapps.compactMap { dapp in
+				guard let dappWithLocker = dappsWithLocker.first(where: { $0.0 == dapp.dappDefinitionAddress }) else {
+					return nil
+				}
+				return DappWithLockerAddress(dapp: dapp, lockerAddress: dappWithLocker.1)
+			}
+		}
+
+		@Sendable
+		func getLockersPerAccount(accounts: Accounts, dapps: [DappWithLockerAddress]) -> [AccountAddress: [String]] {
```
CyonAlexRDX, line 113: Why String? Should be `LockerAddress` right? :)

RadixWallet/Clients/AccountLockersClient/AccountLockersClient+Live.swift
matiasbzurovski: it isn't actually a static func, but a case of the following enum
```Swift
public enum DispatchTimeInterval : Equatable, Sendable {
    case seconds(Int)
    case milliseconds(Int)
    case microseconds(Int)
    case nanoseconds(Int)
}
```

I can make a helper static func though

RadixWallet/Clients/AccountLockersClient/AccountLockersClient+Live.swift
matiasbzurovski: well, to use `LockerAddress` we would need to use the throwing init that takes a Strings and validates it.

Given we will later use its String value only, I don't see the point in doing that actually.

RadixWallet/Clients/CacheClient/CacheClient+Interface.swift
```diff
@@ -161,6 +161,7 @@ extension CacheClient {
 		case rolaWellKnownFileVerification(_ url: String)
 		case tokenPrices(_ currency: FiatCurrency)
 		case dateOfFirstTransaction(_ accountAddress: AccountAddress)
+		case accountLockerClaimDetails(accountAddress: String, lockerAddress: String)
```
GhenadieVP, line 164: Prefer exact Address types

RadixWallet/Clients/AccountLockersClient/AccountLockersClient+Interface.swift
```diff
@@ -0,0 +1,30 @@
...
+
+// MARK: - AccountLockersClient
+public struct AccountLockersClient: DependencyKey, Sendable {
+	public let startMonitoring: StartMonitoring
+
+	/// Async sequence with the AccountLockerClaims for a given Account
+	public let accountClaims: AccountClaims
+
+	/// Async sequence with the set of dapps that have at least one AccountLockerClaim associated to it.
+	public let dappsWithClaims: DappsWithClaims
```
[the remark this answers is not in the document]
matiasbzurovski: yes

RadixWallet/Clients/AccountLockersClient/AccountLockersClient+Interface.swift
```diff
@@ -0,0 +1,30 @@
...
+}
+
+// MARK: - AccountLockerClaims
+/// A struct holding the pending claims for a given locker address & account address.
+public struct AccountLockerClaims: Sendable, Hashable, Codable {
+	let lockerAddress: String
+	let accountAddress: String
+	let dappDefinitionAddress: String
+	let dappName: String?
+	let lastTouchedAtStateVersion: Int64
```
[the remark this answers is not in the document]
matiasbzurovski: you mean to convert the API's `lastTouchedAtStateVersion` into something like `timeIntervalSinceReferenceDate`? I would leave it as is, without modifying it and risking to have the value changed on any scenario, since we only use it to see if it has changed (we don't care about how much time has passed since we last checked, or anything)

RadixWallet/Clients/AccountLockersClient/AccountLockersClient+Live.swift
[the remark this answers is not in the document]
matiasbzurovski: reduced to 5 minutes after today's call

RadixWallet/Clients/AccountLockersClient/AccountLockersClient+Live.swift
```diff
@@ -0,0 +1,210 @@
...
+						continue
+					}
+					var accountLockerClaims: [AccountLockerClaims] = []
+
+					// Loop over each locker associated to this account
+					for lockerState in lockerStates.items {
+						if let cached = getCachedVersionIfValid(lockerState: lockerState) {
+							// The cached claim information is valid.
+							accountLockerClaims.append(cached)
+
```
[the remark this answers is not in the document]
matiasbzurovski: left it on purpose since the `else` body is quite big and helps reading IMO, but will remove

RadixWallet/Clients/AccountLockersClient/AccountLockersClient+Live.swift
```diff
@@ -0,0 +1,210 @@
...
+		@Dependency(\.cacheClient) var cacheClient
+
+		let claimsPerAccountSubject = AsyncCurrentValueSubject<ClaimsPerAccount>([:])
+
+		let timerInterval: DispatchTimeInterval = .seconds(600) // 10 minutes
+
+		@Sendable
+		func startMonitoring() async throws {
+			let dappValues = await authorizedDappsClient.authorizedDappValues()
+			let accountValues = await accountsClient.accountsOnCurrentNetwork()
```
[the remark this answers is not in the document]
matiasbzurovski: yeah, this will trigger on any dapp or account change. For dApps it make sense to trigger always (since linked accounts or deposit hidden/visible must trigger new check). 

For accounts, I will update it so that we only care when the ids of the account change (and not the account details as the name).

RadixWallet/Clients/AccountLockersClient/AccountLockersClient+Interface.swift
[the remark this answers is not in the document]
GhenadieVP: No, I mean that we do have `Epoch` type defined. But it is different from StateVersion from the GW perspective. I would then define a StateVersion typealias at least

RadixWallet/Clients/AccountLockersClient/AccountLockersClient+Live.swift
matiasbzurovski: ended up adding every address type: `LockerAddress`, `ResourceAddress`, `AccountAddress` & `VaultAddress`

CyonAlexRDX: where are we using it as string later? we should do it as late as possible. Yes there is meaning in going through the address type, **like we do everywhere else in the app**, and that is validating that it is a valid address. that is always beneficial.

RadixWallet/Clients/AccountLockersClient/AccountLockersClient+Live.swift
CyonAlexRDX: We cant use `Duration` / `Instant` types, they are from iOS 16: https://developer.apple.com/documentation/Swift/Duration

matiasbzurovski: Yes, but `AsyncTimerSequence` doesn't have an init with `Duration`, only with `DispatchTimeInterval`

RadixWallet/Clients/AccountLockersClient/AccountLockersClient+Live.swift
matiasbzurovski: What I meant is that the `lockerAddress` that is fetched here as a String, is only [used](https://github.com/radixdlt/babylon-wallet-ios/pull/1312/files#diff-8072040d0aad8663b6de7c4bf866234dbe8e39c95d98f94922aab09c16e9e20cR158) to build the `GatewayAPI.AccountLockerAddress` that takes a String as well. Will update so that we parse it as a `LockerAddress` and then get its `var address: String` when needed.

RadixWallet/Features/HomeFeature/Children/AccountRow/Home+AccountRow+Reducer.swift
```diff
@@ -43,9 +46,12 @@ extension Home {
 		@Dependency(\.accountPortfoliosClient) var accountPortfoliosClient
 		@Dependency(\.secureStorageClient) var secureStorageClient
 		@Dependency(\.userDefaults) var userDefaults
+		@Dependency(\.accountLockersClient) var accountLockersClient
 
 		public func reduce(into state: inout State, viewAction: ViewAction) -> Effect<Action> {
 			switch viewAction {
+			case .task:
```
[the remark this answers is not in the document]
GhenadieVP: it is exactly about reducing the number of subscription tasks, which don't always work as expected. Given that all account claims are updated at once, not individually, it makes even more sense to update the rows from the parent.

Also, we were thinking in terms of reducing the TCA bloat, to have eventually one the Home be a reducer, and the children just vanilla swiftUI views view with viewState.

GhenadieVP (approved):

PR #1315, opened 2024-09-04
[ABW-3776] Unable to create new persona fix 

danvleju-rdx:
Jira ticket: [ABW-3776](https://radixdlt.atlassian.net/browse/ABW-3776)

## Description
This PR fixes the issue where users could not create a new persona by explicitly declaring the `body`. This was necessary due to a conflict between our `FeatureReducer` protocol and the `@Reducer` macro.

[ABW-3776]: https://radixdlt.atlassian.net/browse/ABW-3776?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/RadixWallet/Features/CreatePersona/Children/Introduction/IntroductionToPersonas.swift
+++ b/RadixWallet/Features/CreatePersona/Children/Introduction/IntroductionToPersonas.swift
@@ -15,6 +15,10 @@ public struct IntroductionToPersonas: Sendable, FeatureReducer {
 
 	public init() {}
 
+	public var body: some ReducerOf<Self> {
+		Reduce(core)
+	}
+
 	public func reduce(into state: inout State, viewAction: ViewAction) -> Effect<Action> {
 		switch viewAction {
 		case .continueButtonTapped:


matiasbzurovski:
This looks good to me, but I insist we should do what I proposed on its [PR](https://github.com/radixdlt/babylon-wallet-ios/pull/1293#discussion_r1734962224): there is no point for this view to have a reducer with a `ViewAction` that its only logic is to inform the delegate via its only `DelegateAction`

Makes more sense to have it as a plain SwiftUI view that takes a closure instead.

CyonAlexRDX (approved):
oh scary! Have you gone through other (all? 😬 ) Reducers? We might have more bugs caused by this?!

danvleju-rdx:
@CyonAlexRDX Yes, all good. We have only 10 places (for now) where we use `@Reducer` macro.

danvleju-rdx:
@matiasbzurovski Agree with you. Should I do this change in this PR though?

GhenadieVP (approved):

GhenadieVP:
@danvleju-rdx @matiasbzurovski Let's do the improvement in a separate PR.

PR #1325, opened 2024-09-09
Minor fixes

GhenadieVP:
- Show info icon for linking accounts also in TX Review.
- Update the Sargon dependency to fix the Decimal.one build error.


CyonAlexRDX (approved):
LGTM

PR #1326, opened 2024-09-09
Biometrics check updates

danvleju-rdx:
Complements #1286 

## Description
- Request biometrics when user tries to enable Advanced Lock from Preferences.
- Display a message for users on iOS 18 to inform them that they can now use the App Lock system feature.

## Videos

https://github.com/user-attachments/assets/4299cbcb-d52e-4e07-8c80-a71b5b3523b8

https://github.com/user-attachments/assets/bc7eb398-c4fa-4fe2-95b0-cd01822685a6


RadixWallet/Features/SplashFeature/Splash.swift
```diff
@@ -152,17 +159,35 @@ public struct Splash: Sendable, FeatureReducer {
 			}
 
 			return delegateCompleted(context: state.context)
+
+		case .showAppLockMessage:
+			state.destination = .errorAlert(.init(
+				title: .init("Warning"),
+				message: .init("To protect your app, use the App Lock system feature"),
```
danvleju-rdx, line 166: TODO: add localization keys

CyonAlexRDX, line 166: Is the copy final? I think "to protect your app"is too generic a message, especially with the title "Warning", it makes it sound like users are "unsafe" without it.

Probably not your copy! But I suggest you bring this up with Matt / Charlie. 

I would prefer more precise wording alluding to the missed out feature and the purpose of the lock, something along the lines of:
Title: "Recommendation"
Message: "For maximum privacy you might wanna consider enabling Lock Screen"

CyonAlexRDX (approved):
Code looks good, I think we should change the wording though

RadixWallet/Features/SplashFeature/Splash.swift
```diff
@@ -152,17 +159,35 @@ public struct Splash: Sendable, FeatureReducer {
...
+				buttons: [
```
matiasbzurovski, line 167: minor: we should use `init(title:actions:message:)` since this one is deprecated

RadixWallet/Features/SplashFeature/Splash.swift
danvleju-rdx: Yes, already asked Matt

RadixWallet/Features/SplashFeature/Splash.swift
```diff
@@ -91,7 +94,11 @@ public struct Splash: Sendable, FeatureReducer {
 					#endif
 					await send(.internal(.advancedLockStateLoaded(isEnabled: isEnabled)))
 				} else {
-					await send(.internal(.advancedLockStateLoaded(isEnabled: false)))
+					if !userDefaults.appLockMessageShown {
```
matiasbzurovski, line 97: Will this be a generic error message, that it applies for users with iOS 18 (with native App Lock support) and users with iOS 16/17 (that would need to use our custom app lock feature)?

Because I understood on Friday that it was meant only for iOS 18 users, in which case we need to add the `if #available` check to avoid situations as the following:

Screenshot 2024-09-09 at 12 51 10

matiasbzurovski: Also, do we want to have this on Splash view? Maybe it makes more sense to show it next time the user opens the Wallet after setting it up. 

Otherwise, this is the first message that a new user who just downloaded the Wallet on iOS 18 sees.

danvleju-rdx: >Because I understood on Friday that it was meant only for iOS 18 users, in which case we need to add the if #available check to avoid situations as the following:

You're right, an if #available check is needed. Will fix it.

>Also, do we want to have this on Splash view? Maybe it makes more sense to show it next time the user opens the Wallet after setting it up.
Otherwise, this is the first message that a new user who just downloaded the Wallet on iOS 18 sees.

Good point! In this case we can stick with the requirement to inform only users who had enabled Advanced Lock before iOS 18 update. The reason for displaying the message on the Splash view is that, at this step, users expect to authenticate with biometrics (like it used to be for them before iOS 18). Will check with Matt.

matiasbzurovski (approved):
LGTM (pending localisation)

GhenadieVP (approved):

PR #1338, opened 2024-09-12
[ABW-3772] SVG image support

danvleju-rdx:
Jira ticket: [ABW-3772](https://radixdlt.atlassian.net/browse/ABW-3772)

## Description
This PR adds support for SVG images. The conversion of SVGs to PNGs is handled on the image service side by specifying a new query parameter: `format=png`.

In addition to regular URLs, the wallet now supports data URLs (see [RFC 2397](https://datatracker.ietf.org/doc/html/rfc2397)). Support for data URLs by the image service is currently in progress. In the meantime, data URLs will appear as broken images. Once the image service adds support, data URL images will display correctly without any changes required on the wallet side.

## Screenshot
- The "SVGenesis" collection contains data URLs images generated [here](https://svgenesis.trove.tools)

| 1 | 2 | 3 - data URL |
| - | - | - |
| Simulator Screenshot - iPhone 15 Pro - 2024-09-24 at 11 58 49 | Simulator Screenshot - iPhone 15 Pro - 2024-09-24 at 11 58 57 | Simulator Screenshot - iPhone 15 Pro - 2024-09-24 at 11 59 17 |

[ABW-3772]: https://radixdlt.atlassian.net/browse/ABW-3772?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


RadixWallet/Clients/URLFormatterClient/URLFormatterClient+Live.swift
```diff
@@ -23,10 +23,15 @@ extension URLFormatterClient: DependencyKey {
 	#endif
 
 	private static func makeURL(url: URL, imageServiceURL: URL, size: CGSize) -> URL {
-		let originItem = URLQueryItem(name: "imageOrigin", value: url.absoluteString)
-		let sizeItem = URLQueryItem(name: "imageSize", value: imageSizeString(size: size))
-
-		return imageServiceURL.appending(queryItems: [originItem, sizeItem])
+		if url.isDataURL {
+			let imageOrigin = url.absoluteString.addingPercentEncoding(withAllowedCharacters: .alphanumerics) ?? ""
```
danvleju-rdx, line 27: Parses the `:/+;` characters from `data:[<mediatype>][;base64],<data>`

RadixWallet/Clients/URLFormatterClient/URLFormatterClient+Live.swift
```diff
@@ -23,10 +23,15 @@ extension URLFormatterClient: DependencyKey {
...
+			let imageSize = imageSizeString(size: size)
+			return URL(string: "\(imageServiceURL.absoluteString)/?imageOrigin=\(imageOrigin)&imageSize=\(imageSize)")!
```
matiasbzurovski, line 29: Why not use `URL.appending(queryItems:)` instead of manually buliding it and force casting it?

RadixWallet/Clients/BootstrapClient/BootstrapClient+Live.swift
```diff
@@ -7,6 +9,11 @@ extension BootstrapClient: DependencyKey {
 			bootstrap: {
 				appsFlyerClient.start()
 				homeCardsClient.bootstrap()
+
+				ImageDecoderRegistry.shared.register { context in
+					let isSVG = context.urlResponse?.url?.isSVG ?? false
+					return isSVG ? ImageDecoders.Empty() : nil
+				}
```
matiasbzurovski, line 16: I would rather keep this `bootstrap` method independent of specific logic. Until now, it just delegates the `bootstrap`/`start` calls to the corresponding clients (and it is agnostic of what AppsFlyer and HomeCards need to do when the app starts).

I don't think it justifies creating an `ImageDecodingClient` just for this method, so maybe we can do the following?

```Swift
// On new file called Nuke+Extensions or ImageDecoderRegistry+Extensions
extension ImageDecoderRegistry {
	static func bootstrap() {
		ImageDecoderRegistry.shared.register { context in
			let isSVG = context.urlResponse?.url?.isSVG ?? false
			return isSVG ? ImageDecoders.Empty() : nil
		}
	}
}

// On BootstrapClient (which removes the import Nuke)
{
    appsFlyerClient.start()
    homeCardsClient.bootstrap()
    ImageDecoderRegistry.bootstrap()
}
```

RadixWallet/Core/DesignSystem/Components/Thumbnails.swift
```diff
@@ -348,5 +381,5 @@ extension URL {
 		return false
 	}
 
-	private static let vectorImageTypes: [String] = ["svg", "pdf"]
+	private static let unsupportedVectorImageTypes: [String] = ["pdf"]
```
matiasbzurovski, line 384: do is there any other type of vector images asides `pdf` (and the now supported `svg`)? Otherwise we can just remove the for loop and check if it has `.pdf` suffix.

RadixWallet/Core/DesignSystem/Components/Thumbnails.swift
```diff
@@ -180,8 +181,8 @@ public struct LoadableImage<Placeholder: View>: View {
 		placeholder: () -> Placeholder
 	) {
 		if let url {
-			if url.isVectorImage {
-				loggerGlobal.warning("LoadableImage: Vector images are not supported \(url)")
+			if url.isUnsupportedVectorImage {
+				loggerGlobal.warning("LoadableImage: Vector image is not supported \(url)")
```
matiasbzurovski, line 185: And I would change the warning here to `LoadableImage: PDF vector images are not supported \(url)`

matiasbzurovski (changes requested):
looking good, added some comments/questions

RadixWallet/Clients/BootstrapClient/BootstrapClient+Live.swift
danvleju-rdx: Good idea! However, `import Nuke` will still be needed in `BootstrapClient` since we use `ImageDecoderRegistry` there.

RadixWallet/Core/DesignSystem/Components/Thumbnails.swift
danvleju-rdx: I think you're right. I'll remove the for loop and add an `isPDF` check instead.

danvleju-rdx: Addressed feedback [here](https://github.com/radixdlt/babylon-wallet-ios/pull/1338/commits/9bd93c181bcdd8077965533eaba4c849b447e99d). Defined a new method `isVectorImage(type:)` instead.

matiasbzurovski: much better!

RadixWallet/Clients/URLFormatterClient/URLFormatterClient+Live.swift
matiasbzurovski: oh I see. Then I would add a comment similar to this on the code so we avoid changing it in the future. Also, may be useful adding some unit tests?

RadixWallet/Clients/URLFormatterClient/URLFormatterClient+Live.swift
```diff
@@ -23,10 +23,30 @@ extension URLFormatterClient: DependencyKey {
...
-		let originItem = URLQueryItem(name: "imageOrigin", value: url.absoluteString)
-		let sizeItem = URLQueryItem(name: "imageSize", value: imageSizeString(size: size))
+		if url.isDataURL {
+			// Data URLs (e.g., `data:image/svg+xml,`) require special encoding to handle special characters.
+			// Here, we encode the absolute string using percent encoding, allowing only alphanumeric characters.
+			// This is needed because `url.appending(queryItems:)` does not encode special characters correctly for data URLs.
+			let imageOrigin = url.absoluteString.addingPercentEncoding(withAllowedCharacters: .alphanumerics) ?? ""
+			let imageSize = imageSizeString(size: size)
+
+			// Manually construct the query string to ensure that the data URL is correctly encoded
```
CyonAlexRDX, line 33: We would probably benefit from having this written in Sargon. WDYT?

danvleju-rdx: Yeah, especially with `isVectorImage(type:)`. Android will also need to check if an image is an SVG.

RadixWallet/Clients/URLFormatterClient/URLFormatterClient+Live.swift
```diff
@@ -23,10 +23,30 @@ extension URLFormatterClient: DependencyKey {
...
+			// and to avoid issues like double-encoding, which could corrupt the URL.
+			var urlString = "\(imageServiceURL.absoluteString)/?imageOrigin=\(imageOrigin)&imageSize=\(imageSize)"
+
+			if url.isVectorImage(type: .svg) {
+				urlString.append("&format=png")
+			}
+
+			return URL(string: urlString)!
```
matiasbzurovski, line 41: would it be better to default to original url (e.g. `return URL(string: urlString) ?? url)` instead of crashing, if for some reason we receive a weird url for which our manual construction fails?

CyonAlexRDX (approved):
LGTM, good job

GhenadieVP (approved):

GhenadieVP (approved):

matiasbzurovski (approved):

PR #1343, opened 2024-09-16
[ABW-3581] Inform users when signing with Ledger without P2PLinks

matiasbzurovski:
Jira ticket: [ABW-3581](https://radixdlt.atlassian.net/browse/ABW-3581)

## Description
Inform users when they attempt to access a Ledger factor source but the wallet isn't connected to any connector extension. 
Also, improves the UX by removing a glitch where the ledger name wouldn't show up for an instant.

## Videos

| Before | After |
| - | - |
| <video src=https://github.com/user-attachments/assets/65ba3267-6b8b-428d-ac30-9b021bdfc454> | <video src=https://github.com/user-attachments/assets/902b2f3a-5ccf-41e2-8597-51bac5507a24> |

[ABW-3581]: https://radixdlt.atlassian.net/browse/ABW-3581?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/RadixWallet/Features/FactorSourceAccess/FactorSourceAccess.swift
+++ b/RadixWallet/Features/FactorSourceAccess/FactorSourceAccess.swift
@@ -20,7 +20,7 @@ public struct FactorSourceAccess: Sendable, FeatureReducer {
 	}
 
 	public enum InternalAction: Sendable, Hashable {
-		case isConnectedToAnyConnectorExtension(Bool)
+		case hasP2PLinks(Bool)
 	}
 
 	public enum DelegateAction: Sendable, Hashable {
@@ -48,7 +48,7 @@ public struct FactorSourceAccess: Sendable, FeatureReducer {
 		}
 	}
 
-	@Dependency(\.ledgerHardwareWalletClient) var ledgerHardwareWalletClient
+	@Dependency(\.p2pLinksClient) var p2pLinksClient
 
 	public var body: some ReducerOf<Self> {
 		Reduce(core)
@@ -75,8 +75,8 @@ public struct FactorSourceAccess: Sendable, FeatureReducer {
 
 	public func reduce(into state: inout State, internalAction: InternalAction) -> Effect<Action> {
 		switch internalAction {
-		case let .isConnectedToAnyConnectorExtension(isConnected):
-			if !isConnected {
+		case let .hasP2PLinks(hasP2PLinks):
+			if !hasP2PLinks {
 				state.destination = .noP2PLink(.noP2Plink)
 			}
 			return .none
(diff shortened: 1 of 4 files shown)


RadixWallet/Features/Signing/Children/SignWithFactorSource.swift
```diff
@@ -24,7 +24,8 @@ public struct SignWithFactorSource: Sendable, FeatureReducer {
 				self.factorSourceAccess = .init(kind: .device, purpose: .signature)
 			case .ledger:
 				assert(signingFactors.allSatisfy { $0.factorSource.kind == LedgerHardwareWalletFactorSource.kind })
-				self.factorSourceAccess = .init(kind: .ledger(nil), purpose: .signature)
+				let ledger: LedgerHardwareWalletFactorSource? = signingFactors.first?.factorSource.extract()
```
CyonAlexRDX, line 27: Hmm this looks wrong? There is a missing filter and or `first(where { $0.kind == .ledger})`?

matiasbzurovski: if you look on the above line, we are asserting that every signing factor meets such condition

CyonAlexRDX: maybe im sicker than I thought :)

danvleju-rdx (approved):
LGTM

RadixWallet/Features/FactorSourceAccess/FactorSourceAccess.swift
```diff
@@ -4,33 +4,110 @@ public struct FactorSourceAccess: Sendable, FeatureReducer {
...
+			}
+		}
+	}
+
+	private func checkP2PLinksEffect(state: State) -> Effect<Action> {
+		guard case .ledger = state.kind else {
+			return .none
+		}
+		return .run { send in
+			for try await isConnected in await ledgerHardwareWalletClient.isConnectedToAnyConnectorExtension() {
```
GhenadieVP, line 103: I don't think this is what we should check. This can return false if there are P2PLinks, but these are not connected(you browser is closed), I believe that the intention was to inform users when there are no P2PLinks at all.

GhenadieVP (changes requested):

RadixWallet/Features/FactorSourceAccess/FactorSourceAccess.swift
matiasbzurovski: right, will update it to check if there are any p2pLinks instead. 

Do we want to leave it this way when [adding a new Ledger device](https://github.com/radixdlt/babylon-wallet-ios/blob/main/RadixWallet/Features/LedgerHardwareDevices/LedgerHardwareDevices%2BReducer.swift#L206-L210)? Because in such case we will show this alert if there is no active connection, and user could be lead to add a new one when all they need to do is open their browser

GhenadieVP (approved):

PR #1350, opened 2024-09-20
[ABW-3801] Xcode 16 support

danvleju-rdx:
Jira ticket: [ABW-3801](https://radixdlt.atlassian.net/browse/ABW-3801)

## Description
This PR adds support for Xcode 16, with the `SWIFT_STRICT_CONCURRENCY` flag set to `complete`.
It still works with Xcode 15.

Support for Swift 6 will be addressed in a separate PR.

[ABW-3801]: https://radixdlt.atlassian.net/browse/ABW-3801?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


RadixWallet/Clients/AccountPortfoliosClient/AccountPortfoliosClient+Live.swift
```diff
@@ -28,16 +21,31 @@ extension AccountPortfoliosClient: DependencyKey {
...
+		let state = State()
+
+		@Dependency(\.onLedgerEntitiesClient) var onLedgerEntitiesClient
+		@Dependency(\.cacheClient) var cacheClient
+		@Dependency(\.tokenPricesClient) var tokenPricesClient
+		@Dependency(\.appPreferencesClient) var appPreferencesClient
+		@Dependency(\.resourcesVisibilityClient) var resourcesVisibilityClient
+		@Dependency(\.gatewaysClient) var gatewaysClient
+
+		Self.registerSubscribers(state: state)
```
matiasbzurovski, line 42: Is this change just a refactor for cleanup, or do we actually need to call the static method for subscriptions to take place? Because if we do, and it is a consequence of `SWIFT_STRICT_CONCURRENCY`, then there are other clients where we need to follow the path

danvleju-rdx: This is necessary to resolve the compiler error: `Default argument cannot be both actor-isolated and actor-isolated`. The issue arises from using `state` (which is an actor) inside a Task. If other clients follow this pattern, they will encounter the same error. I’m not entirely sure why calling the static method resolves the issue, but it works...

matiasbzurovski: thanks for the explanation. Guess that if it is compiling on other clients (where we do create `Task`s that handle suscriptions but probably without modifying an `actor`)  then we are fine

CyonAlexRDX (approved):
Good job! I guess `Swift 6` mode will be the scary big PR right?

matiasbzurovski (approved):

danvleju-rdx:
@CyonAlexRDX Right, Swift 6 is the big part

GhenadieVP (approved):

PR #1354, opened 2024-09-25
Fix stake unit resource redemption value

GhenadieVP:
The wallet was mistakenly calculating the worth of stake unit resource amount instead of the xrd redemption value.

## Image with the XRD worth - 0.0216 USD XRD price
IMG_0212

## Image with wrong stake unit worth
This multiplies the stake unit resource amount with the XRD price
IMG_0213

## Image with fixed stake unit worth
This multiplies the xrd redemption with xrd price
IMG_0211


CyonAlexRDX (approved):
LGTM

PR #1359, opened 2024-09-30
Add Xcode template for features with @ObservableState

matiasbzurovski:
Adding it as a new template because we still may need to use the _old_ version when creating a feature that uses other one which doesn't have @Reducer and @ObservableState.


templates/Xcode/TCA Observable Feature.xctemplate/___FILEBASENAME___.swift
```diff
@@ -0,0 +1,23 @@
...
+	}
+
+	public typealias Action = FeatureAction<Self>
+
+	public enum ViewAction: Sendable, Equatable {
+		case appeared
+	}
+
+	public init() {}
+
```
CyonAlexRDX, line 16: Should we not include the body here explicitly? I think @danvleju-rdx found out it was needed? Or I might misremember

CyonAlexRDX (approved):
Nice

templates/Xcode/TCA Observable Feature.xctemplate/___FILEBASENAME___.swift
matiasbzurovski: I've seen it as well, but I am able to make it compile without it. @danvleju-rdx, if it compiles without it does it mean we can drop it, or was it added for runtime behavior differences?

matiasbzurovski: just realised it was the second, so added it to template

danvleju-rdx: Yes, runtime issues, [see](https://github.com/radixdlt/babylon-wallet-ios/pull/1315)

danvleju-rdx (approved):

PR #1365, opened 2024-10-03
[ABW-3817][ABW-3263] Accounts & Persona proof of ownership

matiasbzurovski:
Jira ticket: [ABW-3817](https://radixdlt.atlassian.net/browse/ABW-3817)
Jira ticket: [ABW-3263](https://radixdlt.atlassian.net/browse/ABW-3263)

## Description
This PR handles the dApp interaction to prove ownership of Accounts and/or Persona.

## How to test
- Connect to the [sample dApp](https://d1rb5siu4yt28b.cloudfront.net/) that allows sending proof of ownership requests.
- Validate that the app behaves correctly for every potential scenario described in the JIRA tickets.

[ABW-3817]: https://radixdlt.atlassian.net/browse/ABW-3817?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[ABW-3263]: https://radixdlt.atlassian.net/browse/ABW-3263?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


RadixWallet/Features/DappInteractionFeature/Children/Login/Row/PersonaRow+View.swift
```diff
@@ -22,53 +22,89 @@ extension PersonaRow {
 	@MainActor
 	struct View: SwiftUI.View {
 		let viewState: ViewState
-		let isSelected: Bool
-		let action: () -> Void
+		let mode: Mode
```
matiasbzurovski, line 25: Changes in this view are basically allowing to have `selection` mode (as it was currently used), and `display` mode (which doesn't show a `RadioButton` on the trailing side nor embeds the `View` inside a `Button`.

RadixWallet/Features/DappInteractionFeature/Coordinator/DappInteractionFlow.swift
```diff
@@ -199,41 +210,28 @@ struct DappInteractionFlow: Sendable, FeatureReducer {
...
-					let authorizedPersona = authorizedDapp.referencesToAuthorizedPersonas.first(where: { $0.identityAddress == identityAddress })
-				else {
-					await send(.internal(.presentPersonaNotFoundErrorAlert(reason: "")))
-					return
-				}
-
-				await send(.internal(.usePersona(usePersonaRequestItem, persona, authorizedDapp, authorizedPersona)))
-
-			} catch: { error, send in
-				await send(.internal(.presentPersonaNotFoundErrorAlert(reason: error.legibleLocalizedDescription)))
```
matiasbzurovski, line 220: Moved inside `validateAuthUsePersonaEffect`

RadixWallet/Features/DappInteractionFeature/Coordinator/DappInteractionFlow.swift
```diff
@@ -303,6 +301,54 @@ struct DappInteractionFlow: Sendable, FeatureReducer {
...
+			if let accountAddresses = requestItem.accountAddresses {
+				// For every accountAddress set, verify we can access its corresponding Account
+				let allAccounts = try await accountsClient.getAccountsOnCurrentNetwork()
+				let accounts = allAccounts.filter {
+					accountAddresses.contains($0.address)
+				}
+				if accounts.count != accountAddresses.count {
+					await send(.internal(.handleInvalidPersonaOrAccounts))
+				}
+			}
```
matiasbzurovski, line 349: When we have a `DappToWalletInteractionProofOfOwnershipRequestItem`, we will need to prove ownership of:
- 1 Persona
- N Accounts
- 1 Persona & N Accounts

In each corresponding view, we already perform the logic added here to get the actual entities given its addresses, as well as handling the case where at least one address doesn't correspond to a `Persona`/`Account` that the user owns.

However, we still want to perform this check here to avoid a situation when verifying both a persona and accounts. If the `Persona` is correctly own, but any of the `Accounts` isn't, we want to `.handleInvalidPersonaOrAccounts` before the `Persona` ownership is proven. This way we prevent the user from verifying the request _partially_, when we know the end result would be a failure.


I know, you will say that the logic to fetch the `Persona` & `Accounts` is repeated on the `PersonaProofOfOwnership` & `AccountsProofOfOwnership` reducers. Ideally, we would just update the states to inject the given entities (and not have the reducers fetch them again), but I couldn't find any way of doing so without making the code a lot uglier than how it looks now. 
Also, considering the state needs to be created before making this check (and at such point we don't have the entities yet), we would end up with either checking an optional value isn't nil or leaving the logic as it is now anyway.

RadixWallet/Features/DappInteractionFeature/Children/PersonaProofOfOwnership/PersonaProofOfOwnership.swift
```diff
@@ -0,0 +1,107 @@
...
+		.run { send in
+			let persona = try await personasClient.getPersona(id: state.identityAddress)
+			await send(.internal(.setPersona(persona)))
+		} catch: { error, send in
+			loggerGlobal.error("Failed to get Persona to proove its ownership, \(error)")
+			await send(.delegate(.failedToGetPersona))
+		}
+	}
+
+	private func signChallengeEffect(state: State) -> Effect<Action> {
```
Sajjon, line 75: I would really prefer it if this function could be merged with `gatherSignaturePayloadsEffect`

matiasbzurovski: I don't think this is something we want to do. 

For `PersonaProofOfOwnership` (`signChallengeEffect`), the logic could be extracted to some client or global function because the signature is handled by the `deviceFactorSourceClient`.

However, for `AccountsProofOfOwnership` (`gatherSignaturesEffect`), the solution takes two steps: first we gather the signatures inside the mentioned effect, and then we need to present the `Signing.View`, and perform specific logic on each of the 4 delegate calls that the `Signing` reducer can make. 

There aren't really too many things in common, asides from the creation of a `AuthenticationDataToSignForChallengeRequest` used by the `RolaClient`.

RadixWallet/Clients/DappInteractionClient/DappInteractionClient+Live.swift
```diff
@@ -156,3 +161,30 @@ extension DappInteractionClient {
...
+				let _ = try await personasClient.getPersona(id: identityAddress)
+			}
+			if let accountAddresses {
+				// For every accountAddress set, verify we can access its corresponding Account
+				let allAccounts = try await accountsClient.getAccountsOnCurrentNetwork()
+				let accounts = allAccounts.filter {
+					accountAddresses.contains($0.address)
+				}
+				if accounts.count != accountAddresses.count {
+					return false
```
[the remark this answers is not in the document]
matiasbzurovski: yes, it will be invalid because one cannot prove ownership about hidden accounts/personas, and if even one entity is missing then we can stop the interaction at this point

@GhenadieVP has confirmed this on last design session as well.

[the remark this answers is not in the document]
GhenadieVP: yup, the response is all or nothing, no partial response.

RadixWallet/Features/DappInteractionFeature/Children/AccountsProofOfOwnership/AccountsProofOfOwnership.swift
```diff
@@ -0,0 +1,194 @@
...
+			case let .finishedSigning(.signAuth(signedAuthChallenge)):
+				state.destination = nil
+
+				var accountsLeftToVerifyDidSign: Set<Account.ID> = Set(state.accounts.map(\.id))
+
+				let accountAuthProofs: [AccountAuthProof] = signedAuthChallenge.entitySignatures.compactMap { signature in
+					guard let account = try? signature.signerEntity.asAccount() else {
+						return nil
+					}
+					accountsLeftToVerifyDidSign.remove(account.id)
```
GhenadieVP, line 117: A matter of taste, but you could use create a Set from AccountAuthProofs account ids, and compare it with the `Set(state.accounts.map(\.id)) == Set(accountAuthProofs.map(\.account.id))`

RadixWallet/Features/DappInteractionFeature/Children/PersonaProofOfOwnership/PersonaProofOfOwnership.swift
GhenadieVP: @CyonAlexRDX in the context of MFA, for personas it would be possible to have other factor sources right?

Also, wouldn't the following the return the device factor source when signing for a persona? I mean instead of using specifically the device factor source, we can let the factorSourcesClient drive this, and cover the potential future scenarios when there will be other other factor sources that can sign for persona.
```
let signingFactors = try await factorSourcesClient.getSigningFactors(.init(
				networkID: accountsClient.getCurrentNetworkID(),
				signers: signers,
				signingPurpose: .signAuth
			))
```

CyonAlexRDX: Let's sync tomorrow. I think this will be replaced by ROLASigner which @micbakos-rdx writes according to the ticket

matiasbzurovski: Updated solution and isolated the logic on `SignProofOfOwnership`

RadixWallet/Features/DappInteractionFeature/Children/AccountsProofOfOwnership/AccountsProofOfOwnership.swift
matiasbzurovski: I am not only doing the comparison here, but also creating the `WalletToDappInteractionAuthProof` for each signature.

RadixWallet/Features/DappInteractionFeature/Children/AccountsProofOfOwnership/AccountsProofOfOwnership.swift
```diff
@@ -0,0 +1,137 @@
...
+
+	enum DelegateAction: Sendable, Equatable {
+		case provenOwnership([AccountAuthProof], SignedAuthChallenge)
+		case failedToGetAccounts
+		case failedToSign
+	}
+
+	@CasePathable
+	enum ChildAction: Sendable, Equatable {
+		case signature(SignProofOfOwnership.Action)
```
CyonAlexRDX, line 45: i think `signing` or `gather_proof` is a better name than `signature`

CyonAlexRDX (commented):
nice that the signature gather has been put in `SignProofOfOwnership`, still if feels like `PersonaProofOfOwnership` and `AccountProofOfOwnership` ought to have been one and same feature generic over entity type.

RadixWallet/Features/DappInteractionFeature/Children/AccountsProofOfOwnership/AccountsProofOfOwnership.swift
matiasbzurovski: Renamed to `sign` to follow the naming use for the other reducer

matiasbzurovski:
@CyonAlexRDX merged the two reducers into one single `ProofOfOwnership`.

Decided to leave the signature part isolated on `SignProofOfOwnership`, given that it is already implemented and helps keep the `ProofOfOwnership` a small and clean, but can merge if you think it is better.

GhenadieVP (approved):
Amazing work!

PR #1366, opened 2024-10-07
[ABW-3869] Fix crash when transferring multiple NFTs from same collection

matiasbzurovski:
Jira ticket: [ABW-3869](https://radixdlt.atlassian.net/browse/ABW-3869)

## Description
This PR fixes a crash caused by attempting to add multiple values with same `Key` to a dictionary. The solution for this particular case required adding a new wrapper model to use `IdentifiedArrayOf<>` instead. 
Still, the PR adds a custom throwing init for `Dictionary` and `OrderedDictionary` to at least catch these potential failures in other situations. 

## Video

https://github.com/user-attachments/assets/9c6d3cf6-9e8e-4e8b-b006-040307c5c469


CyonAlexRDX (approved):
nice, good job!

PR #1368, opened 2024-10-10
[ABW-3858] Submit transaction and poll its status via SargonOS

matiasbzurovski:
Jira ticket: [ABW-3858](https://radixdlt.atlassian.net/browse/ABW-3858)

## Dependencies
Depends on https://github.com/radixdlt/babylon-wallet-ios/pull/1234

## Description
Use SargonOS to submit a transaction and then poll for its status.
Also, fixes an issue where the `Completion.View` was displayed full screen on iOS 18.

## Video

https://github.com/user-attachments/assets/d1416f64-5f07-44f7-a454-985b30fbc413

## Screenshot

| Before | After |
| - | - |
|  |  |

[ABW-3858]: https://radixdlt.atlassian.net/browse/ABW-3858?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/RadixWallet/Features/AccountPreferencesFeature/Children/ThirdPartyDeposits/ThirdPartyDeposits+Reducer.swift
+++ b/RadixWallet/Features/AccountPreferencesFeature/Children/ThirdPartyDeposits/ThirdPartyDeposits+Reducer.swift
@@ -140,10 +140,11 @@ struct ManageThirdPartyDeposits: FeatureReducer, Sendable {
 					if case let .transaction(tx) = success.items {
 						/// Wait for the transaction to be committed
 						let txID = tx.send.transactionIntentHash
-						try await submitTXClient.hasTXBeenCommittedSuccessfully(txID)
-						/// Safe to update the account to new state
-						try await accountsClient.updateAccount(updatedAccount)
-						await send(.internal(.updated(updatedAccount)))
+						if try await submitTXClient.hasTXBeenCommittedSuccessfully(txID) {
+							/// Safe to update the account to new state
+							try await accountsClient.updateAccount(updatedAccount)
+							await send(.internal(.updated(updatedAccount)))
+						}
 						return
 					}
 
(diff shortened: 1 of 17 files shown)


CyonAlexRDX (approved):
AWESOME! Can we take a moment to reflect about how absolutely amazing this is?! This is using so much advanced tech, with UniFFI and NetworkDrivers implemeneted in Swift, used From Rust, call from Swift, and it is working. With polling... fantastic!

RadixWallet/Features/AccountPreferencesFeature/Children/ThirdPartyDeposits/ThirdPartyDeposits+Reducer.swift
```diff
@@ -155,10 +156,6 @@ struct ManageThirdPartyDeposits: FeatureReducer, Sendable {
 				}
 
 			} catch {
-				/// Polling failure will be displayed in SubmiTransactionView
```
GhenadieVP, line 158: Should this go away?

matiasbzurovski: yes! we no longer need to catch the error, since the client will now return `TransactionStatus.permanentlyRejected` (rather than throw `TXFailureStatus.permanentlyRejected` as it used to do). 

Then, inside [SubmitTransaction](https://github.com/radixdlt/babylon-wallet-ios/pull/1368/files/12cb70b31f4db067b9a851f444440a080fb950eb#diff-892df0bbb6ddf6c18999e02fb9cd9942360dcbe2c84cb787b672bbf870d87c01R133), we react the same as we did, but we no longer have to add this "if error thrown correspond to a transaciton failure, then ignore error" kind of logic in every reducer that submits a transaction :)

GhenadieVP (approved):

PR #1370, opened 2024-10-17
Remove public access modifier

matiasbzurovski:
Remove the `public` modifier everywhere where it isn't necessary


CyonAlexRDX (approved):
LGTM

PR #1371, opened 2024-10-21
[ABW-3821] Pre Authorization general UI 

matiasbzurovski:
JIRA ticket: [ABW-3821](https://radixdlt.atlassian.net/browse/ABW-3821)
JIRA ticket: [ABW-3823](https://radixdlt.atlassian.net/browse/ABW-3823)

## Description 
Adds `PreAuthorizationReview`, as well as logic to present it from the given dApp interaction.

## How to review
The first thing I would do is go over `TransactionReview` & `TransactionReview+View`.  No additional behavior has been added to this flow, but you will see a lot of code has been migrated into the new `InteractionReview.Sections`. This module contains the shared components among `TransactionReview` and `PreAuthorizationReview`.

Once the "common code" refactoring has been reviewed, I'd review the new code over `PreAuthorizationReview` & `PreAuthorizationReview+View`.

[ABW-3821]: https://radixdlt.atlassian.net/browse/ABW-3821?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[ABW-3823]: https://radixdlt.atlassian.net/browse/ABW-3823?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


RadixWallet/Features/TransactionReviewFeature/TransactionReviewAccount/TransactionReviewAccount.swift
```diff
@@ -1,91 +0,0 @@
-import ComposableArchitecture
-import SwiftUI
-
-// MARK: - TransactionReviewAccounts
-struct TransactionReviewAccounts: Sendable, FeatureReducer {
```
matiasbzurovski, line 5: Moved into `InteractionReview/Sections/Accounts/Accounts.swift`

RadixWallet/Features/TransactionReviewFeature/TransactionReviewDapps/UnknownDappComponents+Reducer.swift
```diff
@@ -1,22 +0,0 @@
-struct UnknownDappComponents: FeatureReducer, Sendable {
```
matiasbzurovski, line 1: Reducer deleted since it is now a plain vanilla SwiftUI view, located in `InteractionReview/Destinations/UnknownDappComponents/UnknownDappComponents.swift`

RadixWallet/Features/TransactionReviewFeature/TransactionReviewProofs/TransactionReviewProofs+View.swift
```diff
@@ -1,37 +0,0 @@
-import ComposableArchitecture
-import SwiftUI
-
-// MARK: - TransactionReviewProofs.View
-extension TransactionReviewProofs {
```
matiasbzurovski, line 5: Moved into `InteractionReview/Sections/Proofs/Proofs.swift`

RadixWallet/Features/TransactionReviewFeature/TransactionReviewRawTransaction/TransactionReviewRawTransaction+View.swift
```diff
@@ -1,50 +0,0 @@
-import ComposableArchitecture
-import SwiftUI
-
-// MARK: - TransactionReviewRawTransaction.View
-extension TransactionReviewRawTransaction {
```
matiasbzurovski, line 5: Removed reducer, vanilla SwiftUI view now located under `InteractionReview/Components/RawManifestView.swift`

RadixWallet/Features/PreAuthorizationReview/PreAuthorizationReview.swift
```diff
@@ -0,0 +1,219 @@
...
+				state.displayMode = .detailed
+				return .none
+			}
+
+		case .copyRawManifestButtonTapped:
+			guard let manifest = state.displayMode.rawManifest else {
+				assertionFailure("Copy raw manifest button should only be visible in raw mode")
+				return .none
+			}
+			pasteboardClient.copyString(manifest)
```
GhenadieVP, line 109: This could potentially be part of RawManifestView itself, we either upgrade it to have a reducer, or we could call the client straight from the view.

RadixWallet/Features/PreAuthorizationReview/PreAuthorizationReview+View.swift
```diff
@@ -0,0 +1,268 @@
...
+				guard let offset = positions[navTitleID]?.maxY else {
+					showNavigationTitle = true
+					return
+				}
+				if showNavigationTitle, offset > showTitleHysteresis {
+					showNavigationTitle = false
+				} else if !showNavigationTitle, offset < 0 {
+					showNavigationTitle = true
+				}
+			}
```
GhenadieVP, line 120: Maybe this could be factored out as a view builder along with:
```swift
@SwiftUI.State private var showNavigationTitle = false

		private let coordSpace: String = "PreAuthorizationReviewCoordSpace"
		private let navTitleID: String = "PreAuthorizationReview.title"
		private let showTitleHysteresis: CGFloat = .small3

```

defined at top. Could also add the `header` there.

GhenadieVP (commented):
Looks very good overall 💪, there might be some other places were we could remove the duplications between Transaction and Pre-Auth.

RadixWallet/Core/SharedModels/P2P/Dapp/Models/WalletInteraction+Extensions.swift
```diff
@@ -41,6 +41,20 @@ extension DappToWalletInteractionSendTransactionItem {
 	}
 }
 
+extension DappToWalletInteractionSubintentRequestItem {
+	init(
```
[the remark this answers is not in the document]
matiasbzurovski: it is actually no longer needed, removed

RadixWallet/Features/PreAuthorizationReview/PreAuthorizationReview.swift
matiasbzurovski: The thing is, for `TransactionReview`, we still have the button placed outside the `RawManifestView`, meaning the button tapped action is handled on `TransactionReview`. We could still delegate the call to its child `RawManifestView`. 

I already do something like that for Sections, but I mostly prefer staying away of such pattern, that's why I preferred it this way. Wdyt?

RadixWallet/Features/PreAuthorizationReview/PreAuthorizationReview+View.swift
matiasbzurovski: Attempted this [here](https://github.com/radixdlt/babylon-wallet-ios/pull/1371/commits/0be24a389c9582c057ae25b8a4b4ae1c3c3ea63b), but ended up reverting it due to some issues with the shadow of the top JaggedEdge, as seen below.

Even if it is solved, I am not sure if it is a great approach since the child view (`VisibleHeaderView`), would be setting the `toolbar`, and I've seen some weird behavior in cases like this.

| Before | After |
| -- | -- |
|  |  |

GhenadieVP: sure, fair enough, we don't gain a lot with duplication removal

RadixWallet/Features/PreAuthorizationReview/PreAuthorizationReview.swift
GhenadieVP: Isn't the `Copy` button always inside the RawManifestView?

GhenadieVP (approved):
LGTM, well done!

RadixWallet/Features/PreAuthorizationReview/PreAuthorizationReview+View.swift
```diff
@@ -0,0 +1,268 @@
...
+
+private extension PreAuthorizationReview.View {
+	/// Given an amount of seconds, returns a formatted String using the corresponding unit (days/hours/minutes/seconds).
+	/// A few examples on how should it look for each of them:
+	/// - `8 days` / `1 day`
+	/// - `23:21 hours` / `1:24 hour`
+	/// - `56:02 minutes` / `1:23 minute`
+	/// - `34 seconds` / `1 second`
+	func formatTime(seconds: Int) -> String {
+		typealias S = L10n.PreAuthorizationReview.TimeFormat
```
CyonAlexRDX, line 224: I would probably not have this formatting logic inside a view, can we move it to Sargon and expose a function?

CyonAlexRDX: or just move it from view layer and add some unit tests?

RadixWallet/Features/InteractionReview/Helpers/DisplayMode.swift
```diff
@@ -0,0 +1,13 @@
+import Foundation
+
+extension InteractionReview {
+	enum DisplayMode: Sendable, Hashable {
+		case detailed
+		case raw(String)
```
CyonAlexRDX, line 6: can we use something more structured than `String`? Don't we have some `UnvalidatedManifest` type or something? Maybe we can create a new type or some typealias at least.

I would also put a one line doc on both cases and probably give `raw` a label: `case raw(manifest: ...)`.

RadixWallet/Features/InteractionReview/Sections/Accounts/Accounts.swift
```diff
@@ -0,0 +1,127 @@
+import ComposableArchitecture
+import SwiftUI
+
+// MARK: - InteractionReview.Accounts
+extension InteractionReview {
+	@Reducer
+	struct Accounts: Sendable, FeatureReducer {
```
CyonAlexRDX, line 7: I'd prefer if we could avoid calling reducers the same thing as models, makes searches (or possible find replace all) tricker, can it be called `AccountsSection` or similar? I dont feel too strongly about this though, so feel free to ignore

RadixWallet/Features/DappInteractionFeature/Coordinator/DappInteractionModels.swift
```diff
@@ -131,6 +134,9 @@ extension DappToWalletInteraction {
 			// transactions
 			case .send:
 				0
+			// preAuthorization
```
CyonAlexRDX, line 137: I would probably rename `send` to `submitTransaction` or similar. We now have `subintent` and then having just `send` looks weird. `send` was short for `sendTransasction` - and `sendTransaction` I've always disliked, I think it should be called `submitTransaction` and then we can call `subintent` `signSubintent` perhaps - because we actually do NOT submit the signed subintent, the Dapp does.

CyonAlexRDX (approved):
Bravo, well done! **I've not tested it locally**.

I left some remarks, nothing too important, mostly code cosmetics..

RadixWallet/Features/PreAuthorizationReview/PreAuthorizationReview.swift
matiasbzurovski: Sorry, I had totally misunderstood your comment. Updated

RadixWallet/Features/InteractionReview/Helpers/DisplayMode.swift
matiasbzurovski: We cannot use `UnvalidatedManifest` since in `TransactionReview` we are dealing with a `TransactionManifest` (to which we append the fee & guarantees configurations done at this view), while at `PreAuthroizationReview` we deal with a `SubintentManifest`.

I will add the `raw` label, but I think adding a type around the `String` is an overkill, since this is enum is just internally to display the `RawManifestView`.

RadixWallet/Features/InteractionReview/Sections/Accounts/Accounts.swift
matiasbzurovski: I'd prefer sticking as it is. The find & replace is already pretty bad on Xcode, meaning that even if we name it `AccountsSection`, it will still break if we have other `AccountsSections` anywhere else. When I just renamed the `send` case to `submitTransaction`, Xcode wanted to update tests that were doing `viewStore.send()`.

Then, keeping it short-named and enclosed under `InteractionReview` namespace makes it easier to follow IMO.

RadixWallet/Features/DappInteractionFeature/Coordinator/DappInteractionModels.swift
matiasbzurovski: 100% agree (always disliked `send` as well). Updating

RadixWallet/Features/PreAuthorizationReview/PreAuthorizationReview.swift
GhenadieVP: no worries :).

danvleju-rdx (approved):

PR #1372, opened 2024-10-25
Adjustments for HDPathComponent changes in Sargon

CyonAlexRDX:
# Adjustments of new HDPathComponent in Sargon https://github.com/radixdlt/sargon/pull/243

# Demo
## Import from Olympia 

https://github.com/user-attachments/assets/89c03655-a35f-4c32-a14c-72668fa3ccad

## Account Recovery Scan

https://github.com/user-attachments/assets/ff6b498b-8e21-41e5-b3bc-e2df08e0d2a1

## Create Babylon Accounts and Personas

https://github.com/user-attachments/assets/fc731c7e-fe8b-461e-bba7-8711034cb871


RadixWallet/Clients/FactorSourcesClient/FactorSourcesClient+Live.swift
```diff
@@ -254,8 +254,13 @@ extension FactorSourcesClient: DependencyKey {
 					)
 				).indices
 
-				guard let max = indices.max() else { return 0 }
-				let nextIndex = max + 1
+				guard let max = indices.max() else {
```
matiasbzurovski, line 257: so when implementing MFA, we would extend `nextEntityIndexForFactorSource` to also indicate whether the next index should be on the securified/unsecurified key space, correct?

RadixWallet/Features/DerivePublicKeys/DerivePublicKeys.swift
```diff
@@ -448,19 +448,22 @@ extension DerivePublicKeys {
 			derivationPathScheme: derivationPathScheme,
 			networkID: maybeNetworkID
 		)
+		let hardened = try index.asHardened()
```
matiasbzurovski, line 451: don't we ever derive from unhardened indices?

RadixWallet/Clients/FactorSourcesClient/FactorSourcesClient+Live.swift
Sajjon: We will be using `FactorInstancesProvider` - https://github.com/radixdlt/sargon-mfa/pull/18/ either we will be deriving for Securified half or Unsecurified, correct.

RadixWallet/Features/DerivePublicKeys/DerivePublicKeys.swift
Sajjon: No we don't, since ALL CAP26 (Babylon Account, Babylon Persona, GetID for FactorSourceID) paths MUST harden ALL components due to [SLIP10 requirements](https://github.com/satoshilabs/slips/blame/master/slip-0010.md#L96)

And for BIP44, we use a BIP44 **like** path, yours truely, incorreectly, accidentally hardened the last path component, in Olympia.

(Unfortunately it becomes a little bit more complex, since the Ledger Hardware Devices, on device, chose to correct my mistake, so even if we say HARDEN the last path component of a BIP44 path, those devices unhardened it, so if a user used a Ledger Nano S with Olympia, and then imported the same mnemonic as an Olympia Seed Phrase for a Device FactorSource that user will have problems, but we have not encountered any such user yet.)

GhenadieVP (approved):

PR #1373, opened 2024-10-25
Update SwiftGen config to make resources internal

matiasbzurovski:
## Description
Continuation of https://github.com/radixdlt/babylon-wallet-ios/pull/1370. After migrating to our objects to be `internal`, so should our resources. This PR updates the config so that when running `swiftgen config run`, resources aren't made `public` again.

[SwiftGen](https://github.com/SwiftGen/SwiftGen/blob/6.6.3/Documentation/templates/yaml/inline-swift5.md#customization) only allows to either set `public` or `internal`, meaning we cannot omit the keywoard and have resources be `internal` by default 😞


CyonAlexRDX (approved):
LGTM

PR #1384, opened 2024-11-12
Fix profile access on ProfileStore

GhenadieVP:
Fixed the profile access on Profile store to avoid crashing the app if some parent task was cancelled.

### DIFF
--- a/RadixWallet/Clients/ProfileStore/ProfileStore.swift
+++ b/RadixWallet/Clients/ProfileStore/ProfileStore.swift
@@ -6,7 +6,7 @@ import Sargon
 final actor ProfileStore {
 	static let shared = ProfileStore()
 
-	private let profileSubject: AsyncReplaySubject<Profile> = .init(bufferSize: 1)
+	private let profileSubject: AsyncCurrentValueSubject<Profile?> = .init(nil)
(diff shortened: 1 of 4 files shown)


RadixWallet/Clients/ProfileStore/ProfileStore.swift
```diff
@@ -23,16 +23,11 @@ final actor ProfileStore {
 }
 
 extension ProfileStore {
-	func profile() async -> Profile {
-		try! await profileSubject.first()
```
GhenadieVP, line 27: This implementation would crash the app if parent task was cancelled, as here a force unwrap would be done when a CancellationError is thrown. Instead, `profileSubject` is transformed to `AsyncCurrentValueSubject<Profile?>` to be able to access the current profile.

CyonAlexRDX (approved):
LGTM

matiasbzurovski (approved):

danvleju-rdx (approved):

RadixWallet/Clients/ProfileStore/ProfileStore.swift
GhenadieVP: I didn't do this in the first place exactly for the reasons you describe, as we don't gain anything. Storing in variable might be slightly confusing, as one would do that, if they want to potentially cancel the task at some point.

PR #1387, opened 2024-11-14
Determine fee payer update

GhenadieVP:
When determining fee payer take into account the withdrawn XRD amount for the fee payer candidates from which withdraw is happening.
This is very important in the context of account deletion, when all XRDs will be transferred, so then the Wallet should try to use the account into which the deposit is happening.

# Demo
https://github.com/user-attachments/assets/bd430119-d924-4221-ba5e-a4dc2a2a6e72


danvleju-rdx (approved):

CyonAlexRDX (approved):
LGTM

PR #1388, opened 2024-11-14
Update GH actions to use macos 15

GhenadieVP:
GH removed Xcode 16 from Macos-14 images because policy -> https://github.com/actions/runner-images/issues/10703.


CyonAlexRDX (approved):
LGTM, but I think we should change Sargon to also use `macos-15-xlarge` so we use the **same** runner, WDYT?

PR #1389, opened 2024-11-14
[ABW-3924] Delete Account Flow

danvleju-rdx:
Jira ticket: [ABW-3924](https://radixdlt.atlassian.net/browse/ABW-3924)

## Description
- Adds account deletion flow

## Video

https://github.com/user-attachments/assets/4bedc6b8-0778-46a8-8932-e9baf13e8b07

[ABW-3924]: https://radixdlt.atlassian.net/browse/ABW-3924?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


matiasbzurovski (changes requested):
Looking good! added minor comments mostly

RadixWallet/Features/AccountPreferencesFeature/Children/DeleteAccount/ChooseReceivingAccountOnDelete+Reducer.swift
```diff
@@ -0,0 +1,134 @@
...
+	@CasePathable
+	enum DelegateAction: Sendable, Equatable {
+		case finished(AccountAddress?)
+	}
+
+	struct Destination: DestinationReducer {
+		@CasePathable
+		enum State: Sendable, Hashable {
+			case confirmSkipAlert(AlertState<Action.ConfirmSkipAlert>)
+			case tooManyAssetsAlert(AlertState<Action.TooManyAssetsAlert>)
```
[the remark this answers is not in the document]
danvleju-rdx: Yes, but I expect the `createDeleteAccountManifest` function from Sargon to perform the check and throw a specific error that hosts can handle. Would that be possible?

RadixWallet/Features/ChooseAccounts/ChooseAccounts+Reducer.swift
```diff
@@ -65,22 +65,24 @@ struct ChooseAccounts: Sendable, FeatureReducer {
...
 	@Dependency(\.errorQueue) var errorQueue
 	@Dependency(\.accountsClient) var accountsClient
+	@Dependency(\.onLedgerEntitiesClient) var onLedgerEntitiesClient
 
 	init() {}
 
 	var body: some ReducerOf<Self> {
 		Reduce(core)
-			.ifLet(destinationPath, action: /Action.destination) {
+			.ifLet(\.$destination, action: \.destination) {
```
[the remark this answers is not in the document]
danvleju-rdx: Right. I think we were using the old scope for the destination for the same reason, no?
```Swift
    var destination: PresentationStoreOf<ChooseReceivingAccountOnDelete.Destination> {
        func scopeState(state: State) -> PresentationState<ChooseReceivingAccountOnDelete.Destination.State> {
            state.$destination
        }
        return scope(state: scopeState, action: Action.destination)
    }
```

Haven’t noticed any issues either, but yeah, let’s keep an eye on it.

RadixWallet/Features/AccountPreferencesFeature/Children/DeleteAccount/ChooseReceivingAccountOnDelete+Reducer.swift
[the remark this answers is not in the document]
matiasbzurovski: yes, it is already done [here](https://github.com/radixdlt/sargon/pull/259/files#diff-97b4d7f3bfab301b14d8e78a84347912af6e09704cd36fe48c5f8f7ff89954c6R48-R53). However, the manifest won't be created here (but on `DeleteAccountConfirmationReducer`) so maybe the alert should be there?

Also, this triggers another comment: wouldn't it help to have a `DeleteAccountCoordinator` that is in charge of presenting/pushing views/alerts, as well as executing the logic to actually delete the account?

RadixWallet/Features/ChooseAccounts/ChooseAccounts+Reducer.swift
```diff
@@ -112,17 +116,41 @@ struct ChooseAccounts: Sendable, FeatureReducer {
...
+						let xrdBalance = entity.fungibleResources.xrdResource?.amount.exactAmount?.nominalAmount ?? 0
+						let hasEnoughXRD = xrdBalance >= 1
+
+						return .init(account: account, hasEnoughXRD: hasEnoughXRD)
+					}
+					return receivingAccounts
+						.sorted { $0.hasEnoughXRD && !$1.hasEnoughXRD }
+						.map { State.AccountType.receiving($0) }
+						.asIdentified()
+				}
```
GhenadieVP, line 153: This logic is too specific for the account delete usecase, I think it should be in ChooseReceivingAccountOnDelete and not here. To facilitate this I would allow injecting the accounts loader in ChoseAccounts+Reducer, that would allow having custom loading logic depending on the context where it is used.

RadixWallet/Features/InteractionReview/Sections/Sections+ExecutionSummary.swift
```diff
@@ -41,7 +45,27 @@ extension InteractionReview.Sections {
 
 		switch summary.detailedManifestClass {
 		case nil:
-			return nil
+			guard interactionId?.isWalletAccountDeleteInteraction == true else { return nil }
```
GhenadieVP: this check can go away, and just use the manifest classification that will be merged in Sargon.

RadixWallet/Features/TransactionReviewFeature/TransactionReview.swift
```diff
@@ -29,6 +28,14 @@ struct TransactionReview: Sendable, FeatureReducer {
 		var canApproveTX: Bool = true
 		var sliderResetDate: Date = .now
 
+		var waitsForTransactionToBeComitted: Bool {
+			interactionId.isWalletAccountDepositSettingsInteraction
```
GhenadieVP, line 32: also for delete accounts, users have to wait until the tx is comitted.

RadixWallet/Features/AccountPreferencesFeature/Children/DeleteAccount/ChooseReceivingAccountOnDelete+Reducer.swift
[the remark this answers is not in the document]
danvleju-rdx: I've added `DeleteAccountCoordinator`, which handles the delete account logic. I didn’t include a navigation stack since it made things unnecessarily complicated. Instead, I kept only `destination`, so `tooManyAssetsAlert` is presented by `ChooseReceivingAccountOnDelete`.

RadixWallet/Features/ChooseAccounts/ChooseAccounts+Reducer.swift
```diff
@@ -134,5 +162,39 @@ extension ChooseAccounts.State {
...
 		case assetTransfer
 		case permission(SelectionRequirement)
+		case accountDeletion
+	}
+
+	enum AccountType: Sendable, Hashable, Identifiable {
+		case general(Account)
+		case receiving(ReceivingAccountCandidate)
+
+		var id: Self { self }
```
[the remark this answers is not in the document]
danvleju-rdx: Removed `AccountType`, so not relevant anymore

RadixWallet/Features/InteractionReview/Sections/Sections+ExecutionSummary.swift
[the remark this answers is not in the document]
danvleju-rdx: With the latest updates, we now use the manifest classification and no longer pass interactionId. So it makes sense to treat this like any other section IMHO

RadixWallet/Features/AccountPreferencesFeature/Children/DeleteAccount/DeleteAccountConfirmation+Reducer.swift
```diff
@@ -0,0 +1,105 @@
...
+	}
+}
+
+// MARK: - DeleteAccountConfirmation.State.ReceivingAccountCandidate
+extension DeleteAccountConfirmation.State {
+	struct ReceivingAccountCandidate: Sendable, Hashable {
+		let account: Account
+		let hasEnoughXRD: Bool
+
+		static let minimumRequiredXRD: Decimal192 = 4
```
Sajjon, line 103: This literal seems a bit arbitrary?

Should we let it be a constant declared in Sargon and reused between hosts?

RadixWallet/Core/SharedModels/Assets/OnLedgerEntity.swift
```diff
@@ -673,6 +673,15 @@ extension OnLedgerEntity.OnLedgerAccount {
 	func hasResource(_ resourceAddress: ResourceAddress) -> Bool {
 		allResourceAddresses.contains(resourceAddress)
 	}
+
+	var containsAnyAsset: Bool {
+		let account = nonEmptyVaults
+		return account.fungibleResources.xrdResource != nil
+			|| !account.fungibleResources.nonXrdResources.isEmpty
+			|| !account.nonFungibleResources.isEmpty
+			|| !account.poolUnitResources.poolUnits.isEmpty
```
Sajjon, line 682: Is redeemable LSUs included here? Im bad at LSU logic ...

RadixWallet/Clients/AccountPortfoliosClient/AccountPortfoliosClient+Live.swift
```diff
@@ -177,15 +178,27 @@ extension AccountPortfoliosClient: DependencyKey {
...
+					overlayWindowClient.scheduleAlertAndIgnoreAction(.deletedAccountDetectedAlert)
+				}
+			}
+		}
+
 		return AccountPortfoliosClient(
 			fetchAccountPortfolios: { accountAddresses, forceRefresh in
 				try await Task.detached {
-					try await fetchAccountPortfolios(accountAddresses, forceRefreshEntities: forceRefresh, forceRefreshPrices: forceRefresh)
+					syncAccountsDeletedOnLedger()
```
CyonAlexRDX, line 194: `syncAccountsDeletedOnLedger` is being called from this Task which already detached, and then `syncAccountsDeletedOnLedger` itself uses `Task.detached`, hmm, do we do that because we **want** `fetchAccountPortfolios` (called below) to be performed in parallell? because that is what will happen right? `fetchAccountPortfolios` can be called - and finished - before `syncAccountsDeletedOnLedger`  is done. Is that the intention?

Otherwise since `syncAccountsDeletedOnLedger` is **only** called in two already detached and async contexts I think maybe it (`syncAccountsDeletedOnLedger`) can be marked async? and removed its Task.detached.

I'm not completely sure though. I just to explore the possibility of this working "by chance" and see if we should try to use more structured async. 

We have the `State` value, on which we can have `Task<..>?` as well and cancel any previous task and replace with new ones. 

e.g.

```diff 
actor State {
...
+ var deletedAccountsSync: Tas<...>?
...
}
```

```diff
static let liveValue: AccountPortfoliosClient {
    let state = State()
...
    @Sendable func syncAccountsDeletedOnLedger() {
+        state.deletedAccountsSync.cancel()
+        state.deletedAccountsSync = nil
+        state,deletedAccountsSync = Task { ...}
    }
...
}
```

RadixWallet/Features/AccountPreferencesFeature/Children/DeleteAccount/DeleteAccountCoordinator+Reducer.swift
```diff
@@ -0,0 +1,215 @@
...
+					assertionFailure("Not a transaction Response?")
+				case .dapp(.failure), .none:
+					/// Either user did dismiss the TransctionReview, or there was a failure.
+					/// Any failure message will be displayed in Transaction Review
+					await send(.internal(.accountDeletionFailed))
+				}
+			} catch {
+				switch error as? CommonError {
+				case .MaxTransfersPerTransactionReached:
+					await send(.internal(.accountDeletionFailedDueTooManyAssets))
```
CyonAlexRDX, line 207: future improvement: have a function in sargon:

```rust
impl CommonError {
    fn account_deletion_failed_because_of_too_many_assets_owned(&self) -> bool { ... }
}
``

Comment (CyonAlexRDX) [reply RadixWallet/Features/AccountPreferencesFeature/Children/DeleteAccount/DeleteAccountCoordinator+Reducer.swift:207]: and then in Swift Sargon we can:

```swift
extension Swift.Error {
   var accountDeletionFailedBecauseOfTooManyAssetsOwned: Bool {
       guard let commonError = self as? CommonError else { return false }
       return commonError.accountDeletionFailedBecauseOfTooManyAssetsOwned() // calling rust fn
   }
}
```

RadixWallet/Features/InteractionReview/Sections/Accounts/Accounts.swift
```diff
@@ -68,6 +68,7 @@ extension InteractionReview {
 			let account: InteractionReview.ReviewAccount
 			var transfers: IdentifiedArrayOf<InteractionReview.Transfer>
 			let isDeposit: Bool
+			var isAccountDelete: Bool = false
```
CyonAlexRDX, line 71: `isAccountDelete` sounds wrong, perhaps `isDeletingAccount` or `forDeletionOfAccount`?

CyonAlexRDX: wait can `isDeposit` and `isAccountDeletion` be **both** `true`? else I suggest you introduce an enum a `Purpose`?

danvleju-rdx: Right, will add an enum

RadixWallet/MIGRATE_TO_SARGON/Stage1/Stage1MigrateToSargon+Account.swift
```diff
@@ -42,15 +42,19 @@ extension Account {
 	}
 
 	mutating func hide() {
-		flags.append(.deletedByUser)
+		flags.append(.hiddenByUser)
 	}
 
 	mutating func unhide() {
-		entityFlags.remove(.deletedByUser)
+		entityFlags.remove(.hiddenByUser)
```
CyonAlexRDX, line 49: Notice folder path :) `MIGRATE_TO_SARGON` so perhaps this can be removed altogether and moved to Swift Sargon and probably using Sargon methods/functions.

Would be great with shared functionality with Android Host. I.e. should we simply remove `hiddenByUser` flag? What if one calls this method on an account has been deleted - should not be possibile in UI - but should we throw? Crash? Or do nothing? Would be great to have made that decision once - in Sargon - and  reuse for hosts.

CyonAlexRDX (changes requested):
Nice! Almost there, left some minor remarks

RadixWallet/Clients/AccountPortfoliosClient/AccountPortfoliosClient+Live.swift
danvleju-rdx: Yes, the intention was to perform them in parallel, but it’s not really necessary. I can make `syncAccountsDeletedOnLedger` async and remove Task

RadixWallet/MIGRATE_TO_SARGON/Stage1/Stage1MigrateToSargon+Account.swift
GhenadieVP: There is mark_as_hidden in Sargon which could be used, but I think needs to be done separately and not part of this PR, when we will start migrating more logic from the Wallet to Sargon.

GhenadieVP (approved):

RadixWallet/Features/AccountPreferencesFeature/Children/DeleteAccount/DeleteAccountCoordinator+View.swift
```diff
@@ -0,0 +1,48 @@
...
+
+// MARK: - DeleteAccountCoordinator.View
+extension DeleteAccountCoordinator {
+	@MainActor
+	struct View: SwiftUI.View {
+		private let store: StoreOf<DeleteAccountCoordinator>
+
+		init(store: StoreOf<DeleteAccountCoordinator>) {
+			self.store = store
+		}
```
matiasbzurovski, line 12:
```suggestion
		let store: StoreOf<DeleteAccountCoordinator>
```

matiasbzurovski (changes requested):
Only minor things in code. 
Asides that, two things I've found:
1) Seems we are always showing the alert after deleting account, is that expected?

Screenshot 2024-11-21 at 13 24 32



2) Let's remove this _fake_ `Delete Account` button on `Dev Preferences` now that we have `real` feature on Account Preferences

GhenadieVP:
> Seems we are always showing the alert after deleting account, is that expected?

That is a bug, needs fixing

danvleju-rdx:
Fixed the bug mentioned above. The issue occurred because account portfolios were being fetched immediately after a transaction was successfully committed, but before the account was marked as tombstoned, which triggered the alert incorrectly. The solution was to call `syncAccountsDeletedOnLedger` only where it is needed - specifically in `Home` and `AccountDetails` - to ensure better control over when the alert is displayed.

@matiasbzurovski Also removed the delete button from the `Dev Preferences`.

RadixWallet/Clients/AccountPortfoliosClient/AccountPortfoliosClient+Live.swift
```diff
@@ -196,7 +197,13 @@ extension AccountPortfoliosClient: DependencyKey {
 			portfolioForAccount: { address in
 				await state.portfolioForAccount(address)
 			},
-			portfolios: { state.portfoliosSubject.value.wrappedValue.map { Array($0.values) } ?? [] }
+			portfolios: { state.portfoliosSubject.value.wrappedValue.map { Array($0.values) } ?? [] },
+			syncAccountsDeletedOnLedger: {
+				let isDeletedAccountDetected = try? await SargonOS.shared.syncAccountsDeletedOnLedger()
```
CyonAlexRDX, line 202: 🚀

RadixWallet/Features/AccountPreferencesFeature/Children/DeleteAccount/DeleteAccountConfirmation+Reducer.swift
CyonAlexRDX: I think we can use the constant now right?

CyonAlexRDX (approved):
Good job!

RadixWallet/Features/AccountPreferencesFeature/Children/DeleteAccount/DeleteAccountConfirmation+Reducer.swift
danvleju-rdx: No, because updating Sargon will also bring the proof of ownership updates, but I'll make sure to make this change in the proof of ownership PR.

RadixWallet/Clients/AccountsClient/AccountsClient+Interface.swift
```diff
@@ -31,70 +31,6 @@ struct AccountsClient: Sendable {
 	var hasAccountOnNetwork: HasAccountOnNetwork
 
 	var updateAccount: UpdateAccount
-
```
matiasbzurovski, line 34: ❤️

matiasbzurovski (approved):
Nice work!

PR #1400, opened 2024-11-28
[ABW-4001] Update PreAuthorization models

matiasbzurovski:
Jira ticket: [ABW-4001](https://radixdlt.atlassian.net/browse/ABW-4001)

## Description
Updates response to include `SubintentHash` and `expirationTimestamp`, which allows us to remove the `PreAuthData` helper struct we moved along `DappInteractor` (and children).

## How to test

Play with Anthic test app and verify everything works as expected.

[ABW-4001]: https://radixdlt.atlassian.net/browse/ABW-4001?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/RadixWallet/Clients/PreAuthorizationClient/PreAuthorizationClient+Interface.swift
+++ b/RadixWallet/Clients/PreAuthorizationClient/PreAuthorizationClient+Interface.swift
@@ -29,6 +29,6 @@ extension PreAuthorizationClient {
 
 	struct PollStatusRequest: Sendable {
 		let subintentHash: SubintentHash
-		let expirationTimestamp: Date
+		let expirationTimestamp: Instant
 	}
 }
--- a/RadixWallet/Clients/PreAuthorizationClient/PreAuthorizationClient+Interface.swift
+++ b/RadixWallet/Clients/PreAuthorizationClient/PreAuthorizationClient+Interface.swift
@@ -29,6 +29,6 @@ extension PreAuthorizationClient {
 
 	struct PollStatusRequest: Sendable {
 		let subintentHash: SubintentHash
-		let expiration: DappToWalletInteractionSubintentExpiration
+		let expirationTimestamp: Date
 	}
 }
--- a/RadixWallet/Features/DappInteractionFeature/Coordinator/DappInteractionFlow.swift
+++ b/RadixWallet/Features/DappInteractionFeature/Coordinator/DappInteractionFlow.swift
@@ -512,10 +512,9 @@ extension DappInteractionFlow {
 
 		func handlePreAuthorizationSignature(
 			_ item: State.AnyInteractionItem,
-			_ signedSubintent: SignedSubintent,
-			_ expiration: DappToWalletInteractionSubintentExpiration
+			_ signedSubintent: SignedSubintent
 		) -> Effect<Action> {
-			let preAuthResponse = newWalletToDappInteractionPreAuthorizationResponseItems(signedSubintent: signedSubintent, expirationTimestamp: expiration.timestamp)
+			let preAuthResponse = newWalletToDappInteractionPreAuthorizationResponseItems(signedSubintent: signedSubintent)
 			state.responseItems[item] = .remote(.preAuthorization(preAuthResponse))
 
 			return continueEffect(for: &state)
@@ -588,8 +587,8 @@ extension DappInteractionFlow {
 		     .accountsProofOfOwnership(.delegate(.failedToSign)):
 			return dismissEffect(for: state, errorKind: .failedToSignAuthChallenge, message: nil)
 
-		case let .preAuthorizationReview(.delegate(.signedPreAuthorization(encoded, expiration))):
-			return handlePreAuthorizationSignature(item, encoded, expiration)
+		case let .preAuthorizationReview(.delegate(.signedPreAuthorization(encoded))):
+			return handlePreAuthorizationSignature(item, encoded)
 
 		case let .preAuthorizationReview(.delegate(.failed(error))):
 			return handlePreAuthorizationFailure(error)
--- a/RadixWallet/Features/DappInteractionFeature/Coordinator/DappInteractionFlow.swift
+++ b/RadixWallet/Features/DappInteractionFeature/Coordinator/DappInteractionFlow.swift
@@ -1155,14 +1155,3 @@ struct SavedPersonaDataInPersonaDoesNotMatchWalletInteractionResponseItem: Swift
 struct PersonaDataEntryNotFoundInResponse: Swift.Error {
 	let kind: PersonaData.Entry.Kind
 }
-
-private extension DappToWalletInteractionSubintentExpiration {
-	var timestamp: Date {
-		switch self {
-		case let .afterDelay(afterDelay):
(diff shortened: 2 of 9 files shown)


RadixWallet/Features/DappInteractionFeature/Coordinator/DappInteractionFlow.swift
```diff
@@ -515,10 +515,10 @@ extension DappInteractionFlow {
 			_ signedSubintent: SignedSubintent,
 			_ expiration: DappToWalletInteractionSubintentExpiration
 		) -> Effect<Action> {
-			let preAuthResponse = newWalletToDappInteractionPreAuthorizationResponseItems(signedSubintent: signedSubintent)
+			let preAuthResponse = newWalletToDappInteractionPreAuthorizationResponseItems(signedSubintent: signedSubintent, expirationTimestamp: expiration.timestamp)
```
GhenadieVP, line 518: signedSubintent contains the expiration timestamp already, no?

RadixWallet/Features/DappInteractionFeature/Coordinator/DappInteractionFlow.swift
```diff
@@ -1156,3 +1156,14 @@ struct SavedPersonaDataInPersonaDoesNotMatchWalletInteractionResponseItem: Swift
 struct PersonaDataEntryNotFoundInResponse: Swift.Error {
 	let kind: PersonaData.Entry.Kind
 }
+
+private extension DappToWalletInteractionSubintentExpiration {
+	var timestamp: Date {
+		switch self {
+		case let .afterDelay(afterDelay):
+			Date().addingTimeInterval(TimeInterval(afterDelay.expireAfterSeconds))
```
GhenadieVP, line 1164: `Date()`? Shouldn't the timestamp be UTC?

RadixWallet/Features/DappInteractionFeature/Coordinator/DappInteractionFlow.swift
matiasbzurovski: no, only the request contains the expiration timestamp (which we have now included in the response model as well).

RadixWallet/Features/DappInteractionFeature/Coordinator/DappInteractionFlow.swift
matiasbzurovski: oh dearful dates and timezones.. Let me write down my dummy/cautious reasoning to validate the current logic makes sense.

In this case, the subintent will expire X seconds after signature. Let's use `600` (10 minutes) for this example, which we will set using the [test dApp](https://d1vuj67scmz8qj.cloudfront.net/).

It is currently `13:54 PM GMT+1` in Madrid. This means that when I get to this state, the `timestamp: Date` will return `14:04 GMT+1`. 

Given this `timestamp` is used on the `PollPreAuthorizationStatus+View`, we can validate that after signing the preAuth, the view says that it expires in `9 minutes` (it shows 10 minutes only for a second) and not in `-50 minutes` (as it would be in `GMT` timezone)





So far so good, now we need to make sure that the value sent to Sargon and then to the dApp makes sense. The process here is: 
- convert `Swift.Date` into `Sargon.Timestamp` (happens thanks to magic code in `uniffi.toml`)
- get amount of second since this timestamp and send it to dApp (done [here](https://github.com/radixdlt/sargon/pull/279/files#diff-090b4502eb51dbc6bffb6645ec282687461147f1e0fc282cfb1e1d4ddb5a3d71R33-R38))

Now, if we check the final result sent to the dApp we can see it is `1732799073`.
Screenshot 2024-11-28 at 13 58 36


Then, using a Playground I can confirm this value corresponds to `14:04` in my timezone:
Screenshot 2024-11-28 at 14 00 09

Or if I use a helper tool such as https://www.unixtimestamp.com/ we can verify it is `14:04 GMT+1` and `13:04 GMT`:

RadixWallet/Features/DappInteractionFeature/Coordinator/DappInteractionFlow.swift
GhenadieVP: the response contains the calculated expiration timestamp, why not use?

RadixWallet/Features/DappInteractionFeature/Coordinator/DappInteractionFlow.swift
GhenadieVP: given that the same delay applies, regardless of the base, I think this could be safe, unless user goes and changes the date in the settings while the countdown is ticking 😄.

RadixWallet/Features/DappInteractionFeature/Coordinator/DappInteractionFlow.swift
matiasbzurovski: Updated here and in https://github.com/radixdlt/sargon/pull/282

RadixWallet/Features/DappInteractionFeature/Coordinator/DappInteractionFlow.swift
```diff
@@ -1156,3 +1155,14 @@ struct SavedPersonaDataInPersonaDoesNotMatchWalletInteractionResponseItem: Swift
 struct PersonaDataEntryNotFoundInResponse: Swift.Error {
 	let kind: PersonaData.Entry.Kind
 }
+
+private extension DappToWalletInteractionSubintentExpiration {
```
GhenadieVP, line 1159: I guess this can go away now.

RadixWallet/Clients/PreAuthorizationClient/PreAuthorizationClient+Interface.swift
```diff
@@ -29,6 +29,6 @@ extension PreAuthorizationClient {
 
 	struct PollStatusRequest: Sendable {
 		let subintentHash: SubintentHash
-		let expiration: DappToWalletInteractionSubintentExpiration
+		let expirationTimestamp: Instant
 	}
```
GhenadieVP, line 33: I will not push for more optimizations, we can do it later, but this should not be needed. 
The WalletToDappInteractionSubintentResponseItem contains the full Subintent, which can then be sent to Sargon and it would extract the proper timestamp and hash. This API sort of allows having an unrelated timestamp to a given subintent hash.

GhenadieVP (approved):

RadixWallet/Clients/PreAuthorizationClient/PreAuthorizationClient+Interface.swift
matiasbzurovski: Yes, in fact sending the entire `Subintent` rather than the `SubintentHash` and `Instant` is what I proposed on SU yesterday.

However, we discussed it further with @sergiupuhalschi-rdx and agreed to keep it this way to avoid sending unnecessary information (the more data we send over UniFFI the slowest right?) and making more updates on both sides. We didn't consider the fact that this allows having an unrelated timestamp, though

PR #1401, opened 2024-11-28
Fix crash caused by early profile access

GhenadieVP:
https://radixdlt.atlassian.net/browse/ABW-4004

Fix the places that did access the profile when it was not present yet:
- RadixConnectClient notifying CE with updated right after a profile is selected for import.
- Account recovery scan from the onboarding trying to get appearance id from the Profile

### DIFF
--- a/RadixWallet/Clients/RadixConnectClient/RadixConnectClient+Live.swift
+++ b/RadixWallet/Clients/RadixConnectClient/RadixConnectClient+Live.swift
@@ -14,25 +14,6 @@ extension RadixConnectClient {
 		let radixConnectMobile = RadixConnectMobile()
 		let localNetworkAuthorization = LocalNetworkAuthorization()
 
-		Task {
-			@Dependency(\.accountsClient) var accountsClient
-			for try await accounts in await accountsClient.accountsOnCurrentNetwork() {
-				guard !Task.isCancelled else { return }
-				try? await sendAccountListMessage(accounts: accounts)
-			}
-		}
-
-		Task {
-			let connectedClients = await rtcClients.connectClients()
-				.filter { updates in
-					!updates.flatMap(\.idsOfConnectedPeerConnections).isEmpty
-				}
-			for try await updates in connectedClients {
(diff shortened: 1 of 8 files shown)


RadixWallet/Features/AccountRecoveryScan/Children/AccountRecoveryScanInProgress/AccountRecoveryScanInProgress.swift
```diff
@@ -256,6 +259,20 @@ struct AccountRecoveryScanInProgress: Sendable, FeatureReducer {
...
 	}
+
+	private func getAccountAppearanceID(
+		mode: State.Mode,
+		offset: Int,
+		networkID: NetworkID
+	) async -> AppearanceID {
+		switch mode {
+		case .createProfile:
+			return AppearanceID.fromNumberOfAccounts(offset)
```
GhenadieVP, line 270: @CyonAlexRDX is this assumption correct? Seems to be, but not 100% sure.

RadixWallet/Clients/RadixConnectClient/RadixConnectClient+Live.swift
```diff
@@ -14,25 +14,6 @@ extension RadixConnectClient {
 		let radixConnectMobile = RadixConnectMobile()
 		let localNetworkAuthorization = LocalNetworkAuthorization()
 
-		Task {
```
GhenadieVP, line 17: Don't start the tasks as soon as the Client is instantiated, instead these will be started on main screen, after the profile is available.

RadixWallet/Features/AccountRecoveryScan/Children/AccountRecoveryScanInProgress/AccountRecoveryScanInProgress.swift
CyonAlexRDX: I think that looks good

CyonAlexRDX (approved):
LGTM

PR #1403, opened 2024-12-02
Fix import encrypted profile hang

GhenadieVP:
The Wallet did hang indefinitely when importing an encrypted profile. 

Could not find a strong reason as to why the bug is happening. Seems to be caused by SwiftUI issue of cancelling the view task, in this case ImportMnemonicsFlowCoordinator's view, when the previews view is not yet fully dismissed.

Add a small delay fixes the issue.

## Bug

https://github.com/user-attachments/assets/a72b3100-12f6-4096-959f-4bba13022049


CyonAlexRDX (approved):
🥲 TCA bah!

matiasbzurovski (approved):

PR #1411, opened 2024-12-09
[ABW-3970] Shield Setup Select Factors

danvleju-rdx:
Jira ticket: [ABW-3970](https://radixdlt.atlassian.net/browse/ABW-3970)

## Description
Adds SelectFactorSources flow

## Video
[Video](https://drive.google.com/file/d/18kJWmZgrd6DG13GARPRGo8PHWDhsOMxs/view?usp=drive_link)

[ABW-3970]: https://radixdlt.atlassian.net/browse/ABW-3970?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


RadixWallet/Features/ShieldSetup/SelectFactorSources/Children/SelectFactorSources+View.swift
```diff
@@ -0,0 +1,134 @@
+import ComposableArchitecture
+import SwiftUI
+
+// TODO: move to Sargon
```
Sajjon, line 4: 👍 I really like that you write `// TODO: move to Sargon` like this :) because otherwise you know I will come and review "please move to sargon" :)

RadixWallet/Features/ShieldSetup/SelectFactorSources/Children/SelectFactorSources+View.swift
```diff
@@ -0,0 +1,134 @@
...
+		case .password: 3
+		case .offDeviceMnemonic: 4
+		case .trustedContact: 5
+		case .securityQuestions: 6
+		}
+	}
+}
+
+extension SelectFactorSources.State {
+	var sortedFactorSources: [FactorSource] {
```
Sajjon, line 20: with `displayOrder` in sargon (as per your comment) we will have the main logic of this sorting in Sargon, which is great. But I wonder if we should not try to take it any further! This is kind of still having logic in hosts. I think better if Sargon provided this **whole** `Reducer.State` with its (view)state! i.e. (pseudocode)

```swift
struct ParentReduer {
    func reduce(internalAction: InternalAction) {
       switch internalAction {
          case .selectFactorSourcePressed:
            let sargonPreparedChildState = SargonOS.viewModel(for: .selectFactorSources)
            let childReducerState = SelectFactorSources(state: sargonPreparedChildState)
             state.destination = .selectFactorSources(state: childReducerState)
      }
    }
}
```

Sajjon: merits a bigger decision and Ghenadies presence. But I think we should be really vary of putting **any** logic whatsoever in hosts. And this sorting has a liiitle bit of implicit logic. Since it "decides" that it should use `<` and `$0.factorSourceKind.displayOrder` - which is much more logic than `SargonOS.sortFactorsForSelection(factors: factors)` or as above, letting Sargon build the "ViewModel" (state in `Reducer.State` I guess).

RadixWallet/Features/SecurityCenterFeature/SecurityCenter+View.swift
```diff
@@ -36,7 +36,8 @@ extension SecurityCenter {
 								}
 							}
 
-							ForEach(SecurityProblemKind.allCases, id: \.self) { type in
+							// TODO: fix Sargon
```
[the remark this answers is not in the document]
danvleju-rdx: missing case in Swift Sargon:
```Swift
extension SecurityProblemKind: CaseIterable {
	public static var allCases: [SecurityProblemKind] {
		[.configurationBackup, .securityFactors]
	}
}
```

RadixWallet/Features/ShieldSetup/SelectFactorSources/Children/SelectFactorSources+View.swift
danvleju-rdx: `SargonOS.sortFactorsForSelection(factors: factors)` seems fair enough for now

Sajjon: @danvleju-rdx I ofc meant `FactorSources` not `Factors` (since `Factors` is forbidden! I was just lazy...) 
probably should not live in `SargonOS` namespace. Probably a method on `SecurityShieldBuilder`
```swift
SecurityShieldBuilder().sortFactorsSourcesForSelection(factors)
```


```rust
#[uniffi::export]
impl SecurityShieldBuilder {
    fn sort_factor_sources_for_selection(&self, factor_sourcess: [FactorSources]) ->  { .. }
}
```

That wat you have a more suitable namespace! no need to polute SargonOS with it. and also that way you dont need to have.

Or see if you can `Uniffi` export a ("static") function in UniFFI:

```rust
#[uniffi::export]
impl SecurityShieldBuilder {
// Note lacke of `&self`....
    fn sort_factor_sources_for_selection(factor_sourcess: [FactorSources]) ->  { .. }
}
```

Im not sure if that works. If so you have two solutions:
1. uniffi::export a global freestanding function
2. use the method above, but you might do an `extension SecurityShieldBuilder` in **Swift** Sargon and `public static func` it and use it on a new instance of `SecurityShieldBuilder`. I think making it a `static func` under `SecurityShieldBuilder` is what we want, to make it clear that it is **not a method which mutates the internal state of the `SecurityShieldBuilder`**! Rather just a utility function which we put under the `SecurityShieldBuilder`'s namespace!

RadixWallet/Features/ShieldSetup/SelectFactorSources/Children/SelectFactorSources+View.swift
```diff
@@ -0,0 +1,134 @@
...
+		factorSources.sorted(by: {
+			$0.factorSourceKind.displayOrder < $1.factorSourceKind.displayOrder
+		})
+	}
+
+	var statusMessageInfo: (type: StatusMessageView.ViewType, text: String)? {
+		if (selectedFactorSources ?? []).isEmpty {
```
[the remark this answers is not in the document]
danvleju-rdx: Do we need to involve securityShieldBuilder at this point? We could use `SargonOS.selectedFactorsStatus(factors: factors)` instead

RadixWallet/Features/ShieldSetup/SelectFactorSources/Children/SelectFactorSources+View.swift
matiasbzurovski: agree with the final conclusions: a static exported method on `SecurityShieldBuilder` seems the best choice for now

RadixWallet/Features/SecurityCenterFeature/SecurityCenter+View.swift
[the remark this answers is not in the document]
matiasbzurovski: oh, sorry for missing that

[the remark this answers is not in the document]
danvleju-rdx: No problem, I'll take care of it

RadixWallet/Features/ShieldSetup/SelectFactorSources/Children/SelectFactorSources+Reducer.swift
```diff
@@ -0,0 +1,113 @@
...
+						builder = builder.addFactorSourceToPrimaryThreshold(factorSourceId: factorSource.factorSourceID)
+					}
+				}
+			}
+
+			state.$shieldBuilder.withLock { builder in
+				builder = builder.setThreshold(threshold: UInt8(builder.primaryRoleThresholdFactors.count))
+			}
+
+			if !state.didInteractWithSelection, factorSources != nil {
```
GhenadieVP, line 80: very minor, but probably `!difference.isEmpty()` instead of `factorSources != nil` makes more sense.

RadixWallet/Features/ShieldSetup/SelectFactorSources/Children/SelectFactorSources+Reducer.swift
```diff
@@ -0,0 +1,113 @@
+import ComposableArchitecture
+
+// MARK: - SelectFactorSources
+@Reducer
+struct SelectFactorSources: FeatureReducer, Sendable {
+	@ObservableState
+	struct State: Hashable, Sendable {
+		@Shared(.shieldBuilder) var shieldBuilder
+		var factorSources: [FactorSource] = []
+		var selectedFactorSources: [FactorSource]? {
```
GhenadieVP, line 10: Why optional?

RadixWallet/Features/ShieldSetup/SelectFactorSources/Children/SelectFactorSources+Reducer.swift
```diff
@@ -0,0 +1,113 @@
...
+			factorSources.filter { shieldBuilder.primaryRoleThresholdFactors.contains($0.factorSourceID) }
+		}
+
+		var didInteractWithSelection = false
+
+		init() {
+			$shieldBuilder.withLock { sharedValue in
```
GhenadieVP, line 17: Is it possible to have this?
```swift
extension SharedKey where Self == InMemoryKey<SecurityShieldBuilder>.Default {
	static var shieldBuilder: Self {
		Self[.inMemory("shieldBuilder"), default: SecurityShieldBuilder()]
	}
	
	func initialize() {
	   withLock { sharedValue in 
	       sharedValue = SecurityShieldBuilder()
	   }
	}
}
```

Asking because we might need to initialize the state in some other context - for example when editing a shield.

RadixWallet/Features/ShieldSetup/SelectFactorSources/Children/SelectFactorSources+Reducer.swift
```diff
@@ -0,0 +1,113 @@
...
+				switch change {
+				case let .remove(_, factorSource, _):
+					state.$shieldBuilder.withLock { builder in
+						builder = builder.removeFactorFromPrimary(factorSourceId: factorSource.factorSourceID)
+					}
+				case let .insert(_, factorSource, _):
+					state.$shieldBuilder.withLock { builder in
+						builder = builder.addFactorSourceToPrimaryThreshold(factorSourceId: factorSource.factorSourceID)
+					}
+				}
```
GhenadieVP, line 73: Could be:
```swift
state.$shieldBuilder.withLock { builder in
     switch change {
     case let .remove(_, factorSource, _):
         builder = builder.removeFactorFromPrimary(factorSourceId: factorSource.factorSourceID)
     case let .insert(_, factorSource, _):
	 builder = builder.addFactorSourceToPrimaryThreshold(factorSourceId: factorSource.factorSourceID)
      }
}
```

RadixWallet/Features/ShieldSetup/SelectFactorSources/Children/SelectFactorSources+Reducer.swift
```diff
@@ -0,0 +1,113 @@
...
+				builder = builder.setThreshold(threshold: UInt8(builder.primaryRoleThresholdFactors.count))
+			}
+
+			if !state.didInteractWithSelection, factorSources != nil {
+				state.didInteractWithSelection = true
+			}
+
+			return .none
+
+		case .buildButtonTapped:
```
GhenadieVP, line 86: Build is not the most correct name, as the shield will not be builded, I guess it can be `continueButtonTapped`

RadixWallet/Features/ShieldSetup/SelectFactorSources/Children/SelectFactorSources+View.swift
```diff
@@ -0,0 +1,141 @@
...
+									Text("Cannot use this factor by itself")
+										.textStyle(.body2Regular)
+										.foregroundStyle(.app.alert)
+										.padding(.top, .small3)
+								}
+							}
+							.foregroundStyle(.app.gray2)
+							.padding(.top, .medium3)
+							.flushedLeft
+						}
```
GhenadieVP, line 87: So this is basically the section header, would be nice to have it in the Selection itself, not sure how easy it is todo.
For now one question - does it mean that the header is also tappable? meaning that the first factor source will be selected when user taps on the section header?

RadixWallet/Features/ShieldSetup/SelectFactorSources/SelectFactorSourcesCoordinator.swift
```diff
@@ -0,0 +1,42 @@
+import ComposableArchitecture
+
+@Reducer
+struct SelectFactorSourcesCoordinator: Sendable, FeatureReducer {
```
GhenadieVP, line 4: Is this coordinator needed? Can't ShieldSetupCoordinator use the SelectFactorSources directly? Not opposed to it, but want to understand if this used for technical reasons or rather it is used for general consistency.

RadixWallet/Features/ShieldSetup/SelectFactorSources/Children/SelectFactorSources+Reducer.swift
danvleju-rdx: `Selection` requires `selection: Binding<[Value]?>`

RadixWallet/Features/ShieldSetup/SelectFactorSources/Children/SelectFactorSources+Reducer.swift
danvleju-rdx: Something like this:
```Swift
extension Shared where Value == SecurityShieldBuilder {
    func initialize() {
        withLock { sharedValue in
            sharedValue = SecurityShieldBuilder()
        }
    }
}
```

RadixWallet/Features/ShieldSetup/SelectFactorSources/Children/SelectFactorSources+View.swift
danvleju-rdx: >does it mean that the header is also tappable? meaning that the first factor source will be selected when user taps on the section header?

No, only `FactorSourceCard` is tappable

RadixWallet/Features/ShieldSetup/SelectFactorSources/SelectFactorSourcesCoordinator.swift
danvleju-rdx: Not for technical reasons. Discussed about this [here](https://rdxworks.slack.com/archives/C03QFAWBRNX/p1733492750555539)

RadixWallet/Features/ShieldSetup/SelectFactorSources/Children/SelectFactorSources+Reducer.swift
GhenadieVP: aah right, thought it is not a bidirectional binding, since this is a computed property.

GhenadieVP (approved):

RadixWallet/Features/ShieldSetup/SelectFactorSources/Children/SelectFactorSources+Reducer.swift
```diff
@@ -0,0 +1,110 @@
...
+	func reduce(into state: inout State, viewAction: ViewAction) -> Effect<Action> {
+		switch viewAction {
+		case .task:
+			return .run { [shieldBuilder = state.shieldBuilder] send in
+				let result = await TaskResult {
+					let factorSources = try await factorSourcesClient.getFactorSources().elements
+					return shieldBuilder.sortedFactorSourcesForPrimaryThresholdSelection(factorSources: factorSources)
+				}
+				await send(.internal(.factorSourcesResult(result)))
+			}
```
matiasbzurovski, line 56: is there any benefit on doing this rather than catching the error inside the `run`? Because the latter is how we mostly do it in other reducers, and seems cleaner to me:
```Swift
return .run { [shieldBuilder = state.shieldBuilder] send in
	let factorSources = try await factorSourcesClient.getFactorSources().elements
	let result =  shieldBuilder.sortedFactorSourcesForPrimaryThresholdSelection(factorSources: factorSources)
	await send(.internal(.setFactorSources(result)))
} catch: { error, send in
	errorQueue.schedule(error)
}
```

Then you would also update the `InternalAction`
```diff
- case factorSourcesResult(TaskResult<[FactorSource]>)
+ case setFactorSources([FactorSource])
```

matiasbzurovski (approved):

RadixWallet/Features/ShieldSetup/SelectFactorSources/Children/SelectFactorSources+View.swift
```diff
@@ -0,0 +1,141 @@
...
+			nil
+		}
+	}
+
+	var isValidSelection: Bool {
+		selectedFactorSourcesStatus == .optimal || selectedFactorSourcesStatus == .suboptimal
+	}
+
+	var shouldShowPasswordMessage: Bool {
+		(selectedFactorSources ?? []).contains(where: { $0.factorSourceKind == .password }) &&
```
Sajjon, line 27: Logic! Should go into Sargon imo! 

Add a TODO?

danvleju-rdx: Sergiu will add this to Sargon :) 
I’ll add a TODO comment.

PR #1415, opened 2024-12-11
Wait until first account creation to create Profile

matiasbzurovski:
Fixes bug where users wouldn't be able to create a Profile if they cancelled the creation flow in the middle.

[Reference](https://rdxworks.slack.com/archives/C031A0V1A1W/p1733932355719239)


CyonAlexRDX:
I'm glad you solved this bug! But I'm not happy that we have this kind of logic in hosts. But bigger fish to fry... so I will approve this and then later, much later, we will rewrite all of this, to use Sargon only

CyonAlexRDX (approved):

PR #1419, opened 2024-12-13
Fix profile claim after import

GhenadieVP:


RadixWallet/Clients/TransportProfileClient/TransportProfileClient+Live.swift
```diff
@@ -12,10 +12,11 @@ extension TransportProfileClient: DependencyKey {
 		return Self(
 			importProfile: { profile, factorSourceIDs, skippedMainBdfs, containsP2PLinks in
 				do {
+					try await profileStore.importProfile(profile, skippedMainBdfs: skippedMainBdfs)
```
GhenadieVP, line 15: Importing the profile will also claim it, so claim it first before importing.

matiasbzurovski (approved):

PR #1421, opened 2024-12-13
[ABW-3921] Factor sources list

matiasbzurovski:
Jira ticket: [ABW-3921](https://radixdlt.atlassian.net/browse/ABW-3921)
 
## Description
This PR adds a generic `FactorSourcesList` view that will populate every factor source of a given `kind`.

## Notes
Added support for adding `DeviceFactorSource`, but according to latest design session this will change into a generic flow (from [this view](https://zpl.io/wyqB6Bd)) regardless of the `FactorSourceKind` from which we are coming from. Will implement once it is defined.

## Video
[Link](https://drive.google.com/file/d/1mgFY4YweIhkDMkKfmpb_Ie5fA2si9ljB/view?usp=sharing)

[ABW-3921]: https://radixdlt.atlassian.net/browse/ABW-3921?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


RadixWallet/Core/FeaturePrelude/FactorSourceCard/FactorSourceCard+DataSource.swift
```diff
@@ -87,4 +92,21 @@ extension FactorSourceKind {
...
+		case .ledgerHqHardwareWallet:
+			(.ledger, "Learn about Ledger Nanos")
+		case .offDeviceMnemonic:
+			(.offdevicemnemonic, "Learn about passphrases")
+		case .arculusCard:
+			(.arculuscard, "Learn about Arculus Cards")
+		case .password:
+			(.password, "Learn about passwords")
+		default:
+			fatalError("Not supported yet")
```
matiasbzurovski, line 109: didn't add the localization for these because we want to use the same key than the glossary item, and I am not sure if they are gonna be named as I did.

templates/Xcode/TCA Feature.xctemplate/___FILEBASENAME___+View.swift
```diff
@@ -1,53 +1,18 @@
-extension ___VARIABLE_featureName___.State {
```
matiasbzurovski, line 1: Update template to remove `public`

RadixWallet/Clients/DeviceFactorSourceClient/DeviceFactorSourceClient+Live.swift
```diff
@@ -15,61 +15,22 @@ extension DeviceFactorSourceClient: DependencyKey {
...
-					personasHidden = try await personasClient.getHiddenPersonasOnCurrentNetwork().elements
-				}
-
-				var allNonHiddenEntities = accountNonHidden.map(AccountOrPersona.account)
-				allNonHiddenEntities.append(contentsOf: personasNonHidden.map(AccountOrPersona.persona))
-
-				var allHidden = accountHidden.map(AccountOrPersona.account)
-				allHidden.append(contentsOf: personasHidden.map(AccountOrPersona.persona))
-
-				return (allNonHiddenEntities, allHidden)
```
Sajjon, line 48: nice to rid of this (moved to Sargon), good job!

RadixWallet/Clients/DeviceFactorSourceClient/DeviceFactorSourceClient+Live.swift
```diff
@@ -15,61 +15,22 @@ extension DeviceFactorSourceClient: DependencyKey {
...
+				throw UnexpectedFactorSource()
 			}
-
-			let isMnemonicMarkedAsBackedUp = userDefaults.getFactorSourceIDOfBackedUpMnemonics().contains(factorSource.id)
-
 			return EntitiesControlledByFactorSource(
-				entities: nonHiddenEntitiesForSource,
-				hiddenEntities: hiddenEntitiesForSource,
+				entities: result.accounts.map(AccountOrPersona.account) + result.personas.map(AccountOrPersona.persona),
+				hiddenEntities: result.hiddenAccounts.map(AccountOrPersona.account) + result.hiddenPersonas.map(AccountOrPersona.persona),
```
Sajjon, line 30: feels like we want `result.hiddenAccounts.map(AccountOrPersona.account) + result.hiddenPersonas.map(AccountOrPersona.persona),` in Sargon...?

Sajjon: and `result.accounts.map(AccountOrPersona.account) + result.personas.map(AccountOrPersona.persona),` too

RadixWallet/Features/AppFeature/Overlay/InfoLinkSheet+Reducer.swift
```diff
@@ -63,6 +63,11 @@ extension InfoLinkSheet {
 		case securityshield
 		case buildsecurityshields
 		case nohardwaredevice
+		case device
```
Sajjon, line 66: I would prefer nesting here instead of flat, e.g. `case infoAboutFactors(FactorsInfo)`.

RadixWallet/Features/SecurityCenterFeature/SecurityFactors/FactorSourcesList+View.swift
```diff
@@ -0,0 +1,206 @@
...
+		case .trustedContact, .securityQuestions:
+			fatalError("Not implemented")
+		}
+	}
+
+	var main: Row? {
+		rows.first(where: \.integrity.isExplicitMain)
+	}
+
+	var others: [Row] {
```
Sajjon, line 103: `	var others: [Row] {` should probably be in Sargon?

RadixWallet/Features/SecurityCenterFeature/SecurityFactors/FactorSourcesList.swift
```diff
@@ -0,0 +1,269 @@
...
+
+// MARK: - DeviceFactorSourcesList.State.Row
+extension FactorSourcesList.State {
+	struct Row: Sendable, Hashable {
+		let integrity: FactorSourceIntegrity
+		let linkedEntities: FactorSourceCardDataSource.LinkedEntities
+		let status: Status
+	}
+
+	enum Status: Sendable, Hashable {
```
Sajjon, line 226: This `Status` enum ought to be in Sargon?

templates/Xcode/TCA Feature.xctemplate/___FILEBASENAME___.swift
```diff
@@ -1,17 +1,13 @@
...
 
-public struct ___VARIABLE_featureName___: Sendable, FeatureReducer {
-	public struct State: Sendable, Hashable {
-		public init() {}
-	}
+struct ___VARIABLE_featureName___: Sendable, FeatureReducer {
+	struct State: Sendable, Hashable {}
 
-	public enum ViewAction: Sendable, Equatable {
+	enum ViewAction: Sendable, Equatable {
```
Sajjon, line 6: good

RadixWallet/Clients/DeviceFactorSourceClient/DeviceFactorSourceClient+Live.swift
matiasbzurovski: not really, since it is only for this particular case of entities linked to `DeviceFactorSource` that we want to treat it as `AccountOrPersona` lists. In fact, we may even end up updating the usages of this method (could be using the new UI for other flows) and end removing this implementation as well (and always use `SargonOS` output, as we do for new `FactorSourcesList`).

RadixWallet/Features/AppFeature/Overlay/InfoLinkSheet+Reducer.swift
matiasbzurovski: on a regular enum I would've gone with that categorization for sure. However, we are using the enum cases `rawValue` (which we wouldn't be able to have with nested cases) to support navigation from one glossary item to another. This is, if you are looking at a text that is provided like this:
```
is that users can “stake” [XRD tokens](?glossaryAnchor=xrd) to increase the security of the network
```

then we can link to the user to `InfoLinkSheet.GlossaryItem.xrd` when they tap on `XRD tokens`. It is definitely easier to maintain if we have it this way (`?glossaryAnchor=device`), than if we need to support something more complicated such as `?glossaryAnchor=infoaboutfactors.device`

RadixWallet/Features/DisplayEntitiesControlledByMnemonic/DisplayEntitiesControlledByMnemonic+Reducer.swift
```diff
@@ -108,27 +108,29 @@ struct DisplayEntitiesControlledByMnemonic: Sendable, FeatureReducer {
 	}
 }
 
-private extension [SecurityProblem] {
-	func isMnemonicMarkedAsBackedUp(accounts: [Account], personas: [Persona]) -> Bool {
-		allSatisfy { problem in
+extension [SecurityProblem] {
+	/// Returns if the given collection of `SecurityProblem` has any `.problem3` for at least one of the entities provided
+	func hasProblem3(accounts: [Account], personas: [Persona]) -> Bool {
```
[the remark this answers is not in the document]
matiasbzurovski: Probably yes, but problem 3 and 9 are going away after latest updates on [Confluence](https://radixdlt.atlassian.net/wiki/spaces/AT/pages/3392569357/Security-related+Problem+States+in+the+Wallet), so not gonna spend more time working on logic that will be updated.

RadixWallet/Features/SecurityCenterFeature/SecurityFactors/FactorSourcesList+View.swift
matiasbzurovski: this is just a differentiation for whether we should show a header and main device, or if they get all mixed up together. Being plainly UI, I don't think it is worth moving to Sargon.

RadixWallet/Features/SecurityCenterFeature/SecurityFactors/FactorSourcesList.swift
matiasbzurovski: mm this one is debatable, could be worth moving to Sargon when working on Security Problems updates mentioned on previous comment.

RadixWallet/Features/SecurityCenterFeature/SecurityFactors/FactorSourcesList.swift
```diff
@@ -0,0 +1,269 @@
...
+							location: .intoKeychainAndProfile,
+							onMnemonicExistsStrategy: .appendWithCryptoParamaters
+						),
+						wordCount: .twentyFour
+					)
+				)
+			case .ledgerHqHardwareWallet, .offDeviceMnemonic, .arculusCard, .password:
+				// NOTE: Added `.device` support as placeholder, but not adding the logic for ledger (which we already support)
+				// since Matt mentioned we will probably always present this screen: https://zpl.io/wyqB6Bd
+				// and I don't want to add all the logic for checking if there is a CE or not just to migrate it later.
```
GhenadieVP, line 124: Not clear why you cannot add the current .ledgerHqHardwareWallet flow

RadixWallet/Features/SecurityCenterFeature/SecurityFactors/FactorSourcesList.swift
```diff
@@ -0,0 +1,269 @@
...
+					.notBackedUp
+				}
+				return State.Row(
+					integrity: entity.integrity,
+					linkedEntities: entity.linkedEntities,
+					status: status
+				)
+			}
+			await send(.internal(.setRows(rows)))
+		}
```
GhenadieVP, line 214: Does not seem to need to be an Effect per say.

RadixWallet/Features/SecurityCenterFeature/SecurityFactors/FactorSourcesList.swift
matiasbzurovski: Matt mentioned on latest design session that adding the `Enter Seed Phrase` as a placeholder was fine for devices, but that the Add Factor Source flow would be defined into something generic (as presenting the view I linked regardless of the factor source list you are viewing).

The Add Ledger Flow consist of 3 views: Add Ledger view, No P2P Alert & Add P2P view.

I didn't want to add all of these to this view if they are going to be removed later.

GhenadieVP (approved):

PR #1422, opened 2024-12-16
GeneralSubintent manifest class

GhenadieVP:
Display raw manifest if the Pre-Auth manifest is not classified as GeneralSubintent.

# Demo

https://drive.google.com/file/d/1bywGYgk8GTV6YxxEknSk2OyizHW8N0mu/view

### DIFF
--- a/RadixWallet/SargonExtensions/SargonSecureStorageDriver.swift
+++ b/RadixWallet/SargonExtensions/SargonSecureStorageDriver.swift
@@ -61,6 +61,10 @@ final class SargonSecureStorage: SecureStorageDriver {
 		}
 	}
 
+	func containsDataForKey(key: SecureStorageKey) async throws -> Bool {
+		false
+	}
+
 	private func loadHostId() -> BagOfBytes? {
 		let deviceInfo: DeviceInfo? = {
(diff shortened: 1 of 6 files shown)


matiasbzurovski (approved):

RadixWallet/SargonExtensions/SargonSecureStorageDriver.swift
```diff
@@ -61,6 +61,10 @@ final class SargonSecureStorage: SecureStorageDriver {
 		}
 	}
 
+	func containsDataForKey(key: SecureStorageKey) async throws -> Bool {
+		false
```
danvleju-rdx, line 65: missed implementation?

CyonAlexRDX: not used by host yet I think. Sargon contains newer APIs than hosts needs. Mathias is working on this part of FactorSource work. So should be safe to return `false` here since it is in fact never called from Sargon.

CyonAlexRDX (approved):
LGTM

danvleju-rdx (approved):

PR #1427, opened 2024-12-18
[ABW-3932] Factor Source detail view

matiasbzurovski:
Jira ticket: [ABW-3932](https://radixdlt.atlassian.net/browse/ABW-3932)

## Description
This PR shows the detail view for every supported Factor Source kind. 
Also, merges `UpdateAccountLabel` & `UpdateP2PLinkName` into one single `RenameLabel` feature that also supports renaming a `FactorSource`.

### Notes
For now, it will only allow to rename the Factor Source, as well as view its seed phrase for the `device` case. The spot check will be implemented later (after integrating `KeysCollector`).

Also, for `device` details, if you try to view the seed phrase when it hasn't been entered, an error alert is shown (check video for reference). We may wanna update the flow to allow user to enter the seed phrase directly when reaching this state.

## Video
[Link](https://drive.google.com/file/d/1073nkvIOiwotL_goL4lDTHcTZp25GjiI/view?usp=sharing)

[ABW-3932]: https://radixdlt.atlassian.net/browse/ABW-3932?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


RadixWallet/Core/FeaturePrelude/RenameLabel/RenameLabel.swift
```diff
@@ -0,0 +1,149 @@
...
+		}
+	}
+
+	func reduce(into state: inout State, internalAction: InternalAction) -> Effect<Action> {
+		switch internalAction {
+		case .handleSuccess:
+			overlayWindowClient.scheduleHUD(.renamedLabel)
+			return .send(.delegate(.labelUpdated(state.kind)))
+		case let .handleFactorSourceUpdate(factorSource):
+			state.kind = .factorSource(factorSource, name: factorSource.name)
```
GhenadieVP, line 102: Why this is different from account an connector update? Where the state.kind is set before the update.

matiasbzurovski: Because for the mentioned two cases we can just do:
```Swift
account.displayName = DisplayName(nonEmpty: nonEmpty)
connector.displayName = nonEmpty.rawValue
```

However, when updating a `FactoSource`, we need to call the following async method on SargonOS:
```Swift
func updateFactorSourceName(factorSource: FactorSource, name: String)async throws  -> FactorSource
```

GhenadieVP (approved):

RadixWallet/Features/SecurityCenterFeature/SecurityFactors/FactorSourceDetail+View.swift
```diff
@@ -1,16 +1,127 @@
...
+		private func rows(_ viewStore: ViewStore<FactorSourceDetail.State, FactorSourceDetail.ViewAction>) -> [SettingsRow<FactorSourceDetail>.Kind] {
+			switch viewStore.integrity {
+			case .device:
+				[
+					.header(L10n.FactorSources.Detail.settings),
+					renameRow(viewStore),
+					.header(L10n.FactorSources.Detail.test),
+					spotCheckRow(viewStore),
+					.header(L10n.FactorSources.Detail.advanced),
+					// TODO: Define how to handle case where seed phrase was never entered
```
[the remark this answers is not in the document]
matiasbzurovski: add support for this

giannis-rdx:
@matiasbzurovski I see that details view does not match the zeplin screens. 
Am I missing something?

matiasbzurovski:
@giannis-rdx video is outdated from when I created PR (which had older designs), will update

giannis-rdx:
> @giannis-rdx video is outdated from when I created PR (which had older designs), will update

no need then! I just thought we had new screens that i was not aware of

PR #1433, opened 2025-01-08
1.11.4 hotfixes

GhenadieVP:
- https://radixdlt.atlassian.net/browse/ABW-4037
- https://radixdlt.atlassian.net/browse/ABW-4040
- https://radixdlt.atlassian.net/browse/ABW-4126
- https://radixdlt.atlassian.net/browse/ABW-4127


RadixWallet/Features/CreateAccount/Coordinator/CreateAccountCoordinator+Models.swift
```diff
@@ -5,18 +5,15 @@ import SwiftUI
 struct CreateAccountConfig: Sendable, Hashable {
 	let specificNetworkID: NetworkID?
 	let isFirstAccount: Bool
-	let isNewProfile: Bool
```
GhenadieVP, line 8: In the following files this [change](https://github.com/radixdlt/babylon-wallet-ios/pull/1414/commits/be51e0f0c7be3473436e090a32583e34e85c7680) is reverted, in favour of clearing out the ephemeral profile when user goes again in the create wallet flow.

RadixWallet/Clients/ProfileStore/ProfileStore.swift
```diff
@@ -37,6 +37,7 @@ extension ProfileStore {
 
 extension ProfileStore {
 	func createNewProfile() async throws {
+		try? await SargonOS.shared.deleteWallet()
```
GhenadieVP, line 40: Sargon would throw the missing profile if there is none when deleting the Wallet, which we can safely ignore.

danvleju-rdx (approved):

PR #1436, opened 2025-01-09
[ABW-3973] Shield Setup - Regular Access

danvleju-rdx:
Jira ticket: [ABW-3973](https://radixdlt.atlassian.net/browse/ABW-3973)

## Description
Adds the [Setup Regular Access](https://zpl.io/EdmE5ng) flow.

## How to test
You can use the profile provided [here](https://github.com/radixdlt/sargon/pull/311/files) to test different combinations of factor sources.

## Video
[video](https://rdxworks.slack.com/archives/C03QFAWBRNX/p1736427586111459?thread_ts=1736427550.610909&cid=C03QFAWBRNX)

[ABW-3973]: https://radixdlt.atlassian.net/browse/ABW-3973?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


RadixWallet/Features/ShieldSetup/RolesSetup/Children/RegularAccessSetup/RegularAccessSetup+Reducer.swift
```diff
@@ -0,0 +1,159 @@
+import ComposableArchitecture
+
+// MARK: - SelectFactorSources
+@Reducer
+struct RegularAccessSetup: FeatureReducer, Sendable {
```
CyonAlexRDX, line 5: We have earlier agreed to not call anything `Access` in code :) 

It is `Primary`, `Recovery` and `Confirmation` **everywhere** in code.

CyonAlexRDX: Because that is what the roles are called in Sargon, In RET, in Scrypto, in most old documents and on Slack. Just that Matt chose to change the user facing terms recently.

danvleju-rdx: Right, will update

CyonAlexRDX (approved):
LGTM, goob job!

RadixWallet/Core/DesignSystem/Styles/SecondaryRectangularButtonStyle.swift
```diff
@@ -1,6 +1,7 @@
 // MARK: - SecondaryRectangularButtonStyle
 struct SecondaryRectangularButtonStyle: ButtonStyle {
 	@Environment(\.controlState) var controlState
+	var font: SwiftUI.Font = .app.body1Header
```
matiasbzurovski, line 4:
```suggestion
	let font: SwiftUI.Font
```

since we already provide default on `init`

RadixWallet/Features/ShieldSetup/RolesSetup/Children/PrimaryRoleSetup/SelectNumberOfFactorsView.swift
```diff
@@ -0,0 +1,81 @@
...
+
+			Picker("", selection: $selectedNumberOfFactors) {
+				ForEach(pickerItems, id: \.self) { threshold in
+					Text(threshold.title)
+						.tag(threshold)
+				}
+			}
+			.pickerStyle(.wheel)
+
+			Spacer()
```
matiasbzurovski, line 39: [designs](https://zpl.io/dRxDXDE) don't apply 100% for us given we cannot add the dynamic height, but maybe the `Spacer` should be between `Text` & `Picker` instead?

RadixWallet/Features/ShieldSetup/RolesSetup/Children/PrimaryRoleSetup/PrimaryRoleSetup+View.swift
```diff
@@ -0,0 +1,365 @@
...
+			}
+			.frame(maxWidth: .infinity)
+			.embedInContainer
+		}
+
+		private var thresholdSelectorView: some SwiftUI.View {
+			HStack(spacing: .zero) {
+				let thresholdTitle = store.threshold.titleShort
+				let title = L10n.ShieldWizardRegularAccess.ThresholdDescription.title(thresholdTitle)
+				let parts = title.components(separatedBy: "**\(thresholdTitle)**")
```
matiasbzurovski, line 185: mmm this makes the logic dependent on the localisation: if we ever remove the bold side, this won't work anymore.

RadixWallet/Features/ShieldSetup/ShieldSetupCoordinator.swift
```diff
@@ -55,7 +60,7 @@ struct ShieldSetupCoordinator: Sendable, FeatureReducer {
 		case .path(.element(id: _, action: .prepareFactors(.delegate(.finished)))):
 			return .send(.internal(.selectFactors))
 		case .path(.element(id: _, action: .selectFactors(.delegate(.finished)))):
-			return .none
+			return .send(.internal(.rolesSetup))
```
matiasbzurovski, line 63: we can just do `state.path.append(.rolesSetup(.init()))` and skip the added `InternalAction`. 

This was necessary for `.prepareFactors` & `.selectFactors` given we need to resolve which to show on a throwing function, but if it is straightforward we might skip the boilerplate.

matiasbzurovski (commented):
Looking good!

RadixWallet/Features/ShieldSetup/RolesSetup/Children/PrimaryRoleSetup/PrimaryRoleSetup+View.swift
danvleju-rdx: Well, yes, but I see it as being strongly dependent on the bold side. Sergiu asked about this [here](https://rdxworks.slack.com/archives/C031A0V1A1W/p1736166125055479).

matiasbzurovski: I would recommend at least saving the bold part in a new key, that you send to the `title` one:
```
"shieldWizardRegularAccess_thresholdDescription_title" = "You\'ll need to use %@ of the following";
"shieldWizardRegularAccess_thresholdDescription_selection" = "**%@**";
```
This way we don't assume on the logic that the localisation is something that may not be (or that Crowdin ends up screwing some way)

danvleju-rdx: Sure, will do

GhenadieVP (approved):

RadixWallet/Features/ShieldSetup/RolesSetup/Children/PrimaryRoleSetup/PrimaryRoleSetup+View.swift
```diff
@@ -0,0 +1,365 @@
...
+			nil
+		case .PrimaryRoleMustHaveAtLeastOneFactor:
+			.init(type: .error, text: L10n.ShieldSetupStatus.Transactions.atLeastOneFactor)
+		default:
+			.init(type: .warning, text: L10n.ShieldSetupStatus.invalidCombination)
+		}
+	}
+
+	var canContinue: Bool {
+		validatedRoleStatus != .PrimaryRoleMustHaveAtLeastOneFactor && validatedRoleStatus != .MissingAuthSigningFactor
```
GhenadieVP, line 18: As discussed in demo this should be an OR, not AND. Additionally this seems like a candidate to go in Sargon, but with another PR

danvleju-rdx: Actually, it should be AND. The problem is on the Sargon side, specifically with the `validate()` function, which doesn’t assign the correct priorities to errors. In the demo case I encountered, instead of returning `MissingAuthSigningFactor`, `validate()` returns `SingleFactorUsedInPrimaryMustNotBeUsedInAnyOtherRole` (which doesn’t prevent the user from proceeding to the next step).

I’ll add a TODO comment to fix this and move the logic to Sargon (either Sergiu or I will handle it).

PR #1460, opened 2025-02-11
[ABW-4161] Factor Source Spot Check 

matiasbzurovski:
JIRA ticket: [ABW-4161](https://radixdlt.atlassian.net/browse/ABW-4161)

## Description
Adds support for the new method required by `HostInteractor` to allow performing Spot Check from Factor Source Details or whenever SargonOS triggers it (e.g. when creating a new entity once [ABW-4158](https://radixdlt.atlassian.net/browse/ABW-4158) is implemented).

[ABW-4158]: https://radixdlt.atlassian.net/browse/ABW-4158?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[ABW-4161]: https://radixdlt.atlassian.net/browse/ABW-4161?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


RadixWallet/Clients/OverlayWindowClient/OverlayWindowClient+Interface.swift
```diff
@@ -154,6 +164,14 @@ extension OverlayWindowClient {
 	}
 
 	func authorize(purpose: AuthorizationPurpose) async -> SheetAction {
-		await scheduleSheet(.init(root: .authorization(.init(purpose: purpose))))
+		guard let state = Authorization.State(purpose: purpose) else {
+			assertionFailure("Unable to present Authorization")
+			return .authorization(.cancelled)
+		}
```
matiasbzurovski, line 170: This is required because `FactorSourceAccess` was [updated](https://github.com/radixdlt/babylon-wallet-ios/pull/1460/files#diff-0fa943a9bf085a5875e3474258151ad3e46affed2c69d7cb78d34e9e062bd87eR6) to always be initialized either with a `FactorSource` or a `FactorSourceId`. This makes the code inside such component much cleaner and robust.

However, the downside is that the `Authorization`, which always happens with the main BDFS, needed to be [updated](https://github.com/radixdlt/babylon-wallet-ios/pull/1460/files#diff-b183fcb83ed5779b671ebe64326db6f4ecf8cc1fc38114dcdacd3945a868c375R9-R13) to have a conditional `init`, which leads down to this error handling. 

Another option would have been to force try the access to the main BDFS (`try! SargonOS.shared.mainBdfs()`), but I preferred adding this handling.

RadixWallet/SargonExtensions/SargonHostInteractor.swift
```diff
@@ -114,4 +114,22 @@ final class SargonHostInteractor: HostInteractor {
 			fatalError("Unexpected action")
 		}
 	}
+
+	func spotCheck(factorSource: FactorSource, allowSkip: Bool) async throws -> SpotCheckResponse {
+		let action = await overlayWindowClient.spotCheck(factorSource: factorSource, allowSkip: allowSkip)
```
Sajjon, line 119: nit: I think `performedAction`, `executed` or `outcome` is slightly better, to make it clear that the action **has happened**. (which is clear from `.cancelled`).

RadixWallet/Prelude/ErrorQueue/ErrorQueue+Live.swift
```diff
@@ -13,8 +13,8 @@ extension ErrorQueue: DependencyKey {
 					return
 				}
 
-				if let commonError = error as? CommonError, commonError == .HostInteractionAborted {
-					loggerGlobal.info("Ignroing HostInteractionAborted error ")
+				if error.isHostInteractionAborted {
+					loggerGlobal.info("Ignoring HostInteractionAborted error ")
```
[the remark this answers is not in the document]
matiasbzurovski: no, it is totally expected and fine for users to tap on `X` button to abort an interaction

RadixWallet/Features/SpotCheck/SpotCheck.swift
```diff
@@ -0,0 +1,97 @@
...
+
+		default:
+			errorQueue.schedule(error)
+		}
+	}
+
+	private func loadMnemonic(id: FactorSourceIdFromHash) throws -> MnemonicWithPassphrase {
+		guard let result = try secureStorageClient.loadMnemonic(factorSourceID: id, notifyIfMissing: false) else {
+			struct MissingMnmemonicError: Error {}
+			throw MissingMnmemonicError()
```
Sajjon, line 93: prefer using `(Sargon.)CommonError.UnableToLoadMnemonicFromSecureStorage(badValue: "id: \(id)")` so we get the correct user facing Sargon error code.

RadixWallet/Features/SpotCheck/SpotCheck.swift
```diff
@@ -0,0 +1,97 @@
...
+		case .factorSourceAccess(.delegate(.cancel)):
+			.send(.delegate(.cancelled))
+		default:
+			.none
+		}
+	}
+}
+
+private extension SpotCheck {
+	func perform(factorSource: FactorSourcePerformer) -> Effect<Action> {
```
Sajjon, line 50: nit: We have been a bit lazy in iOS land as of late - and probably should continue to be so to iterate fast. But looks like this logic ought to be a dependency, so that the feature is testable (which is a strength of TCA we are not using ATM). 

So maybe put a comment? 

(A client which Live version is using `ledgerHardwareWalletClient` and `secureStorageClient` etc..)

RadixWallet/Features/Signing/Signing.swift
```diff
@@ -131,16 +131,16 @@ private extension Signing {
 		}
 	}
 
-	private func handleError(factorSource: FactorSource, error: Error, send: Send<Signing.Action>) async {
-		switch factorSource.kind {
+	private func handleError(factorSource: FactorSourcePerformer, error: Error, send: Send<Signing.Action>) async {
```
Sajjon, line 134: CTRL M please

RadixWallet/Features/FactorSourceAccess/FactorSourcePerformer.swift
```diff
@@ -0,0 +1,31 @@
+import Foundation
+
+// MARK: - FactorSourcePerformer
+/// An enum containing the necessary information to use a given `FactorSource`.
+///
+/// This becomes handy to unify factor sources that can be used directly (`device`, `ledger` & `arculusCard`) and those
+/// that require user input to retrieve the associated `MnemonicWithPassphrase` (`offDeviceMnemonic` & `password`).
+enum FactorSourcePerformer: Sendable, Hashable {
```
Sajjon, line 8: Hmm I dont like this name. Can we come up with something more clear. And it is actually somewhat related to `PrivateHierarchicalDeterministicFactorSource` in Sargon. 

`Private` means it holds secrets. Which this does.  `PrivateFactorSource` seems like a better name.

CyonAlexRDX (approved):
LGTM, good job, left some minor remarks

RadixWallet/Features/SpotCheck/SpotCheck.swift
matiasbzurovski: mmm probably not. We have the clients (`deviceFactorSourceClient`, `ledgerHardwareWalletClient`, etc) to access the factor source. But the actual logic to perform the spot check makes sense to be in the `SpotCheck` reducer, the same way the logic to perform signing is inside `Signing` reducer and the logic to derive public keys is inside `DerivePublicKeys` reducer

matiasbzurovski: if we wanted to test this, it would be possible on current state by injecting mock factor source clients

RadixWallet/Features/Signing/Signing.swift
matiasbzurovski: added, but this should be done by Swiftformat (otherwise it is just arbitrary)

danvleju-rdx (approved):

RadixWallet/Features/SpotCheck/SpotCheck.swift
CyonAlexRDX: sure, a matter of test I guess. I like "as-stupid-reduceers-as-possible". And I like putting as much logic in "clients" as possible. Reducers are then mostly "glue" between View/State and Clients. The advantage of this is that when I did write unit tests of the reducer I could often override much fewer methods on clients. e.g. I could have a `BreadClient` and just override `bakeBread` instead of - as you suggest/prefer - have `flourClient`, `ovenClient` and `waterClient` and in unit test have to override `provideFlour`, `heatOven` and `tapWater` which is 3 methods instead of one.

If that makes sense? :)

RadixWallet/Features/Signing/Signing.swift
CyonAlexRDX: true. I think a good threshold is: `if more than two args, CTRL M` but probably should be character based, i.e. if two args with long labels => CTRL M.

RadixWallet/Features/SpotCheck/SpotCheck.swift
matiasbzurovski: good example :)

But in that case, where you need to do 3 steps to perform the example task, it may justify having the mentioned client to decouple responsibilities a bit. In this case the responsibility is very simple, it just needs to perform a task which is get the correct input to call Sargon. I think adding clients adds boilerplate and if testability is the same, better to do it straightforward on the reducer. 
Otherwise you also end up with lots of clients that have similar but different responsibilities and then it is hard to follow the reasoning behind each of them.

CyonAlexRDX: yeah a fine balance for sure!

CyonAlexRDX: I think the cost of writing reducers are quite low thanks to macros in the past year.

PR #1473, opened 2025-04-30
Crashlytics

GhenadieVP:


RadixWallet/Clients/KeychainClient/KeychainClient+Live.swift
```diff
@@ -15,18 +16,28 @@ extension KeychainClient: DependencyKey {
...
-					attributes: attributes
-				)
+				do {
+					try keychainHolder.setDataWithAuth(
+						data,
+						forKey: key,
+						attributes: attributes
+					)
+				} catch {
+					Crashlytics.crashlytics().record(error: error)
```
Sajjon, line 38: Are we 100% sure that `error` does not and never can contain secrets like Mnemonic or Entropy of Mnemonic? 

Might be able to without too much hassle process it before sending it to Crashlytics, eg scan string for 32 bytes hex strings and or BIP32 words, etc

